PR #83, opened 2022-06-29
Drop `await_future` utility function

cutwater:
The `loop.run_in_executor()` method returns a Future.
There is no need to additionally create a wrapper task
that only waits for it.


benthomasson (approved):
This is fine since Futures and Tasks have the same interface.   Creating a task from a future wrapped in a coroutine was unnecessary.

PR #105, opened 2022-07-13
Shutdown the engine after all sources are done

bzwei:
When multiples sources are included in a ruleset, if one source finishes it forces the main engine to shutdown while other sources are still producing events.

This work adds a counter for finished sources and shuts down the engine after all sources have completed.


benthomasson:
> When multiples sources are included in a ruleset, if one source finishes it forces the main engine to shutdown while other sources are still producing events.

This is the desired functionality.

PR #111, opened 2022-07-14
Install flake8-bugbear plugin

cutwater:
Fix warnings introduced by installing new flake8 plugin.

Warnings:

```
./tests/sources/log_scraper.py:15:19: B007 Loop control variable 'loc' not used within the loop body.
    If this is intended, start the name with an underscore.
./tests/event_filters/json_filter.py:34:30: B006 Do not use mutable data structures for argument defaults.
    They are created during function definition time.
    All calls to the function reuse this one instance of that data structure, persisting changes between them.
./tests/event_filters/json_filter.py:34:47: B006 Do not use mutable data structures for argument defaults.
    They are created during function definition time.
    All calls to the function reuse this one instance of that data structure, persisting changes between them.
./ansible_events/inventory.py:18:9: B007 Loop control variable 'groupname' not used within the loop body.
    If this is intended, start the name with an underscore.
./ansible_events/event_filters/json_filter.py:34:30: B006 Do not use mutable data structures for argument defaults.
./ansible_events/event_filters/json_filter.py:34:47: B006 Do not use mutable data structures for argument defaults.
```

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           python-version: '3.10'
       - name: Install dependencies
-        run: python -m pip install flake8
+        run: python -m pip install -r requirements_lint.txt
       - name: Lint with flake8
         run: flake8 . --count --show-source --statistics
 
--- a/ansible_events/event_filters/json_filter.py
+++ b/ansible_events/event_filters/json_filter.py
@@ -30,8 +30,9 @@ def matches_exclude_keys(exclude_keys, s):
     return False
 
 
-# FIXME(cutwater): Mutable instances are used as a default value of a function.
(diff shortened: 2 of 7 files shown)


Alex-Izquierdo:
I suggest, maybe in another PR, use as well:
flake8-builtins 
flake8-eradicate
flake8-debugger
flake8-comprehensions

Alex-Izquierdo:
should be this plugin be installed in the pre-commit config too?

cutwater:
@Alex-Izquierdo That's a good idea. My current plan is to add these plugins iteratively and fix issues, discovered by each of the plugins in the same PR.
Thanks for sharing that list. Found couple of new plugins, I didn't know about.

Regarding `pre-commit`, I'll add it depending on which PR is merged first.

benthomasson (approved):

PR #113, opened 2022-07-15
Starts sphinx docs

benthomasson:


docs/actions.rst
```diff
@@ -0,0 +1,15 @@
+=======
+Actions
+=======
+
+Event driven automation supports several built-in actions that can be run from rules.  These are:
+
+    * run_playbook - run a playbook using ansible-runner
```
Alex-Izquierdo, line 7: Should be documented the params required or accepted by these actions?

docs/usage.rst
```diff
@@ -2,6 +2,20 @@
...
+    ansible-events --help
+
+The normal method for running `ansible-events` is the following::
+
+    ansible-events --inventory inventory.yml --rules rules.yml
+
+If you are using custom event source plugins use the following::
+
+
+    ansible-events --inventory inventory.yml --rules rules.yml -S sources/
```
mkanoor, line 19: @benthomasson Should we mention the --vars here to pass in global variables if any.

tests/sources/template.py
```diff
@@ -0,0 +1,32 @@
+"""
+template.py
+
+An ansible-events event source plugin template.
+
+Arguments:
+    - delay: seconds to wait between events
```
mkanoor, line 7: @benthomasson Can any other others be passed into the source plugin? How would the user specify the,

mkanoor: @benthomasson Should all sources have a name, version, description section. A info or metadata method that can be invoked by the engine as it loads up the source plugins.

docs/rules.rst
```diff
@@ -0,0 +1,25 @@
...
+    name:  An automatic remedation rule
+    condition:  event.outage == True
+    action:
+        run_playbook:
+            name: remediate_outage.yml
+
+This rule searches for a recent event with the data `outage` being `True`. If an event
+with this condition is found then the `remediate_outage.yml` playbook is run.
+
+
```
mkanoor, line 22: @benthomasson What are the constraints when multiple rules are specified in a single file. Can the events be reused across rules. Do rule names have to be unique within a file?

README.rst
```diff
@@ -81,7 +81,11 @@ remediate the issue.
 Features
 --------
 
+* Connect to event streams and handle events in near real time.
 * Conditionally launch playbooks based on rules that match events in event streams.
+* Store facts about the world from data in events
+* Limit the hosts where playbooks run based on event data
+* Run smaller jobs that run more quickly by limiting the hosts where playbooks run based on event data
```
bzwei, line 88: I don't understand line 88, specially when comparing with line 87

docs/usage.rst
```diff
@@ -2,6 +2,20 @@
...
+
+Here `sources` is a directory containing your event source plugins.
```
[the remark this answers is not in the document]
benthomasson: From a collection FQCN should work.

tests/sources/template.py
benthomasson: Arguments can be anything that is valid YAML.

benthomasson: Arguments are passed as a dictionary to the source plugin.   If it is a valid dictionary it can be passed.   The keys are not restricted.

benthomasson: Yes, that metadata is a good idea.   I have not needed it yet so I have not implemented it.    We could use the python standards for modules or Ansible standards.   I am fine with either.

docs/rules.rst
benthomasson: Each rule set is a separate namespace.    An event only matches one rule.    If we want to match multiple rules we need to promote it to a fact.

benthomasson: Rule names are just documentation.   It is probably a good idea if they are unique, but I don't enforce that anywhere.

mkanoor: If a rule names is not provided durable rules adds a default name of r_{index} e.g r_0

PR #117, opened 2022-07-20
Support events in print_event

mkanoor:
The print_event action currently only supports a single event and fails when invoked with multiple events. This PR adds support for multiple events.

https://issues.redhat.com/browse/AAP-4829

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -649,3 +649,31 @@ async def test_25_max_attributes_nested():
     event = event_log.get_nowait()
     assert event["type"] == "Shutdown", "2"
     assert event_log.empty()
+
+
+@pytest.mark.asyncio
+async def test_26_print_events():
+    ruleset_queues, queue, event_log = load_rules(
+        "examples/26_print_events.yml"
+    )
+
+    queue.put_nowait(dict(i=1))
+    queue.put_nowait(dict(i=2))
+    queue.put_nowait(Shutdown())
+
+    await run_rulesets(
+        event_log,
+        ruleset_queues,
+        dict(),
+        dict(),
+    )
+
+    event_log.get_nowait()
+    event = event_log.get_nowait()
+    assert event["type"] == "Action", "1"
+    assert event["action"] == "print_event", "2"
+    event = event_log.get_nowait()
+    assert event["type"] == "ProcessedEvent", "1"
+    event = event_log.get_nowait()
+    assert event["type"] == "Shutdown", "7"
+    assert event_log.empty()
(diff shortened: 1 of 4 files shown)


mkanoor:
@benthomasson Please review

benthomasson (approved):

benthomasson (changes requested):
Add a unit test for this.

mkanoor:
@mkanoor Added multiple events test

eclarizio:
@benthomasson @mkanoor This is outside the scope of this PR, but I've noticed that this `test_examples.py` file is getting pretty large and seems a bit unwieldy to me. We already have two PRs both adding tests to this and one of them will have to change its numbering depending on whose goes in first. Not that that's a huge deal, but it feels odd that we would have to do that any time two PRs adding examples are in flight. 

I'm not sure what the best solution is, but it feels to me like these should be contained within the `examples` folder with their corresponding yaml files and separated out into separate test files. I think this would make it much easier to maintain and (perhaps more importantly) lookup/search which ones you may care about later at a glance instead of digging through a 600+ line file.

Maybe I'm wrong, would like to hear your opinions.

tests/test_examples.py
```diff
@@ -563,3 +563,31 @@ async def test_21_run_playbook():
     event = event_log.get_nowait()
     assert event["type"] == "Shutdown", "8"
     assert event_log.empty()
+
+
+@pytest.mark.asyncio
+async def test_22_print_events():
```
benthomasson, line 569: Please change the name to test_26_print_events

benthomasson (dismissed):

benthomasson:
@eclarizio the examples directory is a place to simplified examples of different features of the rule language that we want to support.   It should not be used a place to put all the unit tests.  We should use other test files for normal unit tests.   For these two PRs they are good examples of language features and can go into this directory.    They do need to be renamed as they already conflict with other tests that I added in the meantime.

eclarizio:
> @eclarizio the examples directory is a place to simplified examples of different features of the rule language that we want to support. It should not be used a place to put all the unit tests. We should use other test files for normal unit tests. For these two PRs they are good examples of language features and can go into this directory. They do need to be renamed as they already conflict with other tests that I added in the meantime.

Yeah, that's fair, I wasn't sure if organizationally my idea was feasible or good, but ultimately running into issues with "conflicting" names seemed to poke at a potentially deeper issue that could be solved fairly easily. Thanks! 👍

tests/test_examples.py
cutwater: Why tests are numbered like this?

benthomasson: They match up with examples that are numbered.

cutwater: Ah, alright, this makes sense.

mkanoor:
@benthomasson Please review I have fixed the example file names

benthomasson (approved):

PR #118, opened 2022-07-20
add ansible as a required dependency

Alex-Izquierdo:
Add ansible as a required dependency: https://github.com/benthomasson/ansible-events/blob/main/ansible_events/collection.py#L17

Also removes ansible-runner from test-requirements (defined also as a required dependency) and ansible-core (installed by ansible)

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -2,9 +2,6 @@ watchdog
 docopt
 psutil
 requests
-ansible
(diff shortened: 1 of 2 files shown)


benthomasson:
Ansible-runner does not include ansible as a package dependency so that the user can install their own version of ansible.  We should follow the same convention here.

PR #119, opened 2022-07-25
Support var_root in multi events

mkanoor:
var_root is used to optionally lift up the keys in a nested events
dictionary.
The var_root has been modified to take in a sigle nested key in the form
of a.b.c or or a list of keys of the form
- a.b.c
- x.y.x

When we have matching on multiple events from different sources the
event payload could have different keys. The var_root can store the
nested keyname for different payloads and pick them from different
events.

If the var_root has the following keys
   ['webhook.payload', 'kafka.message']

it can modify the events from

{'events': {'webhook': {'webhook': {'payload': {'url': 'http://www.example.com', 'action': 'merge'}}},
            'kafka': {'kafka': {'message': {'topic': 'testing', 'channel': 'red'}}}}

to

{'events': {'webhook': {'url': 'http://www.example.com', 'action': 'merge'},
            'kafka': {'topic': 'testing', 'channel': 'red'}}

### DIFF
--- a/ansible_events/builtin.py
+++ b/ansible_events/builtin.py
@@ -10,7 +10,7 @@ import tempfile
 import uuid
 from functools import partial
 from pprint import pprint
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional, Union
 
 import ansible_runner
 import dpath.util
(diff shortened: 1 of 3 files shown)


mkanoor:
@benthomasson Please review

tests/test_examples.py
```diff
@@ -563,3 +563,39 @@ async def test_21_run_playbook():
     event = event_log.get_nowait()
     assert event["type"] == "Shutdown", "8"
     assert event_log.empty()
+
+
+@pytest.mark.asyncio
+async def test_23_multiple_events_var_root():
```
benthomasson, line 569: Please change the name to test_27_multiple_events_var_root

benthomasson (changes requested):

ansible_events/builtin.py
```diff
@@ -56,16 +56,20 @@ async def print_event(
     variables: Dict,
     facts: Dict,
     ruleset: str,
-    var_root: Optional[str] = None,
+    var_root: Optional[Union[str, List[str]]] = None,
```
cutwater, line 59: The type hint here can be slightly simplified as :

```Union[str, List[str], None]```

since `Optional[T]` is the same as `Union[T, None]`

mkanoor:
Closed in favor of https://github.com/benthomasson/ansible-events/pull/126

PR #121, opened 2022-07-26
Fixed a bug in non string type in facts

mkanoor:
- Missing a return statement
- Run a playbook to check the fact values
- After we run a playbook we should be able to return the rc and status

### DIFF
--- a/tests/examples/04_assert_fact.yml
+++ b/tests/examples/04_assert_fact.yml
@@ -12,6 +12,8 @@
         assert_fact:
           fact:
             msg: hello world
+            alpha: 1
+            beta: 2
     - name:
       condition: event.msg == "hello world"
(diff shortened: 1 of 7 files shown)


mkanoor:
@benthomasson Please review

tests/examples/04_assert_fact.yml
```diff
@@ -12,6 +12,8 @@
         assert_fact:
           fact:
             msg: hello world
+            alpha: 1
```
benthomasson, line 15: Make a new test for this instead of changing one of the examples.  Add the test to test_engine instead of test_examples.

benthomasson (changes requested):

benthomasson (approved):

PR #122, opened 2022-07-28
Added a playbook that can validate passed in args

mkanoor:
- Allows for easy testing by specifying validate_args
- Uses jmespath to query nested dictionary keys
- Any extra kwargs to run playblook will be added to the variables
- To aid in looping with ansible uses an extra playbook via
  include_tasks


mkanoor:
@benthomasson Please review

benthomasson (approved):

PR #123, opened 2022-07-28
Raise exception if a playbook fails execution

mkanoor:
1. A user can supress the exception by setting ignore_errors
2. Fixed tests where the playbook was failing during execution
3. Added tests for ignore_errors


mkanoor:
@benthomasson Please review

benthomasson (changes requested):
I am not sure that raising an exception is the right thing to do on a failing playbook.   We should generate a new out going event that describes the failure so that the UI can update the database about the failure.

benthomasson:
@mkanoor this needs rebasing.

mkanoor:
Closed in favor of #132

PR #124, opened 2022-08-01
Fixed package name imports

mkanoor:
This would help us when we switch to drools.
Places where we call durable.lang have been changed to lang


mkanoor:
@benthomasson Please review

benthomasson (approved):

PR #127, opened 2022-08-02
Drools Support: Add `durable-rules` adapter invoking a REST service

evacchi:
A clone of the original `durable-rules`, where `durable_rules_engine` is Python library invoking a REST endpoint.
The REST endpoint is supposed to be https://github.com/mariofusco/drools-yaml-rules and/or a compatible API

This PR also includes @mkanoor's code for switching between the engines: when `RULES_ENGINE` == `drools` then the Drools backend will be used; otherwise it will fallback to the usual `durable_rules` 

I have also added a7a0dca to pick a different host `ANSIBLE_EVENTS_DROOLS_HOST` than localhost:8080


mkanoor:
@benthomasson Please review

Alex-Izquierdo:
I would say that there is a misuse of comments.

evacchi:
@Alex-Izquierdo I think I have addressed most of those
@mkanoor I have pushed the changes you have sent to me, but I haven't pushed the changes to engine.py + lang.py because I think we may want to keep verbatim (in fact I think I left some print statements around that I should revert back)

Notice the one main contribution of this PR is durable_rules_engine.py while the rest is mainly:
- verbatim code from durable-rules (with only imports changed)
- code from @mkanoor to enable switching between engines via ENV var

evacchi:
Follow up: I have now pushed the changes for engine.py + lang.py + additional changes to allow it to hopefully pass all the linter-related issues

tests/test_rules.py
```diff
@@ -4,8 +4,19 @@
 
 import pytest
 import yaml
-from durable import lang
-from durable.lang import assert_fact, c, m, post, rule, ruleset, when_all
+
+
+if os.environ.get("RULES_ENGINE", "durable_rules") == "drools":
+    import ansible_events.drools as durable
```
mkanoor, line 10: @evacchi I think this line is not needed, I added it incorrectly

ansible_events/builtin.py
```diff
@@ -17,6 +17,11 @@
 import yaml
 from durable import lang
```
mkanoor, line 18: @evacchi We don't need this line.
Can you run flake8 . on your repo
You will get all the linter issues.
You might have to tweak your local tox.ini so it excludes the venv directory 
I have a PR for that https://github.com/benthomasson/ansible-events/pull/129

cutwater:
@mkanoor @evacchi If this is a vendorized dependency, I think it should be put into a package that explicitly indicates that (i.e. `ansible_events.vendor`). This will improve code readability and allow us to configure linting exceptions for vendorized dependencies.

evacchi:
@cutwater good point. Indeed drools.engine and drools.lang are vendorized. `durable_rules_engine.py` is custom code, that replaces the original C lib. So maybe we may create:

- drools/vendor/engine.py
- drools/vendor/lang.py
- drools/durable_rules_engine.py

- drools/vendor/* will have to reference drools/durable_rules_engine.py though

evacchi:
@cutwater I have moved the code as explained in https://github.com/benthomasson/ansible-events/pull/127#issuecomment-1203993945
@mkanoor I have double checked and indeed some of those imports were incorrect, it should be all good now

evacchi:
(forgot to run the linter on tests 😬)

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
+import json
```
cutwater, line 1: @evacchi Is it a vendored code too?

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
...
+import logging
+
+import requests
+
+_logger = logging.getLogger(__name__)
+
+
+class DurableRulesEngine:
+    __host = None
```
cutwater, line 10: Is there any specific reason to use double underscore prefix for these class attributes?

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
...
+    __last_resp = None
```
cutwater, line 11: Please initialize instance attributes in the `__init__` method instead of creating class attributes with a default value.

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def __init__(self, host):
+        self.__host = host
+
+    def create_ruleset(self, ruleset_name, ruleset_string):
+        req = {ruleset_name: json.loads(ruleset_string)}
+
+        r = requests.post(
```
cutwater, line 19: It seems like single character names are heavily overused here and below. Could you use more verbose names, like:

* `r -> response`
* `d -> data`

etc.?

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
...
+
+        self.__last_resp = r.json()
+        self.__last_resp.reverse()
+
+        return (0, session_id)
+
+    def start_action_for_state(self, handle):  # real signature unknown
+        try:
+            resp = self.__last_resp.pop()
+        except Exception:
```
cutwater, line 61: Too broad exception handler. Please specify the exact exception classes that are expected here.

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
...
+            return None
+
+        return ('{ "sid":"0", "id":"sid-0", "$s":1}', json.dumps(resp), handle)
+
+    def complete_and_start_action(self, handle):  # real signature unknown
+        _logger.info("complete_and_start_action: %d", handle)
+        try:
+            resp = self.__last_resp.pop()
+        except Exception:
```
cutwater, line 70: same here.

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,230 @@
...
+
+    __weakref__ = property(
+        lambda self: object(), lambda self, v: None, lambda self: None
+    )  # default
+    """list of weak references to the object (if defined)"""
+
+
+# exported methods
+
+__instance = DurableRulesEngine("http://localhost:8080")
```
cutwater, line 114: Why double underscore prefix is used here?

cutwater (changes requested):

ansible_events/drools/vendor/__init__.py
```diff
@@ -0,0 +1,3 @@
+import logging
```
cutwater, line 1: This logger doesn't seem to be used. Why is it created here?

ansible_events/drools/durable_rules_engine.py
mkanoor: @cutwater Most of this code comes from durable_rules we might have to clean it up further down the road.

evacchi: this is actually my own code. It is PoC quality at this time. I'll fix it. BTW the intended usage here is to catch when the list is empty. Probably better to check for emptiness instead :)

ansible_events/drools/durable_rules_engine.py
evacchi: no special reason. Would a single underscore be better for a private global?

ansible_events/drools/vendor/__init__.py
evacchi: just a bad copy paste

ansible_events/drools/durable_rules_engine.py
evacchi: using single underscore, added read from env var, so that is configurable

ansible_events/drools/durable_rules_engine.py
evacchi: sure

mkanoor:
@evacchi You might want to run the following commands

1. isort .
2. black .

The first one fixes the order of imports and the 2nd one resolves any formatting issues.

evacchi:
@mkanoor done!

mkanoor:
@benthomasson Please review

benthomasson:
@evacchi This code looks good but I cannot test it with drools because I am missing drools version 8.26.0-SNAPSHOT.  I attempted to build that and got a failure in the tests:

[INFO] Drools :: Templates ................................ FAILURE [  2.463 s]

benthomasson:
Also this error: 

[INFO] KIE :: Decision Model Notation :: Validation ....... FAILURE [  8.888 s]

ansible_events/drools/durable_rules_engine.py
```diff
@@ -0,0 +1,235 @@
...
+
+        if self._last_response:
+            resp = self._last_response.pop()
+        else:
+            return None
+
+        return ('{ "sid":"0", "id":"sid-0", "$s":1}', json.dumps(resp), handle)
+
+    def complete_and_start_action(self, handle):  # real signature unknown
+        _logger.info("complete_and_start_action: %d", handle)
```
mkanoor, line 69: @evacchi The %d should be %s handle has a type of str
<class 'str'>

evacchi:
> @evacchi This code looks good but I cannot test it with drools because I am missing drools version 8.26.0-SNAPSHOT. I attempted to build that and got a failure in the tests:
> 
> [INFO] Drools :: Templates ................................ FAILURE [ 2.463 s]

I think you should be able to build the latest main with `mvn clean install -Dquickly`

ansible_events/drools/durable_rules_engine.py
evacchi: fixed

mkanoor:
@evacchi @benthomasson I tried the following commands to run the drools server and it worked for me

You need maven, java installed on your machine

1. git clone https://github.com/mariofusco/drools-yaml-rules
2. git clone https://github.com/kiegroup/drools
3. cd drools
4. mvn install -Dquickly
5. cd ../drools-yaml-rules
7.  mvn install
8. cd drools-yaml-rules-durable 
9. mvn quarkus:dev

evacchi:
> @evacchi @benthomasson I tried the following commands to run the drools server and it worked for me
[...]

hey good summary, I'll add this to the README.md under `drools`

evacchi:
FYI: You can also download a snapshot of the last working build from https://github.com/mariofusco/drools-yaml-rules/actions/

e.g. click a green build and then download the artifact

Screen Shot 2022-08-04 at 11 13 02

then run the jar with `java -jar *-runner.jar`

mkanoor:
@evacchi I downloaded the jar and tested it. It works. Is there a way to get a stable URL for downloading that tar.
Do we need a shell script start_drools.sh and stop_drools.sh that can 
1. Check if Java is installed on the machine check java version
2. Open port (??) 
3.  java -jar drools-yaml-rules-durable-1.0.0-SNAPSHOT-runner.jar
4. Store the pid which can be used to stop it.

benthomasson (approved):

benthomasson:
It works for me with the downloaded jar.

evacchi:
> @evacchi I downloaded the jar and tested it. It works. Is there a way to get a stable URL for downloading that jar.
In a separate PR do we need a shell script start_drools.sh and stop_drools.sh that can

that's why I want to actually create a GH action for releases, that way it _will_ have a stable URL. With a plain GHA build getting a URL is a bit more involved https://docs.github.com/en/rest/actions/artifacts

creating a script should be easy enough. It could print out the instructions if the jar is missing for now :)

PR #129, opened 2022-08-02
Added flake8 config setting

mkanoor:
When we run flake8 locally we want to be able to skip
the venv and .venv directories.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,7 +61,7 @@ search = __version__ = '{current_version}'
 replace = __version__ = '{new_version}'
 
 [flake8]
-extend-exclude = docs
+extend-exclude = docs, venv, .venv
 # show-source = True
 # Flake8 default ignore list:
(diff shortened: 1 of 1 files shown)


mkanoor:
@benthomasson Please review

cutwater (changes requested):
1. Flake8 settings are stored in the `setup.cfg`. Please use this configuration file if further `flake8` customization is required.
2. Flake8 excludes `.git` and `__pycache__` directories by default. Setting `exclude` option overrides default exclude list. Please use `extend-exclude` to add excluded directories.

setup.cfg
```diff
@@ -61,7 +61,7 @@ search = __version__ = '{current_version}'
 replace = __version__ = '{new_version}'
 
 [flake8]
-extend-exclude = docs
+extend-exclude = docs, venv
```
cutwater, line 64: Should it include `.venv` as well?

cutwater (changes requested):

cutwater (approved):
Please don't forget to squash your commits before merging.

benthomasson (approved):

PR #133, opened 2022-08-04
Fixed the source in performance_test to be asyncio compatible

mkanoor:
ERROR:root:Source error Entrypoint is not a coroutine function.

We had fixed this in tests/sources but not in preformance_test/sources


mkanoor:
@benthomasson Please review

benthomasson (approved):

PR #134, opened 2022-08-05
Drools Support: Add support to post_event

evacchi:
Currently "emulated" via "assert_fact", the change at https://github.com/mariofusco/drools-yaml-rules/commit/b9ee5331e7a5bca9856b47507e10a3c58932cb19 ensures that the fact is actually being removed from the WM as soon as there is one match. We are also replacing the endpoint names to avoid confusion https://github.com/mariofusco/drools-yaml-rules/commit/03eb912d11e0b304e51b0281b2902c1937a29f9b

You can download the latest snapshot at (FIXED link) https://github.com/mariofusco/drools-yaml-rules/actions/runs/2876048896

### DIFF
--- a/ansible_events/drools/durable_rules_engine.py
+++ b/ansible_events/drools/durable_rules_engine.py
@@ -28,10 +28,19 @@ class DurableRulesEngine:
         return response.text
 
     def assert_event(self, session_id, serialized_fact):
-        _logger.warning("assert_event not yet implemented: using assert_fact")
-        return self.assert_fact(session_id, serialized_fact)
+        return self._process_message(
+            session_id, serialized_fact, "process-events"
+        )
 
     def assert_fact(self, session_id, serialized_fact):
+        return self._process_message(
+            session_id, serialized_fact, "process-facts"
+        )
+
+    def _process_message(self, session_id, serialized_fact, command):
+        if command not in ["process-events", "process-facts"]:
+            raise Exception("Unknown command " + command)
+
         fact = json.loads(serialized_fact)
         if "j" in serialized_fact:
             fact["j"] = 1
@@ -39,7 +48,7 @@ class DurableRulesEngine:
         serialized_fact = json.dumps(fact)
 
         response = requests.post(
-            f"{self._host}/rules-durable-executors/{session_id}/process",
+            f"{self._host}/rules-durable-executors/{session_id}/{command}",
             json=json.loads(serialized_fact),
         )
 
@@ -124,7 +133,7 @@ def abandon_action(*args, **kwargs):  # real signature unknown
 
 
 def assert_event(session_id, serialized_fact):
-    return _instance.assert_fact(session_id, serialized_fact)
+    return _instance.assert_event(session_id, serialized_fact)
 
 
 def assert_events(*args, **kwargs):  # real signature unknown


mariofusco (approved):

mkanoor:
@benthomasson Please review

mkanoor:
@evacchi I think this method should call _instance.assert_event instead of _instance.assert_fact
```
def assert_event(session_id, serialized_fact):
    return _instance.assert_fact(session_id, serialized_fact)
```

evacchi:
ouch you are right, that's the one place where it gets actually dispatched 🤦

mkanoor:
@mariofusco @evacchi 
When we change from process_facts to process_events the speed improves but we get failures in our tests

pytest tests/test_examples.py::test_15_multiple_events_all

In this test we have a condition that looks like this 
```
condition:
        all:
          - event.nested.i == 1
          - event.nested.j == 1
```

And the events sent to the system look like

{'nested': {'i': 1, 'j': 0}}
{'nested': {'i': 0, 'j': 1}}

if we use process_facts it works if we change it to process_events it fails since the events get removed as soon as it matches.
You had mentioned that if it is an event it needs to have a temporal component should that be provided by us or does the drools engine add that automatically.

ansible_events/drools/durable_rules_engine.py
```diff
@@ -124,7 +133,7 @@ def abandon_action(*args, **kwargs):  # real signature unknown
 
 
 def assert_event(session_id, serialized_fact):
-    return _instance.assert_fact(session_id, serialized_fact)
+    return _instance.assert_event(session_id, serialized_fact)
```
mkanoor, line 136: @evacchi This causes test failures when multiple events are involved

mkanoor:
@mariofusco @evacchi @benthomasson 

There is a difference between how we handle multiple events and single events with multiple checks with an and

In the example below we are talking about a single event and both the checks have to pass for the condition to be true
```
condition: event.nested.i == 1 and event.nested.j == 1
```


In the case below we are talking about 2 distinct events which will occur at different times. This introduces a timeout and both the conditions have to match within a specified time and if they don't match within that time we should get a timeout error.
```
condition:
        all:
          - event.nested.i == 1
          - event.nested.j == 1
```

I think a good use case for this would be if you are expecting a service to go down for a brief moment of time for applying maintenance and then comes back up

```
condition:
        all:
          - event.message == "Service Stopped"
          - event.message == "Apply Maintenance"
          - event.message == "Service Started"
        timeout: 300 seconds
```

If one of the condition matches we should start a timer and cache all the facts till the timer expires. If within that time we get all the events we have a success, If the timer expires and all events have not matched we should raise a timeout error which the user can use to post an event that the Service has failed and take some corrective action to notify an IT personnel.

With the assert_event change in drools yaml we are clearing the cache of events as soon as one of them matches. I think for conditions that are wrapped in a timeout we would need to keep those events till the timer expires or all events have been matched.

evacchi:
interesting, I think it all boils down to a semantic mismatch, that was implemented in a way that it's too naive on the Java end /cc @mariofusco we'll have to think if there is a simple way to realize this or we should start using CEP w/ explicit expirations etc.

mariofusco:
@evacchi ok, this makes sense and I know how to fix this, but it will take me a bit of time.
@mkanoor I see the problem and I will fix it. I'm on PTO this week, if this is not blocking you and not extremely urgent I'll fix it next week and keep you updated.

evacchi:
I think this can wait @mariofusco :)

mkanoor:
@evacchi @mariofusco 
I was reading this article https://docs.jboss.org/drools/release/6.2.0.CR3/drools-docs/html/DroolsComplexEventProcessingChapter.html and the duration concept matches the timeout that we were talking about from durable rules

evacchi:
@mkanoor the issue should be solved, you can download the new jar (I have fixed the link at the top)

mkanoor:
@evacchi Would this work with JPY or do we have to start the REST API server?

mkanoor:
@evacchi Tested the tests with this change and they pass.

mariofusco (approved):

benthomasson (approved):

PR #139, opened 2022-08-14
improve jinja support and fix AAP-5176

Alex-Izquierdo:
Fixes https://issues.redhat.com/browse/AAP-5176

There is no need to use dpath since jinja render can get the data from dicts with dot notation. In this way we add real support to jinja templates in the right side on the condition including jinja filters and arithmetic operations.

### DIFF
--- a/ansible_events/util.py
+++ b/ansible_events/util.py
@@ -2,7 +2,6 @@ import os
 from pprint import pprint
 from typing import Any, Dict, List, Union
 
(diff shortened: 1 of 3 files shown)


Alex-Izquierdo:
@benthomasson please review

benthomasson (approved):
I forget why I added 25-27 but if it didn't fail a test I guess I didn't really need it.

PR #144, opened 2022-08-23
[AAP-5370] Support for Drools via JPY and a separate python package

mkanoor:
Uses a separate drools_jpy package

* Java 11+ installed
* Environment variable JAVA_HOME set accordingly
* Maven 3.8.1+ installed

JPY has a wheel package that bundles the binaries but might have mismatches with your platform. To be reliable we would have to build the package from source so that the shared object is for your platform.
To force a build from source we need to set
**export PIP_NO_BINARY=jpy**

https://issues.redhat.com/browse/AAP-5370


mkanoor:
@benthomasson Please review externalized the drools_jpy package
The drools_jpy package is stored in pypi.org the source code is at https://github.com/mkanoor/drools_jpy
I have also added an action to run the tests via drools.

benthomasson:
@mkanoor add me (benthomasson) as a collaborator on https://pypi.org/project/drools-jpy/

Alex-Izquierdo:
@mkanoor as far as I know, there is no need to set any envvar, just to put in `setup.cfg`
```ini
[global]
no-binary = jpy
```

Have you tried it?

setup.cfg
```diff
@@ -40,6 +40,7 @@ install_requires =
 	janus
 	ansible-runner
 	websockets
+        drools_jpy
```
Alex-Izquierdo, line 43: This line is not needed because you have defined the dependency in the setup.cfg. isn't?

.github/workflows/tests.yml
```diff
@@ -1,5 +1,10 @@
 name: Tests
 
+# The wheel package of jpy is not reliable and causes
+# run time errors, Forcing to build from source
+env:
+  PIP_NO_BINARY: jpy
```
Alex-Izquierdo, line 6: I think there is no need to setup the envvar, https://github.com/benthomasson/ansible-events/pull/144#issuecomment-1226043523

Alex-Izquierdo (changes requested):

setup.cfg
mkanoor: @Alex-Izquierdo This is in setup.cfg

.github/workflows/tests.yml
mkanoor: @Alex-Izquierdo the jpy package comes in source and wheel distribution. The wheel distribution has a shared library for Java Native Interface which loads the share library for jvm. That has some issues which the jpy team is looking into. Till then using the src package gets us up an running. This environment variable is needed so that pip can skip the wheel and download the tar
```
Skipping wheel build for jpy, due to binaries being disabled for it.
```

setup.cfg
Alex-Izquierdo: @mkanoor  Sorry, I mistake the line. I meant in the dockerfile: https://github.com/benthomasson/ansible-events/pull/144/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R19

.github/workflows/tests.yml
Alex-Izquierdo: @mkanoor Yeah, I'm saying that there is no need to deal with the envvar and you can ensure the compilation with the configuration in `setup.cfg` that I mentioned before.

mkanoor: 👍

mkanoor: @Alex-Izquierdo I tried setting the global in setup.cfg and it ended up downloading the wheel
```
Downloading jpy-0.11.0-cp310-cp310-macosx_10_15_universal2.whl (185 kB)
```

Alex-Izquierdo: yeah, I'm doing some tests and doesn't seem to work. :(

setup.cfg
mkanoor: 👍

.github/workflows/tests.yml
benthomasson: @Alex-Izquierdo what doesn't work for you?

benthomasson (approved):
This seems to work for me with jpy binaries disabled.   Building them was fine since I had Java already set up on my system.

.github/workflows/tests.yml
Alex-Izquierdo: Sorry, it didn't work the configuration that I've proposed: https://github.com/benthomasson/ansible-events/pull/144#issuecomment-1226043523 not the PR

Alex-Izquierdo (approved):

PR #146, opened 2022-08-25
Adds a worker mode to ansible-events to support container deployment

benthomasson:
Supports [AAP-5269](https://issues.redhat.com/browse/AAP-5269)

Needed for https://github.com/benthomasson/ansible-events-ui/pull/74

### DIFF
--- a/ansible_events/app.py
+++ b/ansible_events/app.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import os
+from asyncio.exceptions import CancelledError
 from typing import Any, Dict, List, Tuple
 
 import yaml
@@ -14,18 +15,29 @@ from ansible_events.collection import (
 from ansible_events.engine import run_rulesets, start_source
 from ansible_events.rule_types import RuleSet, RuleSetQueue
 from ansible_events.util import load_inventory
-from ansible_events.websocket import send_event_log_to_websocket
+from ansible_events.websocket import (
+    request_workload,
+    send_event_log_to_websocket,
+)
 
 logger = logging.getLogger("ansible_events")
 
 
 # FIXME(cutwater): Replace parsed_args with clear interface
 async def run(parsed_args) -> None:
-    inventory = {}
-    variables = load_vars(parsed_args)
-    rulesets = load_rules(parsed_args)
-    if parsed_args.inventory:
-        inventory = load_inventory(parsed_args.inventory)
+
+    if parsed_args.worker and parsed_args.websocket_address and parsed_args.id:
+        logger.info("Starting worker mode")
+
+        inventory, variables, rulesets = await request_workload(
+            int(parsed_args.id), parsed_args.websocket_address
+        )
+    else:
+        inventory = {}
+        variables = load_vars(parsed_args)
+        rulesets = load_rules(parsed_args)
+        if parsed_args.inventory:
+            inventory = load_inventory(parsed_args.inventory)
 
     event_log = asyncio.Queue()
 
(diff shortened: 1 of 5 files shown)


benthomasson:
@cutwater @bzwei please review

ansible_events/app.py
```diff
@@ -57,7 +69,10 @@ async def run(parsed_args) -> None:
     logger.info("Cancelling event source tasks")
     for task in tasks:
         task.cancel()
-    await asyncio.gather(*tasks)
+    try:
+        await asyncio.gather(*tasks)
```
cutwater, line 73: I have doubts about this particular logic. If first task raises the `CancelledError`, the remaining tasks may not be cleaned up properly. I need to verify this.

ansible_events/key.py
```diff
@@ -0,0 +1,39 @@
...
+        logger.debug(cmd_args)
+
+        proc = await asyncio.create_subprocess_exec(
+            ssh_add,
+            *cmd_args,
+            cwd=local_working_directory,
+        )
+
+        await proc.wait()
+    finally:
```
cutwater, line 37: Please do not use ```try .. finally: pass```. It doesn't do anything, only increases code depth and makes it reduces readability.

ansible_events/key.py
```diff
@@ -0,0 +1,39 @@
...
+        # shutil.rmtree(local_working_directory)
```
cutwater, line 38: Dead code

ansible_events/key.py
```diff
@@ -0,0 +1,39 @@
...
+import tempfile
+
+logger = logging.getLogger()
+
+ssh_add = shutil.which("ssh-add")
+
+
+async def install_private_key(private_key):
+
+    local_working_directory = tempfile.mkdtemp()
```
cutwater, line 15: Better use `tempfile.TemporaryDirectory` that provides a context manager to ensure that the directory is deleted after the process is finished.

cutwater (changes requested):

ansible_events/key.py
```diff
@@ -0,0 +1,39 @@
...
+    try:
+
+        key_file_fd = os.open(
+            os.path.join(local_working_directory, "key"),
+            os.O_WRONLY | os.O_CREAT | os.O_EXCL,
+            mode=stat.S_IRUSR | stat.S_IWUSR,
```
cutwater, line 21: A comment explaining combination of flags would be nice here. Something like:

`# Create a file for write with mode 0600`

ansible_events/websocket.py
```diff
@@ -1,12 +1,60 @@
...
+
+from ansible_events import rules_parser as rules_parser
+from ansible_events.key import install_private_key
 
 logger = logging.getLogger("ansible_events.websocket")
 
 
+async def request_workload(activation_id, websocket_address):
+    logger.info("websocket %s connecting", websocket_address)
+    async for websocket in websockets.connect(websocket_address):
```
cutwater, line 17: This code looks odd. Can you please explain the logic here?

If that's a one time connection, you should use `async with websocket.connect(...) as websocket`.
If the intention is to automatically reconnect on disconnects, the `websockets.ConnectionClosed` exception should be handled.

ansible_events/app.py
cutwater: UPD: The `CancelledError` is raised out of `asyncio.gather()` after all all tasks completed, so the described above behaviour is not an issue. However it may be a good idea to call `asyncio.gather(*tasks, return_exceptions=True)` to log any exception that is distinct from `CancelledError`.

bzwei (dismissed):

ansible_events/app.py
```diff
@@ -57,7 +69,11 @@ async def run(parsed_args) -> None:
     logger.info("Cancelling event source tasks")
     for task in tasks:
         task.cancel()
-    await asyncio.gather(*tasks)
+    exceptions = await asyncio.gather(*tasks, return_exceptions=True)
+    for exception in exceptions:
+        if isinstance(exception, CancelledError):
+            continue
+        logger.error(exception)
```
cutwater, line 76: Minor:
```suggestion
        if not isinstance(exception, CancelledError):
            logger.error(exception)
```

cutwater (dismissed):

PR #147, opened 2022-08-25
[AAP-5126] Allow to rerun a playbook on failure

bzwei:
Add option `retry: True` to rerun a playbook once on failure,
or `retries: <count>` to retry multiple times until succeed.
Option `delay: <seconds>` to pause a few seconds before rerun.

The same options can be applied to rerun a module.

See tests/examples/33_run_playbook_retry.yml or
tests/examples/34_run_playbook_retries.yml or
tests/examples/32_run_module_fail.yml for examples.

Resolves: AAP-5126

### DIFF
--- a/ansible_events/builtin.py
+++ b/ansible_events/builtin.py
@@ -134,6 +134,9 @@ async def run_playbook(
     var_root: Union[str, Dict, None] = None,
     copy_files: Optional[bool] = False,
     json_mode: Optional[bool] = False,
+    retries: Optional[int] = 0,
+    retry: Optional[bool] = False,
+    delay: Optional[int] = 0,
     **kwargs,
 ):
     logger = logging.getLogger()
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo (commented):
IMO the `retry` boolean option is redundant and is not needed. 
Following the ansible-playbook behavior, defining a number of retries enables retrying implicitly:
https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met

Also I would like to consider, @benthomasson @bzwei 
Is it necessary a retry option here when ansible-playbook provides that feature?

bzwei:
@Alex-Izquierdo Surely we can remove the boolean option, but I followed the requirement from AAP-5126. I think this feature is useful when we don't have the ability to edit the playbook.
I will change the option `retry_interval` to `delay`.

benthomasson (changes requested):
Please update schema/ruleset_schema.json with the new options.  Run a schema validator on  your examples to check it.

benthomasson (approved):

PR #148, opened 2022-08-26
improve sensu-kafka demo for mac users

Alex-Izquierdo:
Improve the sensu-kafka demo specially for mac users. Include:
- Renamed the dir (docker-compose uses by default the dirname as prefix)
- Build custom images to avoid volume mounts. (In mac the `:z` flag is not ignored if SElinux is not present and requires more complex installation of podman machine.)
- Removed the topic-creator as is not needed, instead added a dependency on `sensu-cli` pod to ensure the topic is created before running ansible-events
- Increase timeouts for healthchecks (has been detected startup issues in slow computers)


Alex-Izquierdo:
@mkanoor please review

mkanoor:
@Alex-Izquierdo Tested works on Mac the only thing missing was linking podman to docker. Should we document that?

Alex-Izquierdo:
@mkanoor instructions updated

benthomasson (approved):

PR #150, opened 2022-09-01
Run each ruleset in a separate asyncio task

bzwei:
Before the refactoring all events are processed sequentially. A
long running action will block the processing of an event from
another ruleset.

Now each ruleset is running independently in its own asyncio task.
Events from the same ruleset are still processed sequentially.

Resolves: AAP-5471


bzwei:
@benthomasson please review

benthomasson:
@bzwei please rebase

bzwei:
@benthomasson rebase is done

PR #156, opened 2022-09-13
Check if the exception value is None before logging it as error

### LINKED ISSUE #155: When an asyncio.gather returns we are not checking for None before we deem it as an exception
* ansible-events version: 0.6.0
* Python version: 3.10.6
* Operating System: Mac OS 

### Description

When running performance test there is an error reported
```
ansible-events --rules ./performance_test/1k_event_rules.yml --inventory ./performance_test/inventory1.yml -S ./performance_test/sources/
```
ERROR:ansible_events:None

### What I Did

```
ansible-events --rules ./performance_test/1k_event_rules.yml --inventory ./performance_test/inventory1.yml -S ./performance_test/sources/
```

mkanoor:
When a co-routine exists and it doesn't return anything, the gather function returns None. We assume the returned values are exceptions and log None as an error.
Fixed the code to check if the returned value is None before checking its exception type

Fixes #155

### DIFF
--- a/ansible_events/app.py
+++ b/ansible_events/app.py
@@ -71,7 +71,7 @@ async def run(parsed_args) -> None:
         task.cancel()
     exceptions = await asyncio.gather(*tasks, return_exceptions=True)
     for exception in exceptions:
-        if not isinstance(exception, CancelledError):
+        if exception and not isinstance(exception, CancelledError):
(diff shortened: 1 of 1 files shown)


mkanoor:
@benthomasson Please review

PR #158, opened 2022-09-20
Start a new plan queue for each event

bzwei:
This is a preparation for running events in parallel whenever possible.
    Before all actions resulting from events are placed in one queue (plan).
    It is not a problem when events are processed sequentially.
    
    Now each event will use a dedicated queue to store resulting actions.
    It helps to collect and concatenate results from these actions while
    allow events to be processed in parallel.
    
    This work also eliminates the use of data type RuleSetQueuePlan because
    we can directly convert a RuleSetQueue to EngineRuleSetQueuePlan without
    using RuleSetQueuePlan as a middle carrier.
    
    See Also: AAP-4436

### DIFF
--- a/ansible_events/engine.py
+++ b/ansible_events/engine.py
@@ -37,7 +37,6 @@ from ansible_events.rule_types import (
     EngineRuleSetQueuePlan,
     EventSource,
     RuleSetQueue,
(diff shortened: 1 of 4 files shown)


benthomasson:
@bzwei rebase please

PR #191, opened 2022-10-12
[AAP-6237]: Add ansible-rulebook CLI documentation

abaiken:
[AAP-6237:](https://issues.redhat.com/browse/AAP-6237)
Documentation: update docs/usage.rst

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -3,19 +3,60 @@ Usage
 =====
 
 
-`ansible-rulebook` is either used as a command line tool or used as a web service with `ansible-rulebook-ui`.  
+`ansible-rulebook` is either used as a command line tool or used as a web service with `ansible-rulebook-ui`.
+
+
+======================
+CLI Usage and Examples
+======================
+
+
+The `ansible-rulebook` CLI supports the following options::
+
+    Options:
+    -h, --help                  Show this page
+    -i, --inventory=<i>         Inventory
+    --rulebook=<r>              The rulebook file or rulebook from a collection
+    -S=<S>, --source_dir=<S>    Source dir
+    --vars=<v>                  Variables file
+    --env-vars=<e>              Comma separated list of variables to import
+                                from the environment
+    --debug                     Show debug logging
+    --verbose                   Show verbose logging
+    --version                   Show the version and exit
+    --websocket-address=<w>     Connect the event log to a websocket
+    --id=<i>                    Identifier
+    --worker                    Enable worker mode
+    --project-tarball=<p>       Project tarball
 
 To get help from `ansible-rulebook` run the following::
 
     ansible-rulebook --help
 
+To check the version of `ansible-rulebook` run the following::
+
+    ansible-rulebook --version
+
 The normal method for running `ansible-rulebook` is the following::
 
-    ansible-rulebook --inventory inventory.yml --rulebook rules.yml
+    ansible-rulebook --inventory inventory.yml --rulebook rules.yml --vars vars.yml
+
+.. note::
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -3,19 +3,52 @@ Usage
...
+
+    ansible-rulebook --version
+
 The normal method for running `ansible-rulebook` is the following::
 
     ansible-rulebook --inventory inventory.yml --rules rules.yml
 
-If you are using custom event source plugins use the following::
+.. note::
+    The `--rules` option requires the `--inventory` option.
```
Alex-Izquierdo, line 47: `--rules` flag has been changed to `--rulebook`
Ref: https://github.com/ansible/ansible-rulebook/pull/189

Alex-Izquierdo (changes requested):

docs/usage.rst
```diff
@@ -3,19 +3,53 @@ Usage
...
+======================
+CLI Usage and Examples
+======================
+
+
+The `ansible-rulebook` CLI supports the following options::
+
+    Options:
+    -h, --help                  Show this page
+    -v, --vars=<v>              Variables file
```
Alex-Izquierdo, line 18: Duplicated, `-v`is not supported yet

Alex-Izquierdo (changes requested):

docs/usage.rst
```diff
@@ -3,19 +3,60 @@ Usage
...
 
+If you are using custom event source plugins use the following::
 
     ansible-rulebook --inventory inventory.yml --rulebook rules.yml -S sources/
 
-Here `sources` is a directory containing your event source plugins.
+.. note::
+    Here `sources` is a directory containing your event source plugins.
+
+If `eda-server` is running `ansible-rulebook` the `--worker`, `--id`, and `--websocket-address` options should be used::
```
Alex-Izquierdo, line 54: I would remove any mention to EDA-server, as it can be used independently. I would just describe that does the flags.

abaiken: I can do that but I thought these flags were only used when also using `eda-server`. Are there use cases where these specific flags are used without `eda-server`?

Alex-Izquierdo: Well, websocket is a way to expose the event stream data, can be used in multiple ways.  

Anyway, even if those flags are intended for eda-server as this project is independent, I only would explain what they do. What do you think @benthomasson?

Alex-Izquierdo (dismissed):
Minor comment, otherwise LGTM

mkanoor:
@abaiken The cli needs JAVA_HOME env var to be set like mentioned in the code here

https://github.com/ansible/ansible-rulebook/blob/310ec2414d96f60a295645b514c52a23a56cc9eb/ansible_rulebook/cli.py#L125

We should add a note about this, we need Java to be installed and JAVA_HOME set.

Alex-Izquierdo:
> @abaiken The cli needs JAVA_HOME env var to be set like mentioned in the code here
> 
> https://github.com/ansible/ansible-rulebook/blob/310ec2414d96f60a295645b514c52a23a56cc9eb/ansible_rulebook/cli.py#L125
> 
> We should add a note about this, we need Java to be installed and JAVA_HOME set.

That is covered here: https://github.com/ansible/ansible-rulebook/pull/196

docs/usage.rst
abaiken: @Alex-Izquierdo thank you for the explanation! I removed the references to `eda-server` - take another look and make sure what I did makes sense!

Alex-Izquierdo (approved):
LGTM

ddonahue007 (approved):
lgtm

shannon-donahue (approved):
lgtm

PR #193, opened 2022-10-13
Remove github token, add arg for collection devel

ttuffin:


Dockerfile
```diff
@@ -17,7 +17,10 @@ RUN pip install -U pip \
     aiokafka \
     watchdog \
     azure-servicebus \
-    && ansible-galaxy collection install git+https://$GITHUB_TOKEN@github.com/ansible/event-driven-ansible.git
+    && ansible-galaxy collection install git+https://github.com/ansible/event-driven-ansible.git
+
+RUN bash -c "if [ $DEVEL_COLLECTION_LIBRARY -ne 0 ]; then \
```
Alex-Izquierdo, line 22: I would do the opposite, install the published as default and install from git if the boolean is given.

ttuffin: done

Alex-Izquierdo (approved):
LGTM

PR #202, opened 2022-10-14
Start a new plan queue for each event

bzwei:
This is a preparation for running events in parallel whenever possible. Before all actions resulting from events are placed in one queue (plan). It is not a problem when events are processed sequentially.

Now each event will use a dedicated queue to store resulting actions. It helps to collect and concatenate results from these actions while allow events to be processed in parallel.

This work also eliminates the use of data type RuleSetQueuePlan because we can directly convert a RuleSetQueue to EngineRuleSetQueuePlan without using RuleSetQueuePlan as a middle carrier.

See Also: AAP-4436

Replace #158


ansible_rulebook/engine.py
```diff
@@ -350,6 +344,17 @@ async def run_ruleset(
...
+            # Asset added by other ruleset, not related to current event
+            while not ruleset_queue_plan.plan.queue.empty():
+                item = cast(
+                    ActionContext, await ruleset_queue_plan.plan.queue.get()
+                )
+                result = await call_action(*item, event_log=event_log)
+
+            # create a new plan queue for current event so that results
+            # can be concanated
+            ruleset_queue_plan.plan.queue = asyncio.Queue()
```
mkanoor, line 356: @bzwei I am not sure if this is in the correct place. If we have events flowing thru the system wether they match a rule or not we create these Queue objects for every event. So for a simple range source plugin that generates 10 events and you have a condition that says event.i > 8, this code will unconditionally create 10 queues when only 2 might be needed.

ansible_rulebook/engine.py
```diff
@@ -350,6 +344,17 @@ async def run_ruleset(
             continue
         results = []
         try:
+            # Asset added by other ruleset, not related to current event
```
mkanoor, line 347: @bzwei Each ruleset is running in independent sessions so which other ruleset does this relate to? The rulesets communicate with each other by doing a post_event/set_fact with the name of the ruleset. For a running ruleset it doesn't know if the data was generated by a source plugin or another running ruleset from the rulebook.

ansible_rulebook/engine.py
bzwei: The queue is needed by [lang.post(name, data)](https://github.com/ansible/ansible-rulebook/pull/202/files#diff-33ca52e665e828cd48de6eec0723f8b5e74fbcc7ecda7df2d392171ae1987365L354). It may or may not added an item to the queue, but we don't know before hand.

ansible_rulebook/engine.py
bzwei: When ruleset1 sets a fact to ruleset2, it may add a new action to the current plan of the ruleset2 after ruleset2 already finished processing all items in the plan. This code gives it a chance to repeat the processing of the plan before it starts a new plan.

ansible_rulebook/engine.py
mkanoor: @bzwei This lang.post code doesn't use any queues, its targeting a running ruleset by name and sending that data directly to the Drools rule engine.
It hits this post code in the drools_jpy and gets forwarded to Drools which processes all the rules.

https://github.com/mkanoor/drools_jpy/blob/c0936793d06a7231b8a64bd7c4929a009b3fd934/src/drools/ruleset.py#L151


This picture might help clarify things

Screen Shot 2022-10-14 at 7 28 53 PM

All Post Event/Assert Fact/Set Event don't go thru queues they are directly sent to the running ruleset in both Drools and durable rules.

benthomasson (approved):

hsong-rh (approved):

PR #207, opened 2022-10-17
add an endless range source plugin

Alex-Izquierdo:


bzwei:
@Alex-Izquierdo Can you use the existing source plugin https://github.com/ansible/event-driven-ansible/blob/main/plugins/event_source/tick.py?

Alex-Izquierdo:
> @Alex-Izquierdo Can you use the existing source plugin https://github.com/ansible/event-driven-ansible/blob/main/plugins/event_source/tick.py?

@bzwei Not really. First, `tick` not even is adapted for asyncio and it generates an infinite sequence of numbers, so it's not possible to simulate an expected ratio of event matches. With infrange, for instance, you can generate an infinite range of 10 with a condition like `event-i == 5` where do you have assured a ratio 1:10 of triggered events.

mkanoor (approved):

benthomasson (approved):

PR #208, opened 2022-10-17
update minimal python version

Alex-Izquierdo:


cutwater (approved):

ttuffin (approved):
Note Madhu has addressed the 3.8 compatibility issue in AAP-6326, so we may want to reconsider supporting 3.8 further down the track. For now I think we continue as is with support for >=3.9.

cutwater:
Btw, with this change we can start using new syntax sugar for type annotations introduced in python 3.9:

For example: 

`list[str]` instead of `typing.List[str]`

PR #210, opened 2022-10-18
[AAP-4436] Combine the same playbook on multiple hosts

bzwei:
Run the same playbook only once based on multiple events in a
configurable time window through environmental variable
RUN_PLAYBOOK_DELAY (seconds). Combine hosts collected from all
related events.

Resolves AAP-4436: Combine the same action together on multiple hosts
based on multiple events in a certain period of time

This work only places run_playbook actions into the combinable queue. Technically all other actions can be treated the same way, but I feel there is no need since they are all quick actions. The only exception is the run_module action which can be easily handled the same way after the review.

depends on #202 
replaces #159

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import os
 import runpy
+from collections import OrderedDict
 from datetime import datetime
 from pprint import pformat
 from typing import Any, Dict, List, Optional, cast
@@ -20,7 +21,7 @@ else:
     )
 
 import ansible_rulebook.rule_generator as rule_generator
-from ansible_rulebook.builtin import actions as builtin_actions
+from ansible_rulebook.builtin import actions as builtin_actions, run_playbook
 from ansible_rulebook.collection import (
     find_source,
     find_source_filter,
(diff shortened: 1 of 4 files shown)


benthomasson (approved):

ansible_rulebook/engine.py
```diff
@@ -318,74 +206,278 @@ async def run_rulesets(
...
+        self.event_log = event_log
+        self.ruleset_queue_plan = ruleset_queue_plan
+        self.name = ruleset_queue_plan.ruleset.name
+        self.hosts_facts = hosts_facts
+        self.variables = variables
+        self.project_data_file = project_data_file
+
+    async def run_ruleset(self):
+        prime_facts(self.name, self.hosts_facts, self.variables)
+        self.pa_runner_task = asyncio.create_task(
```
hsong-rh, line 230: So far we still cannot sure the action in the current ruleset has `run_playbook` or not, right? is it too earlier to create a playbook runner task? should we wait to do this until we know we have `run_playbook` action, like the calling of `wait_for_playbook`?

hsong-rh (commented):
@bzwei Can you add some real test cases to verify your logic, playbooks combined, hosts updated?

bzwei:
> @bzwei Can you add some real test cases to verify your logic, playbooks combined, hosts updated?

Added a test case to show playbook is run only once with combined hosts.

ansible_rulebook/engine.py
bzwei: We could do a lazy initialization, but since the main use of ansible-rulebook is to run playbooks I feel the code is simple and easy to read by creating the runner task up the front.

hsong-rh (approved):
LGTM

benthomasson (approved):

PR #211, opened 2022-10-18
[AAP-6388] Enhanced the test case for is defined

mkanoor:
The is defined test now checks if the dictionary value is supported.

https://issues.redhat.com/browse/AAP-6388

### DIFF
--- a/tests/examples/19_is_defined.yml
+++ b/tests/examples/19_is_defined.yml
@@ -15,7 +15,8 @@
       condition: event.msg is defined
       action:
         debug:
-
-
-
-
+    - name: r3
+      condition: event.payload is defined
+      action:
+        print_event:
+          pretty: true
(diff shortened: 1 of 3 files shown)


tests/examples/19_is_defined.yml
```diff
@@ -15,7 +15,8 @@
       condition: event.msg is defined
       action:
         debug:
-
-
-
-
+    - name: r3
+      condition: event.payload is defined
```
bzwei, line 19: This rulebook uses `range` as the source. Neither the `range` source nor the `set_fact` action ever generates `payload`. Although you test the rule in `test_examples` by manually creating a `payload`, will it be better to move the rule to another rulebook or add another `set_fact` to create a `payload`?

mkanoor: @bze We have a lot of test cases where the event is injected into the Queue directly without having the source plugins to do that. We don't have a source plugin that doesn't insert anything and allows others to insert the data. The data insertion is controlled by the tests.

Alex-Izquierdo (approved):

bzwei (approved):

PR #218, opened 2022-10-20
Get latest artifacts

bzwei:
Since the rollback of using --rotate-artifacts=1 we need to make sure to always get latest artifacts by sorting the folders by time.

Also refactor to share common code.


bzwei:
@mkanoor I am wondering too why this test on module could fail but not always. So far I have no clue, but clearly it is not caused by the rotate-artifacts option as we thought before. In this PR this option is not used at all.

bzwei:
The test failure was indeed caused by the `rotate_artifacts=1` option which I accidentally reintroduced. This option is now removed and the tests succeed.

https://github.com/ansible/ansible-runner/issues/1160 has been created to report the error.

hsong-rh (approved):
LGTM

mkanoor (approved):

PR #219, opened 2022-10-21
Validate rulebooks against json schema

mkanoor:
- Validates Actions
- Mandatory Action Args
- Rule names


ansible_rulebook/validators.py
```diff
@@ -0,0 +1,31 @@
...
+            with open(schema_file) as f:
+                cls.schema = json.load(f)
+        return cls.schema
+
+    @classmethod
+    def rulebook(cls, instance: List[Dict]) -> None:
+        try:
+            validate(instance=instance, schema=cls._get_schema())
+        except ValidationError:
+            logger.exception("Rulebook failed validation ")
```
Alex-Izquierdo, line 30:
```suggestion
            logger.exception("Rulebook failed validation.")
```

ansible_rulebook/validators.py
```diff
@@ -0,0 +1,31 @@
+import json
+import logging
+import os
+from typing import ClassVar, Dict, List
+
+from jsonschema import validate
+from jsonschema.exceptions import ValidationError
+
+DEFAULT_RULES_SCHEMA = "../schema/ruleset_schema.json"
```
Alex-Izquierdo, line 9: I think the naming is misleading. This variable should be named `DEFAULT_RULEBOOK_SCHEMA` as well as the file should be named `rulebook_schema.json`

ansible_rulebook/validators.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Validate:
+    schema: ClassVar[Dict] = None
+
+    @classmethod
+    def _get_schema(cls):
+        if cls.schema is None:
+            package_dir = os.path.dirname(os.path.realpath(__file__))
+            schema_file = os.path.join(package_dir, DEFAULT_RULES_SCHEMA)
```
Alex-Izquierdo, line 20: Is preferable to use the builtin `resources` library instead. Example: https://github.com/ansible/eda-server/blob/375aaddca1f9928a4105708f05f5cf5ddc226076/src/eda_server/config/__init__.py#L39

tests/test_engine.py
```diff
@@ -419,11 +419,11 @@ async def test_duplicate_rule_names():
 
 @pytest.mark.asyncio
 async def test_empty_rule_names():
-    with pytest.raises(RulenameEmptyException):
+    with pytest.raises(ValidationError):
```
Alex-Izquierdo, line 422: I have detected a bug in this test. The rulebook file doesn't contain a value for "name" field, so yaml parser returns None instead of an empty string. It must have a `""` to force the empty string. Then, I think the test will fail, because the schema would accept an empty string, the schema should have defined the `name` field with min_length attr.

Alex-Izquierdo (changes requested):
This PR's needs more changes in the package metadata because currently the schemas are not included in the distributed package.

tests/test_engine.py
mkanoor: @Alex-Izquierdo I have added a minLength and pattern to the rule name.

ansible_rulebook/validators.py
mkanoor: @Alex-Izquierdo Changed to use resources, I am not sure of there is a well known place to store json schemas in python packages.

Alex-Izquierdo: Good point. I think it should be inside the package. It needs a `manifest` to include the file in the distributed package: https://packaging.python.org/en/latest/guides/using-manifest-in/

schema/ruleset_schema.json
```diff
@@ -115,9 +124,40 @@
...
+                        "copy_files": { "type": "boolean" },
+                        "name": {"type": "string" },
+                        "post_events": { "type": "boolean" },
+                        "set_facts": { "type": "boolean" },
+                        "verbosity": {"type": "integer" },
+                        "var_root": {"type": ["string","object"] },
+                        "json_mode": { "type": "boolean" },
+                        "retry": { "type": "boolean" },
+                        "retries": { "type": "integer" },
+                        "delay": { "type": "integer" },
```
bzwei, line 149: delay type should be  `number`, not limit to integer. Same for run-playbook-action.

mkanoor:
@bzwei @Alex-Izquierdo Please review.

.github/workflows/tests.yml
```diff
@@ -1,10 +1,5 @@
 name: Tests
 
-# The wheel package of jpy is not reliable and causes
```
Alex-Izquierdo, line 3: Why do you remove this?

Alex-Izquierdo (changes requested):
It needs the manifest, otherwise the schema would not be shipped in the pypi package. Also I think the schema should be in ansible_rulebook dir. 
https://packaging.python.org/en/latest/guides/using-manifest-in/

.github/workflows/tests.yml
mkanoor: @Alex-Izquierdo On Python 3.8 I was getting this failure
https://github.com/ansible/ansible-rulebook/actions/runs/3473072376/jobs/5804676382
Which says no not have that the pip_no_binary

Alex-Izquierdo: It was you who determined that the precompiled version of JPY is not reliable. Can we consider it reliable here? or shall we investigate why the compilation fails and fix the workflow accordingly?

Alex-Izquierdo: I did a quick test with python3.8 and it compiles fine, so it must be some issue only with the GH worker.

ttuffin: The `ubuntu-latest` runner still points to ubuntu-20.04. Might be worth updating the workflow to force it to use `ubuntu-22.04`?

Alex-Izquierdo: Good point, that could solves the issue!

mkanoor: @Alex-Izquierdo The JPY team has been making changes to improve the packaging when I started a while back there were issues and I had to ensure that we compile it locally instead of using the wheel directly. Given all the changes we could switch to using the wheel packages if they exist otherwise it would get compiled naturally.

mkanoor:
> It needs the manifest, otherwise the schema would not be shipped in the pypi package. Also I think the schema should be in ansible_rulebook dir.
> https://packaging.python.org/en/latest/guides/using-manifest-in/

@Alex-Izquierdo https://github.com/ansible/ansible-rulebook/blob/0723032336e035b35e326fa620c47eddd722ca4c/MANIFEST.in#L13 already has this added

Alex-Izquierdo:
> > It needs the manifest, otherwise the schema would not be shipped in the pypi package. Also I think the schema should be in ansible_rulebook dir.
> > https://packaging.python.org/en/latest/guides/using-manifest-in/
> 
> @Alex-Izquierdo
> 
> https://github.com/ansible/ansible-rulebook/blob/0723032336e035b35e326fa620c47eddd722ca4c/MANIFEST.in#L13
> 
> already has this added

OMG, I didn't realize it. Great.

Alex-Izquierdo:
I have an error, the location of the schema is wrong:
```
ERROR:ansible_rulebook.cli:Unexpected exception
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/cli.py", line 145, in main
    asyncio.run(app.run(args))
  File "/usr/local/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 650, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/app.py", line 44, in run
    rulesets = load_rulebook(parsed_args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/app.py", line 121, in load_rulebook
    Validate.rulebook(data)
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/validators.py", line 36, in rulebook
    validate(instance=instance, schema=cls._get_schema())
                                       ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/validators.py", line 29, in _get_schema
    data = path.read_text(encoding="utf-8")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/pathlib.py", line 1058, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/pathlib.py", line 1044, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.11/site-packages/ansible_rulebook/../schema/ruleset_schema.json'
```

How to reproduce:
In a clean env
```
export RULES_ENGINE=durable_rules # to skip java installation
pip install -U build
python -m build
pip install dist/ansible_rulebook-0.9.4.tar.gz
cd tests
ansible-rulebook --rulebook ./examples/21_run_playbook.yml --inventory ./playbooks/inventory.yml -S ./sources/
```

mkanoor:
@Alex-Izquierdo Please review, I had to move the schema directory to under the ansible_rulebook directory for the schema to be deployed along with the package.

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -106,18 +115,49 @@
...
+                        "var_root": {"type": ["string","object"] },
+                        "json_mode": { "type": "boolean" },
+                        "retry": { "type": "boolean" },
+                        "retries": { "type": "integer" },
+                        "delay": { "type": "number" }
+                    },
+                    "required": [
+                        "name"
+                    ],
+                    "additionalProperties": true
```
Alex-Izquierdo, line 127: This field should be `false` (run module has it too)
Otherwise an action like:
```
      action:
        run_playbook:
          name: playbooks/hello.yml
          action: "wat"
```
Is not checked and throws the following exception:
```
ERROR:ansible_rulebook.engine:Error calling run_playbook
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/engine.py", line 451, in execute
    result = await run_playbook(**action_args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/ansible_rulebook/builtin.py", line 224, in run_playbook
    temp, playbook_name = await pre_process_runner(
                                ^^^^^^^^^^^^^^^^^^^
TypeError: pre_process_runner() got multiple values for argument 'action'
```

Alex-Izquierdo (changes requested):

ansible_rulebook/schema/ruleset_schema.json
mkanoor: @Alex-Izquierdo In the run module we have an attribute called **module_args** which allows you to set all the module specific args in a single object. And the additional properties is set to false for run_module.
For run_playbook should we add a **playbook_args** and then we can set additional properties to false.

Alex-Izquierdo (approved):
Verified

ansible_rulebook/schema/ruleset_schema.json
Alex-Izquierdo: ok then

mkanoor:
@Alex-Izquierdo Please review I have added playbook_args which allows an end user to keep their playbook specific args separate from ours without causing conflicts.

Alex-Izquierdo (dismissed):
I'm a bit concerned because now we are forcing users to access the variables defined in `playbook_extra_vars` within that key in the playbook. For example:
```
    - name: Print event information
      ansible.builtin.debug:
        msg: '{{ playbook_extra_vars.myvariable }}'
```
Which seems a bit odd. 

The variables defined in the file passed by the `--vars` flag are accessible in the root, so also it's a bit confusing. 

IMO, the content of the `playbook_extra_vars` should be exposed in the root of the ansible variables as happens with the variables defined through "--vars":

```
    - name: Print event information
      ansible.builtin.debug:
        msg: '{{ myvariable }}'
```

Apart from this, the PR works as expected and is verified.

mkanoor:
@Alex-Izquierdo If all variables get merged into root then how would you resolve conflicts if the same variable name appears in playbook_extra_vars and also in --vars?

Alex-Izquierdo:
> @Alex-Izquierdo If all variables get merged  into root then how would you resolve conflicts if the same variable name appears in playbook_extra_vars and also in --vars?

One of them takes precedence. https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#understanding-variable-precedence

Alex-Izquierdo (dismissed):

bzwei (approved):

Alex-Izquierdo (approved):

PR #228, opened 2022-11-10
Combine hosts when running a module

bzwei:
We can already combine hosts for run_playbook action. run_module action is very similar to run_playbook thus can be enhance the same way.

solves AAP-4436: Combine the same action together on multiple hosts based on multiple events in a certain period of time.

See also #210


ansible_rulebook/engine.py
```diff
@@ -373,7 +377,9 @@ async def call_action(
                     action_args["project_data_file"] = self.project_data_file
                     action_args["source_ruleset_name"] = ruleset
                     action_args["source_rule_name"] = rule
-                    return await self.pa_runner.wait_for_playbook(action_args)
+                    return await self.pa_runner.wait_for_playbook(
```
hsong-rh, line 380: can we rename `wait_for_playbook`? now it's not for playbook only

ansible_rulebook/engine.py
```diff
@@ -443,28 +449,35 @@ def stop(self):
 
     async def execute(self):
         for key in list(self.actions.keys()):
+            if key.startswith("run_playbook"):
+                action = "run_playbook"
```
hsong-rh, line 453: Can you store `action` and `run_method` into the `self.actions` via `action_args`, just like what you did on `async_condition`? so here we don't need these comparisons, just pop up and run them.

ansible_rulebook/engine.py
bzwei: I added some comments to explain the whole class is for both playbook and module.

hsong-rh (approved):
LGTM

Alex-Izquierdo (approved):
verified

PR #235, opened 2022-11-21
Do not send ProcessedEvent for unmatched events

bzwei:
There is no need to add empty ProcessedEvent to event_log for events that do not match any rules.

Resolves AAP-6141

Type `ProcessedEvent`, `EmptyEvent`, `MessageNotHandled` are not processed by eda-server at all. In theory we could stop sending these event-log completely. Open for discussion.

We could also added an ENV to preserve the feature of adding a ProcessedEvent for every event like before as discussed in [AAP-6141](https://issues.redhat.com/browse/AAP-6141). Is there any use for such event log?

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -259,6 +259,7 @@ class RuleSetRunner:
                     lang.end_session(self.name)
                 return
             if not data:
+                # TODO: is it really necessary to add such event to event_log?
                 await self.event_log.put(dict(type="EmptyEvent"))
                 continue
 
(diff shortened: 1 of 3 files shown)


mkanoor:
@bzwei We should discuss this with @benthomasson what are the use cases for ProcessedEvent, is that to count the number of events sent into the ansible-rulebook. If we just need to count the number of events sent in we could add a ProcessedEvents at the very end before a shutdown to include the number of events processed.

mkanoor (approved):

PR #237, opened 2022-11-22
Set logger for websocket connect

bzwei:
The websockets library will print connection failure and retry messages at INFO level through the provided logger.

Resolves AAP-6564: ansible-rulebook doesn't handle wrong websocket servers

### DIFF
--- a/ansible_rulebook/websocket.py
+++ b/ansible_rulebook/websocket.py
@@ -85,7 +85,9 @@ async def request_workload(activation_id, websocket_address):
 
 async def send_event_log_to_websocket(event_log, websocket_address):
     logger.info("websocket %s connecting", websocket_address)
-    async for websocket in websockets.connect(websocket_address):
+    async for websocket in websockets.connect(
+        websocket_address, logger=logger
+    ):
         logger.info("websocket %s connected", websocket_address)
         event = None
         try:


mkanoor (approved):

hsong-rh (approved):

Alex-Izquierdo:
should there be messages every time the connection is lost or re-established?

bzwei:
> should there be messages every time the connection is lost or re-established?

Yes. It logs every failed connection and retry interval. What get logged is out of our control.

Alex-Izquierdo (changes requested):
This PR can not resolve the bug IMO
This is what I found on my tests:
- The logger only works when the server is disconnected, but not when the connection is restablished, and only if it was connected at the beginning of execution. 
- There is no any message if the server is not available at the beginning, which is the case described in the bug: 
1.
```
ansible-rulebook --rulebook examples/21_run_playbook.yml  -i playbooks/inventory.yml  -S sources --websocket-address ws://localhost:6667/ws # there is no server listening on localhost:6667
``` 
2.
```
ansible-rulebook --rulebook examples/21_run_playbook.yml  -i playbooks/inventory.yml  -S sources --websocket-address ws://idonotexist:6667/ws # this never will work. 
```

bzwei:
@Alex-Izquierdo you need to turn on verbose flag
```
ansible-rulebook --rulebook examples/21_run_playbook.yml  -i playbooks/inventory.yml  -S sources --websocket-address ws://localhost:6667/ws --verbose
```
This limitation is from the `websockets` library.

Alex-Izquierdo (approved):

PR #239, opened 2022-11-29
Pin flake8 version

Alex-Izquierdo:
There is a new release of flake (6.0.0) that seems is not parsing correctly the comment in the setup.cfg, causing all our tests to fail. This PR pin the version to the most recent working version.


ddonahue007 (approved):
lgtm

abaiken (approved):

bzwei (approved):

PR #243, opened 2022-12-05
New JobTemplateRunner class

bzwei:
A job template runner takes the name of a job template and its organization and identifies its id. It launches the job template and monitors the job events. It accepts a call back method to process job events.

Need follow up work to enable SSL.

Resolves AAP-7535

### DIFF
--- a/ansible_rulebook/job_template_runner.py
+++ b/ansible_rulebook/job_template_runner.py
@@ -1,0 +1,204 @@
+import asyncio
+import json
+import logging
+import os
+from typing import Any, Callable
+from urllib.parse import parse_qsl, urljoin, urlparse
+
+import aiohttp
+import dpath.util
+
+logger = logging.getLogger(__name__)
+
+
+class JobTemplateRunner:
+    JOB_TEMPLATE_SLUG = "/api/v2/job_templates"
+    VALID_POST_CODES = [200, 201, 202]
+    JOB_COMPLETION_STATUSES = ["successful", "failed", "error", "canceled"]
+
+    def __init__(self, host: str, token: str):
+        self.token = token
+        self.host = host
+        self.refresh_delay = int(
+            os.environ.get("EDA_JOB_TEMPLATE_REFRESH_DELAY", 10)
+        )
+
+    async def get_job_template(self, name: str, organization: str) -> dict:
+        job_template = await self._search_list(
+            self.JOB_TEMPLATE_SLUG,
+            {"name": name, "summary_fields.organization.name": organization},
+        )
+        if not job_template:
+            raise Exception(
+                "Job template %s in organization %s does not exist"
+                % (name, organization)
+            )
+
+        return job_template
+
+    async def _search_list(self, href_slug: str, query: dict) -> dict:
+        async with aiohttp.ClientSession(
+            headers=self._auth_headers()
+        ) as session:
+            url_info = urlparse(href_slug)
+            params = dict(parse_qsl(url_info.query))
+            while True:
+                response = await self._get_page(session, url_info.path, params)
+                if response["status"] != 200:
+                    raise Exception(
+                        "Failed to get from %s. Status: %s, Body: %s"
+                        % (
+                            href_slug,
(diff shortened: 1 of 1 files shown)


bzwei:
This initial implementation uses the pulling method to monitor the job events. According to https://github.com/ansible/awx/blob/devel/docs/websockets.md the websocket from awx/controller is meant to be used by UI. Not sure it is suitable for our use. There are not enough examples showing how to use the websocket.

The job event messages are printed to console. The event_handler call back method can be used to add the details of each job event to rulebook's event_log.

ansible_rulebook/job_template_runner.py
```diff
@@ -0,0 +1,204 @@
+import asyncio
+import json
+import logging
+import os
+from typing import Any, Callable
+from urllib.parse import parse_qsl, urljoin, urlparse
+
+import aiohttp
+import dpath.util
```
mkanoor, line 9: @bzwei I noticed this
DeprecationWarning: The dpath.util package is being deprecated. All util
functions have been moved to dpath package top level
I have created a PR for this issue #242

ansible_rulebook/job_template_runner.py
```diff
@@ -0,0 +1,204 @@
...
+                            response["status"],
+                            response.get("body", "empty"),
+                        )
+                    )
+
+                json_body = json.loads(response["body"])
+                for obj in json_body["results"]:
+                    match = True
+                    for key, value in query.items():
+                        if dpath.util.get(obj, key, ".") != value:
```
mkanoor, line 61: @bzwei this could become dpath.get

bzwei:
Added a new commit to cache job template ids.
    
    Not many job templates will be used in a rulebook. We can avoid
    retrieving all job templates in order to match the name and
    organization by caching the known ids. However we still need to
    retrieve the job template by id to verify it is still valid.

mkanoor:
@bzwei Are you going to integrate this into the builtin list separately. We should also investigate if response body contains the artifacts that might be comparable to what we see with the Ansible runner integration.

bzwei:
> @bzwei Are you going to integrate this into the builtin list separately. We should also investigate if response body contains the artifacts that might be comparable to what we see with the Ansible runner integration.

I am working on another PR to add run_job_template to builtin action list. Trying to make the event_log as close to run_playbook as possible.

benthomasson:
@bzwei please rebase

Alex-Izquierdo:
Hi @bzwei the jira ticket in the title and description is wrong.

bzwei:
@Alex-Izquierdo Pushed an update to use named url which avoids multiple API calls to search for job template's ID.

ansible_rulebook/job_template_runner.py
```diff
@@ -0,0 +1,180 @@
...
+                json_body = json.loads(response["body"])
+                job_status = None
+                for event in json_body["results"]:
+                    job_status = dpath.get(
+                        event, "summary_fields.job.status", "."
+                    )
+                    counter = event["counter"]
+                    if counter not in counters:
+                        counters.append(counter)
+                        print(event["stdout"])
```
Alex-Izquierdo, line 94: This should use the logger, don't?

ansible_rulebook/exception.py
```diff
@@ -26,3 +26,13 @@ class RulenameEmptyException(Exception):
...
 
     pass
+
+
+class ControllerApiException(Exception):
+
+    pass
+
+
+class JobTemplateNotExistException(Exception):
```
Alex-Izquierdo, line 36: This exception is not used anywhere

Alex-Izquierdo (commented):
Some minor comments, but my

ansible_rulebook/job_template_runner.py
bzwei: No. This is stdout, suppose to be printed to console.

Alex-Izquierdo (approved):
LGTM

benthomasson (approved):

PR #247, opened 2022-12-09
[AAP-7647] Support for in and contains in condition

mkanoor:
Added support for
* in
* not in
* contains
* not contains

https://issues.redhat.com/browse/AAP-7647

e.g.

   1. event.mylist contains "alert"
   2. event.name in ["alert","warning"]
   3. event.name not in ["error","warning"]
   4. event.mylist not contains "error"

This is a Drools only feature and wont work with durable rules

### DIFF
--- a/ansible_rulebook/condition_parser.py
+++ b/ansible_rulebook/condition_parser.py
@@ -12,13 +12,20 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import logging
+import sys
+
 from pyparsing import (
     Combine,
+    Group,
     Literal,
     OpAssoc,
+    ParseException,
     ParserElement,
     QuotedString,
+    Suppress,
     ZeroOrMore,
+    delimitedList,
     infix_notation,
     one_of,
     pyparsing_common,
@@ -35,6 +42,8 @@ from ansible_rulebook.condition_types import (  # noqa: E402
     OperatorExpression,
     String,
 )
+
+logger = logging.getLogger(__name__)
 
 integer = pyparsing_common.signed_integer.copy().add_parse_action(
     lambda toks: Integer(toks[0])
@@ -61,12 +70,23 @@ string2 = (
     QuotedString('"').copy().add_parse_action(lambda toks: String(toks[0]))
 )
 
+delim_value = Group(delimitedList(integer | ident | string1 | string2))
+list_values = Suppress("[") + delim_value + Suppress("]")
+
+
+def as_list(var):
+    if hasattr(var.__class__, "as_list"):
+        return var.as_list()
+    return var
+
 
 def OperatorExpressionFactory(tokens):
     return_value = None
     while tokens:
         if return_value is None:
-            return_value = OperatorExpression(tokens[0], tokens[1], tokens[2])
+            return_value = OperatorExpression(
+                as_list(tokens[0]), tokens[1], as_list(tokens[2])
+            )
             tokens = tokens[3:]
         else:
             return_value = OperatorExpression(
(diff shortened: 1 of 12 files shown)


Alex-Izquierdo:
closes https://issues.redhat.com/browse/AAP-5169

Alex-Izquierdo:
Just out of curiosity, why two operator to perform the same operation? "contains" and "in" operators are not redundant? In ansible only exists the "in" way. 
From python zen: _There should be one-- and preferably only one --obvious way to do it._

mkanoor:
@Alex-Izquierdo Its based on this https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tests.html#testing-if-a-list-contains-a-value

"Ansible includes a **contains** test which operates similarly, but in reverse of the Jinja2 provided **in** test."

[1,2,3] contains 3
versus
3 in [1,2,3]

Alex-Izquierdo (approved):
I can not verify too much if the engine doesn't support it yet. I executed some regression tests without issues.

Alex-Izquierdo (approved):

mkanoor:
@Alex-Izquierdo Examples added, the new drools_jpy jar has these operators

ansible_rulebook/json_generator.py
```diff
@@ -161,6 +161,42 @@ def visit_condition(parsed_condition: ConditionTypes, variables: Dict):
                     ),
                 }
             }
+        elif parsed_condition.operator == "in":
+            return {
+                "ItemInListExpression": {
```
bzwei, line 166: Seems all the expressions having lhs and rhs have the same value. Can you extract them to share a function?

mkanoor: @bzwei made it into a function call.

benthomasson (approved):

ansible_rulebook/json_generator.py
```diff
@@ -172,6 +188,15 @@ def visit_condition(parsed_condition: ConditionTypes, variables: Dict):
         raise Exception(f"Unhandled token {parsed_condition}")
 
 
+def create_binary_node(name, parsed_condition, variables):
+    return {
+        name: {
+            "lhs": visit_condition(parsed_condition.left, variables),
+            "rhs": (visit_condition(parsed_condition.right, variables)),
```
bzwei, line 195: extra () not needed

ansible_rulebook/json_generator.py
```diff
@@ -161,6 +161,22 @@ def visit_condition(parsed_condition: ConditionTypes, variables: Dict):
                     ),
                 }
             }
+        elif parsed_condition.operator == "in":
+            return create_binary_node(
```
bzwei, line 165: can we refactor all existing applicable operators to use the same create_binary_node method?

ansible_rulebook/condition_parser.py
```diff
@@ -158,4 +178,8 @@ def OperatorExpressionFactory(tokens):
 
 
 def parse_condition(condition_string: str) -> Condition:
-    return condition.parseString(condition_string)[0]
+    try:
+        return condition.parseString(condition_string, parse_all=True)[0]
+    except ParseException as pe:
+        print(pe.explain(depth=0))
```
bzwei, line 184: why print to console not logger?

ansible_rulebook/json_generator.py
mkanoor: @bzwei In a separate PR since its unrelated to these changes.

Alex-Izquierdo (approved):

bzwei (approved):

ansible_rulebook/json_generator.py
mkanoor: @bzwei https://github.com/ansible/ansible-rulebook/pull/267

PR #248, opened 2022-12-12
Remove uuid print in --version flag

Alex-Izquierdo:
I really believe this print is completely useless for the "--version" flag, it generates a new uuid each time is executed. It has no relation with the version and confuses everybody.


ddonahue007 (approved):
Agreed no reason to print out the UUID on this.

PR #251, opened 2022-12-13
Added temporal use cases and handle async responses from Drools

mkanoor:
- Added time constraints in rules schema
- Allow for async communication between Drools and Python
- Added
* Throttle support via
       1. once_within
       2. once_after
* Time windows for conditions
      timeout: nnn seconds
* Not all conditions met in a time window
```
   condition:
             not_all:
                 - condition1
                 - condition2
             timeout: nnn seconds 
```
https://issues.redhat.com/browse/AAP-7354
https://issues.redhat.com/browse/AAP-7360
https://issues.redhat.com/browse/AAP-7284
https://issues.redhat.com/browse/AAP-7706

This PR contains examples which can be executed standalone to demonstrate some of these features.

**time window example**
```
ansible-rulebook --inventory ./tests/playbooks/inventory.yml -S ./tests/sources --rulebook tests/examples/54_time_window.yml
```

**not_all example**
```
ansible-rulebook --inventory ./tests/playbooks/inventory.yml -S ./tests/sources --rulebook tests/examples/55_not_all.yml
```

**once_within**
```
ansible-rulebook --inventory ./tests/playbooks/inventory.yml -S ./tests/sources --rulebook tests/examples/52_once_within.yml
```

**once_after**
```
ansible-rulebook --inventory ./tests/playbooks/inventory.yml -S ./tests/sources --rulebook tests/examples/56_once_after.yml
```

**once_after with multiple timers**
```
ansible-rulebook --inventory ./tests/playbooks/inventory.yml -S ./tests/sources --rulebook tests/examples/57_once_after_multi.yml
```

Since these time based examples take time to run they have been excluded from running during CI testing.
Locally in the dev environment you can set

**export EDA_RUN_TIME_TESTS=yes**
and then run the pytest and it will run the time based tests

Some of the examples take over a minute to run when there are multiple timers involved.

### DIFF
--- a/ansible_rulebook/rule_types.py
+++ b/ansible_rulebook/rule_types.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import asyncio
 from dataclasses import dataclass
-from typing import Any, Dict, List, NamedTuple, Union
+from typing import Any, Dict, List, NamedTuple, Optional, Union
 
 from drools.ruleset import Ruleset as EngineRuleSet
 
@@ -44,6 +44,13 @@ class Action(NamedTuple):
 class Condition(NamedTuple):
     when: str
     value: List[ct.Condition]
+    timeout: Optional[str] = None
+
+
+class Throttle(NamedTuple):
+    group_by_attributes: List[str]
+    once_within: Optional[str] = None
+    once_after: Optional[str] = None
 
 
 class Rule(NamedTuple):
@@ -51,6 +58,7 @@ class Rule(NamedTuple):
     condition: Condition
     action: Action
     enabled: bool
+    throttle: Optional[Throttle] = None
 
 
 class RuleSet(NamedTuple):
--- a/ansible_rulebook/schema/ruleset_schema.json
+++ b/ansible_rulebook/schema/ruleset_schema.json
@@ -51,6 +51,28 @@
                 ]
             }
         },
+        "throttle": {
+            "type": "object",
+	    "oneOf": [
+		    {"required": ["once_within", "group_by_attributes"]},
+		    {"required": ["once_after", "group_by_attributes"]}
+	    ],
+            "properties": {
+                "once_within" : {
+                    "type": "string",
+                    "pattern": "^\\d+\\s(milliseconds?|seconds?|minutes?|hours?|days?)$"
+                },
+                "once_after" : {
+                    "type": "string",
+                    "pattern": "^\\d+\\s(milliseconds?|seconds?|minutes?|hours?|days?)$"
+                },
+                "group_by_attributes": {
+                    "type": "array",
+                    "items": { "type": "string" }
+                }
+            },
+            "additionalProperties": false
+        },
         "rule": {
             "type": "object",
             "properties": {
@@ -59,11 +81,13 @@
                     "minLength": 1,
                     "pattern": "\\S"
                 },
+                "throttle": {"$ref": "#/$defs/throttle"},
                 "condition": {
                     "anyOf": [
                         { "type": "string" },
                         {"$ref": "#/$defs/all-condition"},
-                        {"$ref": "#/$defs/any-condition"}
+                        {"$ref": "#/$defs/any-condition"},
+                        {"$ref": "#/$defs/not-all-condition"}
                     ]
                 },
                 "action": {
@@ -95,8 +119,30 @@
                 "all": {
                     "type": "array",
                     "items": { "type": "string" }
+                },
+                "timeout" : {
+                    "type": "string",
+                    "pattern": "^\\d+\\s(milliseconds?|seconds?|minutes?|hours?|days?)$"
                 }
             },
+            "additionalProperties": false
+        },
+        "not-all-condition": {
+            "type": "object",
+            "properties": {
+                "not_all": {
+                    "type": "array",
+                    "items": { "type": "string" }
+                },
+                "timeout" : {
+                    "type": "string",
+                    "pattern": "^\\d+\\s(milliseconds?|seconds?|minutes?|hours?|days?)$"
+                }
+            },
+            "required": [
+                "timeout",
+                "not_all"
+            ],
             "additionalProperties": false
         },
         "any-condition": {
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -36,6 +36,7 @@ A condition can contain
  * One condition
  * Multiple conditions where all of them have to match
  * Multiple conditions where any one of them has to match
+ * Multiple conditions where not all one of them have to match
 
 Supported Operators
 *******************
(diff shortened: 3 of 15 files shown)


ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -75,6 +75,25 @@
...
+			"pattern": "^\\d+\\s{1}(milliseconds|seconds|minutes|hours|days)$"
+		},
+		"timed_out" : {
+			"type": "string",
+			"minLength": 6,
+			"pattern": "^\\d+\\s{1}(milliseconds|seconds|minutes|hours|days)$"
+		},
+                "unique_attributes": {
+                    "type": "array",
+                    "items": { "type": "string" }
```
Alex-Izquierdo, line 96: It should be in `once_within` isn't?

tests/test_examples.py
```diff
@@ -1026,3 +1026,127 @@ async def test_38_shutdown_action():
...
     event = event_log.get_nowait()
     assert event["type"] == "Shutdown", "2"
+
+
+@pytest.mark.asyncio
+@pytest.mark.skipif(
+    os.environ.get("EDA_RULES_ENGINE", "durable_rules") == "durable_rules",
+    reason="durable rules does not support time based rules",
+)
+async def test_40_coalesce_events_single_host():
```
Alex-Izquierdo, line 1036: How the test verifies that the time window is honored?

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -75,6 +75,25 @@
                         {"$ref": "#/$defs/none-action"},
                         {"$ref": "#/$defs/shutdown-action"}
                     ]
+                },
+		"once_within" : {
+			"type": "string",
+			"minLength": 6,
+			"pattern": "^\\d+\\s{1}(milliseconds|seconds|minutes|hours|days)$"
```
Alex-Izquierdo: Minor comment: no need for the `{1}`

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -75,6 +75,25 @@
                         {"$ref": "#/$defs/none-action"},
                         {"$ref": "#/$defs/shutdown-action"}
                     ]
+                },
+		"once_within" : {
+			"type": "string",
+			"minLength": 6,
```
Alex-Izquierdo: This is not necessary if we have a regex, isn't?

Alex-Izquierdo (changes requested):
- I believe it should include documentation. 
- I don't understand the difference between time_window and timed_out. 
- This PR makes unnecessary the environment variable RUN_PLAYBOOK_MAX_DELAY, isn't? Should it be removed? I suppose the combination with it can generate unexpected behaviors. 
- The title should include also the async responses implementation. 

- Minor comment: in multiple places I see `for i in list(range(10))...`. There is no need to create a list.

ansible_rulebook/rule_types.py
```diff
@@ -50,11 +50,18 @@ class Condition(NamedTuple):
     value: List[ct.Condition]
 
 
+class TemporalConstraints(NamedTuple):
+    name: str
+    value: str
+    unique_attributes: List[str] = None
```
Alex-Izquierdo, line 56: If the default is None, the type hint should be `Optional[List[str]]'

ansible_rulebook/rule_types.py
```diff
@@ -50,11 +50,18 @@ class Condition(NamedTuple):
...
+
+
 class Rule(NamedTuple):
     name: str
     condition: Condition
     action: Action
     enabled: bool
+    time_constraints: TemporalConstraints
```
Alex-Izquierdo, line 64: If this can be None, it should be `Optional[TemporalConstraints]`

Alex-Izquierdo (changes requested):

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -75,6 +75,25 @@
                         {"$ref": "#/$defs/none-action"},
                         {"$ref": "#/$defs/shutdown-action"}
                     ]
+                },
+                "once_within" : {
+                    "type": "string",
+                    "minLength": 6,
+                    "pattern": "^\\d+\\s{1}(milliseconds|seconds|minutes|hours|days)$"
```
Alex-Izquierdo, line 82: I guess that singular values are not supported, right? for example: `1 day`, `1 hour` etc...
This should be emphasized in the documentation

Alex-Izquierdo (commented):
The first time I saw the "timed_out" I thought: what is the difference between the time_window? Now I see that the timed_out is more like a negative trigger. IMHO opinion is confusing, I would like to know more opinions, but I think this could have a more intuitive name. In fact, I believe that "timed_out" is more appropriate for the "time_window" functionality.

ansible_rulebook/schema/ruleset_schema.json
mkanoor: @Alex-Izquierdo unique_attributes is an optional attribute used in conjunction with once_within. Not sure how I would nest it within once_within.

Alex-Izquierdo: I think you need this: https://json-schema.org/understanding-json-schema/reference/conditionals.html

tests/test_examples.py
mkanoor: @Alex-Izquierdo Most of the time validation would have to be tested in integration tests, we are implementing the unit tests in drools_jpy by advancing the clock. Didn't want to repeat those same tests here.
https://github.com/ansible/drools_jpy/blob/1ce4fca15553cd9b77e5226c3d700a1daca1d5aa/tests/test_ruleset.py#L508
Since we have to advance the clock and we are using 2 different languages (python and java) we can't use the time travel python packages like **freeze_gun** to move the time forward. The Java layer allows us to create a pseudo clock on the Java side and we can advance it like here to expedite the tests.
https://github.com/ansible/drools_jpy/blob/1ce4fca15553cd9b77e5226c3d700a1daca1d5aa/tests/test_ruleset.py#L549

Alex-Izquierdo: Ok, It was only curiosity if I missed something. Anyway, I will test it as well in the future e2e tests, so everything is fine. :)

ansible_rulebook/schema/ruleset_schema.json
mkanoor: @Alex-Izquierdo it supports both plural and singular form. It doesn't check if singular digit values are truly unit values. Like 1 day versus 1 days.

mkanoor:
@Alex-Izquierdo I am not content with the timed_out name either we need to discuss it further and come up with a good name maybe expired. I am trying to add documentation in the JIRA stories once we finalize the names we can make changes to the rst files.

mkanoor:
@Alex-Izquierdo I am not sure about the RUN_PLAYBOOK_MAX_DELAY use case @bzwei might have more knowledge about it. In this use cases the time values are configurable per rule instead of globally and they are not specific to any action. The rule engine doesn't know if the action is a run playbook or run module. The unique attributes in the once_within use case allows the author to use specific attributes to make unique sets of events within the time window. For the once_within the event is fired right away and then it would suppress events till the time window expires. It doesn't wait till the once_within time window expires and then trigger the event. The once_within currently only supports 1 condition it doesn't work with multiple conditions we have ongoing discussions with the Drools team to extend this feature as we can come up with more use cases.

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -75,13 +75,33 @@
                         {"$ref": "#/$defs/none-action"},
                         {"$ref": "#/$defs/shutdown-action"}
                     ]
+                },
+                "once_within" : {
+                    "type": "string",
+                    "pattern": "^\\d+\\s(milliseconds|seconds|minutes|hours|days|millisecond|second|minute|hour|day)$"
```
Alex-Izquierdo, line 81: `?` means the preceding character is optional. In this case I propose this more concise version: 
```suggestion
                    "pattern": "^\\d+\\s(milliseconds?|seconds?|minutes?|hours?|days?)$"
```

ansible_rulebook/schema/ruleset_schema.json
Alex-Izquierdo: Ok! thanks for the clarification

mkanoor:
@Alex-Izquierdo Changed unique_attributes to group_by_attributes

bzwei:
The once_within use case fires the event immediately and suppress other similar events. RUN_PLAYBOOK_MAX_DELAY use case on the contrary holds the running of a playbook for a while to see whether there are other requests to run the same playbook on different hosts, if yes, then run the playbook only one time but on all selected hosts.

The once_within use case is mainly for easing event storms. The RUN_PLAYBOOK_MAX_DELAY use case is not necessary for event storms.
> @Alex-Izquierdo I am not sure about the RUN_PLAYBOOK_MAX_DELAY use case @bzwei might have more knowledge about it. In this use cases the time values are configurable per rule instead of globally and they are not specific to any action. The rule engine doesn't know if the action is a run playbook or run module. The unique attributes in the once_within use case allows the author to use specific attributes to make unique sets of events within the time window. For the once_within the event is fired right away and then it would suppress events till the time window expires. It doesn't wait till the once_within time window expires and then trigger the event. The once_within currently only supports 1 condition it doesn't work with multiple conditions we have ongoing discussions with the Drools team to extend this feature as we can come up with more use cases.

mkanoor:
@Alex-Izquierdo @ttuffin Please review this has all the changes that we discussed about 2 weeks back in the dev sync.

docs/conditions.rst
```diff
@@ -357,6 +358,131 @@ Negation Example
...
+| attributes in the event payload which create unique event pairs. In the above example
+| we are using event.meta.hosts and event.code. If we got 2 separate events one that had
+| event.code=warning and another one with event.code=error they would be treated as distinct
+| events and would result in matching multiple events when the action is triggered.
+| Its mandatory to have group_by_attributes specified when using the once_after option.
+| One of the advantages of the **once_after** is that you can collect all the
+| unique events that match the condition and trigger a single action based on multiple
+| matching events, allowing you to combine host information.
+| The timeout units are **milliseconds**, **seconds**, **minutes**, **hours**, **days**.
+| The once_after will only work with single condition and doesn't support multiple conditions
```
ttuffin, line 480:
```suggestion
| The once_after will only work with a single condition and doesn't support multiple conditions.
```

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -51,6 +51,28 @@
                 ]
             }
         },
+        "throttle": {
+            "type": "object",
+	    "anyOf": [
```
ttuffin, line 56: Correct me if I am wrong, but I don't think there are any cases where you would use `once_within` and `once_after` in a single throttle configuration? If this is true then I think the constraint here should be `oneOf`.

ttuffin (commented):
A couple of suggestions and a question. Otherwise, this looks great :+1:

ansible_rulebook/schema/ruleset_schema.json
Alex-Izquierdo: Also it's bad indented. But I see multiple places where we have a bad indentation making it more difficult to read. We can fix everything in a separated PR

Alex-Izquierdo (changes requested):
This is a critical feature and I don´t like the idea to excluding it from CI or other places. We still having a reasonable execution time. 

However, instead of an environment variable which is an odd way and it doesn't scale well, we must use a mark: 
https://docs.pytest.org/en/7.1.x/how-to/mark.html

Then, in that case, excluding it in the CI workflow, but as I said, I don't want to exclude it.

mkanoor:
@Alex-Izquierdo The examples now have a marker called **temporal** so they can be skipped if necessary. Most of the clock manipulation to speed tests happen in drools_jpy like here https://github.com/ansible/drools_jpy/blob/2406ad6b2776d0df3b5965a9f7f99b17babff23b/tests/test_ruleset.py#L548. In our examples we can't move the clock forward so these are real working examples to demonstrate the product features.

mkanoor:
@Alex-Izquierdo @ttuffin Please re-review I have made the necessary changes

ansible_rulebook/schema/ruleset_schema.json
mkanoor: @ttuffin Fixed

Alex-Izquierdo (approved):

ttuffin (approved):

PR #252, opened 2022-12-13
Updates file event source plugin to be async compatible

benthomasson:
See: https://issues.redhat.com/browse/AAP-7703

### DIFF
--- a/tests/sources/file.py
+++ b/tests/sources/file.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import asyncio
 import os
 
 import yaml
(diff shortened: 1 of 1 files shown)


hsong-rh:
@benthomasson I tested and it always hanged at `observer.join()`. Did you have the similar issue?

tests/sources/file.py
```diff
@@ -86,10 +87,15 @@ def on_moved(self, event):
         observer.stop()
 
 
+async def main(queue, args):
+    loop = asyncio.get_event_loop()
+    await loop.run_in_executor(None, sync_main, loop, queue, args)
```
bzwei, line 92: After ctrl-c to break the cli often I get `ERROR:asyncio:Task was destroyed but it is pending!`

I did not see same error if I convert it to the same use as https://github.com/ansible/event-driven-ansible/blob/main/plugins/event_source/file_watch.py#L94

One more question - is https://github.com/ansible/event-driven-ansible/blob/main/plugins/event_source/file_watch.py#L88 needed and why?

bzwei:
> @benthomasson I tested and it always hanged at `observer.join()`. Did you have the similar issue?

I tested the change and it ran fine in my machine.

PR #254, opened 2022-12-15
[AAP-7755] Create NullQueue if no websocket address provided

mkanoor:
https://issues.redhat.com/browse/AAP-7755
If there is no websocket we dont have to create a asyncio Queue since no one is reading from it and it will keep growing

May help with
https://issues.redhat.com/browse/AAP-6289

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -58,7 +58,6 @@ async def run(parsed_args) -> None:
         ) = await request_workload(
             int(parsed_args.id), parsed_args.websocket_address
         )
-        event_log = asyncio.Queue()
     else:
         inventory = {}
         variables = load_vars(parsed_args)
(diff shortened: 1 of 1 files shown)


ansible_rulebook/app.py
```diff
@@ -52,15 +58,15 @@ async def run(parsed_args) -> None:
...
     else:
         inventory = {}
         variables = load_vars(parsed_args)
         rulesets = load_rulebook(parsed_args)
         if parsed_args.inventory:
             inventory = load_inventory(parsed_args.inventory)
         project_data_file = parsed_args.project_tarball
-
-    event_log = asyncio.Queue()
+        event_log = NullQueue()
```
bzwei, line 69: @mkanoor We can technically provide the CLI a websocket address but no activation id or worker mode, then a valid asyncio Queue is still needed.

ansible_rulebook/app.py
```diff
@@ -35,6 +35,12 @@
     send_event_log_to_websocket,
 )
 
+
+class NullQueue:
+    async def put(self, _data):
```
Alex-Izquierdo, line 40: Should it override `put_nowait` method as well?

Alex-Izquierdo (approved):

ansible_rulebook/app.py
mkanoor: @Alex-Izquierdo I don't see any references in the code for event_log.put_nowait.

PR #253, opened 2022-12-14
[AAP-7571] New action: run_job_template

bzwei:
New run_job_template action
    
    run_job_template action arguments can be seen in ruleset_schema.json.
    See a rulebook example of the usage in examples/40_job_template.yml
    
    run_job_template is very much like run_playbook in all aspects.

Depends on #243

PR #255, opened 2022-12-16
[AAP-6818] Configure controller API access

bzwei:
Take controller base url and authentication token from CLI options or through websocket in worker mode.

Resolves AAP-6818

Includes all commits from #243  #253. This concludes the backend support of run_job_template action using a non-SSL connection.
Future work will enable SSL.


ansible_rulebook/cli.py
```diff
@@ -167,6 +179,14 @@ def main(args: List[str] = None) -> int:
         print("Error: inventory is required")
```
Alex-Izquierdo, line 179: This is not related with this PR but I'm looking this line and thinking that now inventory should not be mandatory with run_job_template action.

Alex-Izquierdo (changes requested):
Documentation should indicate that these flags must be defined to use this action

Alex-Izquierdo:
Is there some specification about how must be provided the credentials? I wonder if we should support environment variables as well.

Alex-Izquierdo:
> Documentation should indicate that these flags must be defined to use this action

Leave it to me please. I'm working on a big refactor of the docs and the rebase would be a pain.

ansible_rulebook/cli.py
bzwei: agree

bzwei:
@Alex-Izquierdo Is there any code change needed for this PR?

mkanoor (approved):

PR #259, opened 2022-12-21
Don't create an asyncio.Queue for every event coming in

mkanoor:
We have a single plan queue per Ruleset, where all rules can queue their Actions. The rule engines gives us back the event data when a rule is triggered, we don't need to separate every event into a separate asyncio.Queue.

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -285,10 +285,6 @@ class RuleSetRunner:
                 )
                 result = await self.call_action(*item)
 
-            # create a new plan queue for current event so that results
-            # can be concatenated
-            self.ruleset_queue_plan.plan.queue = asyncio.Queue()
-
             try:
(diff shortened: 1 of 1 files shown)


ansible_rulebook/engine.py
```diff
@@ -285,10 +284,6 @@ async def run_actions(self, data: Dict):
                 )
                 result = await self.call_action(*item)
 
-            # create a new plan queue for current event so that results
-            # can be concatenated
-            self.ruleset_queue_plan.plan.queue = asyncio.Queue()
```
bzwei, line 290: I did some tests. Only these three lines can be removed while other changes are not needed.

mkanoor: @bzwei Updated please review

bzwei (approved):

Alex-Izquierdo (approved):

PR #274, opened 2023-01-10
Choose number of loops for generic source plugin

Alex-Izquierdo:
Choose a fixed number of iterations, negative numbers create an infinite loop.

### DIFF
--- a/tests/sources/generic.py
+++ b/tests/sources/generic.py
@@ -21,16 +21,15 @@ async def main(queue: asyncio.Queue, args: Dict[str, Any]):
     payload = args.get("payload")
     randomize = args.get("randomize", False)
     delay = int(args.get("delay", 0))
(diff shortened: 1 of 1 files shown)


tests/sources/generic.py
```diff
@@ -14,28 +14,27 @@
...
 
-    if loop_mode:
-        payload = cycle(payload)
-
-    for event in payload:
-        await queue.put(event)
-        await asyncio.sleep(delay)
+    iteration = 0
+    while iteration != loop_count:
+        for event in payload:
```
mkanoor, line 34: @Alex-Izquierdo Should we move the random.shuffle in here so its changing for every iteration

Alex-Izquierdo: ok, updated

mkanoor (approved):

hsong-rh (approved):

PR #278, opened 2023-01-12
Remove the use of sphinx-ansible-theme.

ddonahue007:
Remove the use of sphinx-ansible-theme

This is causing not only the docs build to fail, but also from a clean repo I am not able to run `pip install -r requirements_dev.txt` successfully.


Alex-Izquierdo:
But this is was added to be aligned with ansible. Shouldn't we change whatever we need to making it working in readthedocs? 
@ddonahue007 @benthomasson

PR #280, opened 2023-01-13
Increase timeout of e2e tests

Alex-Izquierdo:
Increase the timeout of e2e tests to avoid occasional failures due to GH actions performance. 

In next iterations we will implement other improvements to reduce the time execution of some tests.


ttuffin (approved):
lgtm, although perhaps we could move `DEFAULT_TIMEOUT` to utils.py and override it on a per test file/case basis where necessary.

ddonahue007 (approved):
lgtm

PR #282, opened 2023-01-13
Refactor how to set extra_vars for playbook

bzwei:
1. All user defined variables must be placed under extra_vars argument in the action section.
2. Variables from a static file are not automatically included in extra_vars set to a playbook. They however can be referenced in user defined extra_vars section.
3. Rulebook automatically inserts fact/facts/event/events variables under the common top level variable ansible_eda.

Resolves AAP-8088 and AAP-7760


bzwei:
Will add a new doc to explain in detail how to set extra vars and how to reference them in playbooks.

tests/rules/test_set_facts.yml
```diff
@@ -14,16 +14,17 @@
...
       action:
         run_playbook:
           name: playbooks/validate_args_playbook.yml
-          rulebook_extra_vars:
-             fact.msg: hello world
-             fact.alpha: 1
-             fact.beta.name: R2D2
-             fact.beta.location: Naboo
+          extra_vars:
+            rulebook_extra_vars:
```
mkanoor, line 25: @bzwei can we change "rulebook_extra_vars" to "my_vars" will remove confusion from earlier use of rulebook_extra_vars

docs/actions.rst
```diff
@@ -55,8 +55,8 @@ Run an Ansible playbook.
    * - var_root
      - If the event is a deeply nested dictionary, the var_root can specify the key name whose value should replace the matching event value. The var_root can take a dictionary to account for data when we have multiple matching events.
      - No
-   * - rulebook_extra_vars
-     - In addition to the passed in vars and matching events the rulebook_extra_vars would be passed into the playbook as extra vars.
+   * - extra_vars
+     - Additional vars to be passed into the playbook as extra vars.
```
mkanoor, line 59: @bzwei Should we document the precedence here, also QE would have to validate that. If we pass in --variables here they would override the variables defined in the rulebook in the extra_vars section. Maybe we can add a test for this and use the validate_args_playbook

mkanoor: @bzwei vars can be passed in a file as well as from the environment vars.
https://github.com/ansible/ansible-rulebook/blob/a6c031f15b12c1b75a2574a19f03da6fde3bf783/ansible_rulebook/app.py#L126

mkanoor: @bzwei ansible uses extra-vars with a dash instead of an underscore. https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#defining-variables-at-runtime Shold we align the rulebook to follow the same. Also I am thinking we should changes the ansible-rulebook to use -e or --extra-vars instead of  --variables

ansible_rulebook/builtin.py
```diff
@@ -815,3 +823,15 @@ def _get_events(variables: Dict):
     elif "events" in variables:
         return variables["events"]
     return {}
+
+
+def _collect_extra_vars(variables: Dict, user_extra_vars: Dict, facts: Dict):
```
Alex-Izquierdo, line 828: If I have understood well this function, if an environment variable name matches an user-extra-variable name, now the extra_var it would be replaced by the environment variable, but extra-vars should take precedence. Am I wrong?  If the user wants to override this behavior should use the environment variable in the extra_vars declaration. Such as:

```yaml
    - name: r1
      condition: event.status == "OK"
      action:
        run_playbook:
          name: test.yml
          extra_vars:
            project_name: "{{ project_name }}" # this is the project_name taken from environment variables instead of extra-vars file. 
```

What do you think?  @mkanoor @bzwei

docs/actions.rst
Alex-Izquierdo: totally agree with renaming extra-vars flag.

ansible_rulebook/builtin.py
```diff
@@ -815,3 +823,15 @@ def _get_events(variables: Dict):
...
+    extra_vars = user_extra_vars.copy() if user_extra_vars else {}
+    eda_vars = {}
+    for key in ["fact", "facts", "event", "events"]:
+        if key in variables:
+            eda_vars[key] = variables[key]
+    if facts:
+        eda_vars["facts"] = facts
```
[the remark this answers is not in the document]
mkanoor: @bzwei It looks like an empty dict all the time it starts its journey here
https://github.com/ansible/ansible-rulebook/blob/38f3bab204ad12234705c06d532ca1630c3b45e1/ansible_rulebook/rule_generator.py#L236
We will have to ask @benthomasson if he planned to use it in some way.

ansible_rulebook/builtin.py
mkanoor: @Alex-Izquierdo Both the variables and the env vars all go into the variables pool here and can be used for substitution. 
https://github.com/ansible/ansible-rulebook/blob/38f3bab204ad12234705c06d532ca1630c3b45e1/ansible_rulebook/app.py#L126

Alex-Izquierdo: Yeah, but what if there is a name collision? which takes precedence? We must document it.

bzwei: Sure. Will include in document

ansible_rulebook/builtin.py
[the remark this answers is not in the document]
bzwei: I skipped the facts from argument for now since it is always empty.

ansible_rulebook/builtin.py
bzwei: Document added.

bzwei:
Added extra_vars.rst with explanations and examples.

ttuffin (approved):

Alex-Izquierdo (dismissed):
lgtm

ansible_rulebook/builtin.py
```diff
@@ -849,3 +857,13 @@ def _get_events(variables: Dict):
     elif "events" in variables:
         return variables["events"]
     return {}
+
+
+def _collect_extra_vars(variables: Dict, user_extra_vars: Dict):
+    extra_vars = user_extra_vars.copy() if user_extra_vars else {}
+    eda_vars = {}
+    for key in ["fact", "facts", "event", "events"]:
```
mkanoor, line 865: @bzwei Is this wrong place to be doing this, when we add facts/fact/event/events here 
https://github.com/ansible/ansible-rulebook/blob/68b7fd2f0552d9c22001aec3b8ac9768c484fcb5/ansible_rulebook/engine.py#L353
Should we be just setting the name we want to use instead of delaying it till the end. Because if the user had passed facts/events/event/fact we would have over written them first and then changed their names here.

mkanoor: @bzwei A test would also help, we can pass variables into the ruleset execution and see what the value of the events are going to be.
Also this is going to be breaking change since the events/event have moved so any of the playbooks that some of our beta users are using will also have to change. Unless we plan on supporting both for a while

bzwei:
@mkanoor I now set the prefix key in the beginning. Refactored a test to include the variable substitution.

ansible_rulebook/engine.py
```diff
@@ -349,18 +349,20 @@ async def call_action(
                     else:
                         multi_match = rules_engine_result._m
                 variables_copy = variables.copy()
+                variables_eda = {}
+                variables_copy[KEY_EDA_VARS] = variables_eda
```
mkanoor, line 353: @bzwei Does this impact playbooks written by our beta users, which might be using the root events/facts? If yes how do we communicate that.

bzwei: Yes any playbook references to {{ event.key }} now needs to change to {{ ansible_eda.event.key }}. This is a question to @josephpisciotta or @benthomasson. The good thing is now we have documented how to use such variables in playbooks.

mkanoor:
@bzwei @Alex-Izquierdo It might make sense to capture the rule name and the ruleset name in our extra vars along with event/events/facts/fact.

Alex-Izquierdo:
> @bzwei @Alex-Izquierdo It might make sense to capture the rule name and the ruleset name in our extra vars along with event/events/facts/fact.

What would be the usecase? logging purposes? 

Probably it could be added to the event.meta, don't?

Alex-Izquierdo:
e2e tests fails because the playbook points to {{event}}

mkanoor:
> > @bzwei @Alex-Izquierdo It might make sense to capture the rule name and the ruleset name in our extra vars along with event/events/facts/fact.
> 
> What would be the usecase? logging purposes?
> 
> Probably it could be added to the event.meta, don't?

@Alex-Izquierdo It completes it, since the playbook or job template knows which ruleset/rule triggered an action. It's up to the playbook/job template if they want to use it. We are capturing it on our side in the event log and sending it to eda server. I think it would be  root key in the ansible.eda{ruleset=abc,rule=xyz,event=...,events=...,fact=...,facts=...}

ttuffin (approved):

PR #284, opened 2023-01-14
Added a builtin action : echo

mkanoor:
The debug/print_event have lot of information and don't allow for terse messages. This becomes difficult to comprehend when we have multiple conditions in a rule, which match multiple events. When doing a demo instead of seeing a lot of attributes flow by we can echo terse messages along with a time stamp. The timestamp allows for timer based rules to see if they popped at the expected times.

e.g.
```
- name: r1
      condition: event.i == 1
      action:
        echo:
          message: Hurray it works
```

Displays
```
2023-01-13 21:20:47.028365 : Hurray it works
```

### DIFF
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -134,6 +134,39 @@ async def print_event(
     )
 
 
+async def echo(
+    event_log,
+    inventory: Dict,
+    hosts: List,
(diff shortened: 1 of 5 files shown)


mkanoor:
@Alex-Izquierdo docs changed

bzwei (approved):

PR #290, opened 2023-01-16
Pin drools version

Alex-Izquierdo:
Pip doesn't upgrade dependencies in existing installations if there is no a specified version, which can lead to confusion and generate unexpected results since a critical part of the functionality relies on this package. Pinning it we enforce that we are using always the right version of the package.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
 	janus
 	ansible-runner
 	websockets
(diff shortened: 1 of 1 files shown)


ddonahue007 (approved):
lgtm

hsong-rh (approved):

mkanoor (approved):

PR #293, opened 2023-01-18
Update to use drools_jpy 0.1.7

mkanoor:
Updated setup.cfg to point to the new version
- Support for new temporal AST changes
- The rule engine logs are written to stdout/stderr


Alex-Izquierdo (approved):

mkanoor:
@ttuffin This is need for the logging issue from the rules engine

PR #296, opened 2023-01-18
Refactor how to consume action items from the plan queue

bzwei:
1. Use a seperate asyncio task to pull from the plan queue which is independent on pulling from the source queue

2. No longer generate ProcessedEvent in event_log

3. No longer combine hosts when running playbook like actions. Drop the use of EDA_RUN_PLAYBOOK_MAX_DELAY

Resolves AAP-8425

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -17,7 +17,6 @@ import asyncio
 import logging
 import os
 import runpy
-from collections import OrderedDict
 from datetime import datetime
 from pprint import PrettyPrinter, pformat
 from typing import Any, Dict, List, Optional, cast
@@ -220,7 +219,7 @@ async def run_rulesets(
 
 
 class RuleSetRunner:
-    COMBINEABLE_ACTIONS = ("run_playbook", "run_module", "run_job_template")
+    ANSIBLE_ACTIONS = ("run_playbook", "run_module", "run_job_template")
 
     def __init__(
         self,
@@ -233,7 +232,7 @@ class RuleSetRunner:
     ):
         self.pa_runner = PlaybookActionRunner()
         self.pa_runner_task = None
-        self.action_tasks = []
+        self.action_loop_task = None
         self.event_log = event_log
         self.ruleset_queue_plan = ruleset_queue_plan
         self.name = ruleset_queue_plan.ruleset.name
@@ -241,14 +240,39 @@ class RuleSetRunner:
         self.variables = variables
         self.project_data_file = project_data_file
         self.parsed_args = parsed_args
+        self.stopped = True
 
     async def run_ruleset(self):
+        self.stopped = False
         prime_facts(self.name, self.hosts_facts, self.variables)
         self.pa_runner_task = asyncio.create_task(
             self.pa_runner.start(self.name), name=self.name
         )
+        self.action_loop_task = asyncio.create_task(
+            self._drain_actionplan_queue()
+        )
+        source_loop_task = asyncio.create_task(self._drain_source_queue())
+        await asyncio.wait([source_loop_task])
 
-        logger.info("Waiting for event from %s", self.name)
+    async def _stop(self):
+        # Wait for items in queues to be completed. Mainly for spec tests.
+        await asyncio.sleep(0.01)
+
+        logger.info("Attempt to stop ruleset %s", self.name)
+        self.stopped = True
+        self.pa_runner.stop()
+        # helps to break waiting on the plan queue when it is empty
+        try:
+            self.ruleset_queue_plan.plan.queue.put_nowait(Shutdown())
+        except asyncio.QueueFull:
+            pass
+
+        await asyncio.wait([self.pa_runner_task, self.action_loop_task])
+        await self.event_log.put(dict(type="Shutdown"))
+        lang.end_session(self.name)
+
+    async def _drain_source_queue(self):
+        logger.info("Waiting for events from %s", self.name)
         while True:
             data = await self.ruleset_queue_plan.source_queue.get()
             if self.parsed_args and self.parsed_args.print_events:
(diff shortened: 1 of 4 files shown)


ansible_rulebook/engine.py
```diff
@@ -234,56 +233,62 @@ def __init__(
...
+        self.action_loop_task = asyncio.create_task(
+            self._drain_actionplan_queue()
+        )
+        source_loop_task = asyncio.create_task(self._drain_source_queue())
+        await asyncio.wait([source_loop_task])
 
-        logger.info("Waiting for event from %s", self.name)
+    async def _stop(self):
+        # Wait for items in queues to be completed. Mainly for spec tests.
+        await asyncio.sleep(0.01)
```
mkanoor, line 260: @bzwei Shouldn't the Shutdown event take care of this.

ansible_rulebook/engine.py
```diff
@@ -234,56 +233,62 @@ def __init__(
...
+
+        logger.info("Attempt to stop ruleset %s", self.name)
+        self.stopped = True
+        self.pa_runner.stop()
+        # helps to break waiting on the plan queue when it is empty
+        try:
+            self.ruleset_queue_plan.plan.queue.put_nowait("STOP")
```
mkanoor, line 267: @bzwei Why do we have to put this in the plan queue, which typically contains ActionContext objects. You could have multiple ActionContexts queued but when a shutdown comes the queue would contain
```
    ["ActionContext1","ActionContext2","ActionContext3","STOP"]
```
So this would wait till all ActionContexts are drained before it gets to the STOP

We should be able to stop the action_plan task when the source queue gets a Shutdown

mkanoor: If we ran 2 tasks, the source_queue_task and action_plan_queue_task and run them separately, store each one in the ruleset_tasks. When the source_queue_task gets a "Shutdown" it ends the ruleset and returns when a task is completed we cancel all the ruleset_tasks, which will stop the action_plan_queue_task. It doesn't need special processing

ansible_rulebook/engine.py
```diff
@@ -431,35 +434,33 @@ async def call_action(
...
-        return result
+        self.action_tasks.pop(action_id)
 
 
 class PlaybookActionRunner:
     """Run playbook like actions in background. Queue such actions
-    and run them in an interval and combine hosts if possible.
-    These actions are defined in COMBINEABLE_ACTIONS including
+    and run them in sequential order.
+    These actions are defined in ANSIBLE_ACTIONS including
```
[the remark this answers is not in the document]
bzwei: For now we want all playbooks to run one after another. We will further enhance this class to run them in parallel if there is no host conflict. We don't need user to decide to run them in parallel or sequentially.

ansible_rulebook/engine.py
bzwei: Let me try it out. Thanks

ansible_rulebook/engine.py
```diff
@@ -293,32 +298,29 @@ async def run_actions(self, data: Dict):
...
-        except Exception:
-            logger.exception("Error processing %s", data)
-
-    async def call_action(
+    async def _drain_actionplan_queue(self):
+        logger.info("Waiting for actions on events from %s", self.name)
+        action_id = 0
+        while not self.stopped:
+            queue_item = await self.ruleset_queue_plan.plan.queue.get()
+            if isinstance(queue_item, Shutdown):
```
mkanoor, line 306: @bzwei In the action plan queue who inserts a ShutDown Message? I thought they would always be ActionContext coming from here.
https://github.com/ansible/ansible-rulebook/blob/bfe9f6ebdbc42e39aa793eeb0311af00e2425c09/ansible_rulebook/rule_generator.py#L63

ansible_rulebook/engine.py
```diff
@@ -293,32 +298,29 @@ async def run_actions(self, data: Dict):
...
+            # event. Spec tests assume every thing is processed without lossing
+            # if self.stopped:
+            #    break
+
+            action_item = cast(ActionContext, queue_item)
+            action_id += 1
+            self.action_tasks[action_id] = asyncio.create_task(
+                self._call_action(action_id, *action_item)
+            )
+        logger.info("Stopped waiting for actions on events from %s", self.name)
```
mkanoor, line 319: @bzwei Since this task runs independently it might make sense to put it in a try/except block and catch the asyncio.CancelledError to log a clean shutdown when a task is cancelled.

ansible_rulebook/engine.py
```diff
@@ -328,10 +330,11 @@ async def call_action(
         hosts: List,
         facts: Dict,
         rules_engine_result,
-    ) -> Dict:
+    ) -> None:
 
         logger.info("call_action %s", action)
 
+        result = None
```
mkanoor, line 337: @bzwei Should the action in builtin_actions be called before we create a task, so we don't have to just create a task to say that the action is not supported or not defined.

ansible_rulebook/engine.py
```diff
@@ -234,56 +233,62 @@ def __init__(
...
+
+        logger.info("Attempt to stop ruleset %s", self.name)
+        self.stopped = True
+        self.pa_runner.stop()
+        # helps to break waiting on the plan queue when it is empty
+        try:
+            self.ruleset_queue_plan.plan.queue.put_nowait(Shutdown())
+        except asyncio.QueueFull:
+            pass
+        tasks = list(self.action_tasks.values())
```
mkanoor, line 270: @bzwei Can the tasks be figured out to figure out if there are any pending tasks and the use the asyncio.sleep if the tasks is not empty

ansible_rulebook/engine.py
```diff
@@ -234,56 +233,62 @@ def __init__(
...
+        tasks.append(self.pa_runner_task)
+        tasks.append(self.action_loop_task)
+        await asyncio.wait(tasks)
```
mkanoor, line 273: @bzwei would this cause the playbook task to keep running till it ends.

ansible_rulebook/engine.py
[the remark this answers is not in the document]
mkanoor: If there are multiple rulesets then we are not running the playbooks/job templates one at a time. Since each ruleset is running its own queue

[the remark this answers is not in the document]
Alex-Izquierdo: In a remediation use case, if a user configures a time window to allow time for the playbook to run and recover a service, that time estimation will assume that the playbook is executed as soon as the condition matches. It that action can be queued the execution is unpredictable and the time window would be impossible to estimate. For me, the parallel execution of these actions seems a must.

ansible_rulebook/engine.py
bzwei: No. Shutdown is a special indication of ending the source https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/engine.py#L174. There is no ActionContext for shutdown.

bzwei:
@mkanoor as we discussed off line this PR will wait for queued ansible actions to complete before actual shutdown. We will create another PR that shutdowns immediate upon receiving the command.

ansible_rulebook/engine.py
mkanoor: @bzwei if the Plan.Queue is filled with other ActionItems you would have to drain all of them before you get to the Shutdown. The same happens even in the Playbook serialization Queue. If there are multiple entries in the Playbook queue we would not be able to get to the Shutdown or Stop till all the items have been drained.

mkanoor (approved):
This still has a lot of issues that have to be resolved. But for now its ok and we can handle these items in another PR

1. Should playbooks/modules/job_templates be run sequentially?
2. The 2 Queues(action plan and playbook)  have the Shutdown() message but it might be at the end when the queue is filled with other Actions and might not be accessible without reading all queue items.

bzwei:
At shutdown stage the queue tasks will terminate as soon as there is a message in the queue. The Shutdown() messages are inserted to help if the queues are empty. The code does not drain all messages until reaching the Shutdown().
This code is likely to go away in a following up PR.
> 2. The 2 Queues(action plan and playbook)  have the Shutdown() message but it might be at the end when the queue is filled with other Actions and might not be accessible without reading all queue items.

Alex-Izquierdo (dismissed):

mkanoor (approved):

Alex-Izquierdo (approved):

PR #297, opened 2023-01-18
Support for standalone booleans in conditions

mkanoor:
This helps when we want to print events without knowing the contents of the event.
The literal boolean true allows the condition to pass and the action print_event can be used to print the event

This would usually be used for bulding the rulebook, you would not use it in production.

```
    - name: print event always
      condition: true
      action:
        print_event:
          pretty: true
```

Once you have satisfied with getting the event data you can disable this rule by using
```
    - name: never run
      condition: false
      action:
        print_event:
          pretty: true
```

In programming parlance this is equivalent to doing

```
   if true:
     .....
```
or
```
   if false:
     .....
```
An easier way to comment out a condition

### DIFF
--- a/ansible_rulebook/rules_parser.py
+++ b/ansible_rulebook/rules_parser.py
@@ -124,12 +124,14 @@ def parse_action(action: Dict) -> rt.Action:
 def parse_condition(condition: Any) -> rt.Condition:
     if isinstance(condition, str):
         return rt.Condition("all", [parse_condition_value(condition)])
+    elif isinstance(condition, bool):
+        return rt.Condition("all", [parse_condition_value(str(condition))])
     elif isinstance(condition, dict):
         keys = list(condition.keys())
         if len(condition) == 1 and keys[0] in ["any", "all"]:
             when = keys[0]
             return rt.Condition(
-                when, [parse_condition_value(c) for c in condition[when]]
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo:
I think this could be a good replacement for the --print-all-events flag, isn't? @benthomasson 
https://issues.redhat.com/browse/AAP-8089

I personally prefer a statement in the rulebook instead of a flag.

Alex-Izquierdo (approved):

benthomasson (changes requested):
I am not sure this is the right way to print events.

AssafKatz3:
Hi @Alex-Izquierdo and @benthomasson,
I have another use case for it - we want to process all Kafka events on a topic.  And I believe that we will have more use cases for this option.
Thanks

Alex-Izquierdo:
Hi @AssafKatz3 I suggest to open a new feature request for it. Anyway, this is currently possible. 
You can write a condition that will always match. So, if you events will contain a name field, you can use `event.name is defined`

Even if the schema of your events is dynamic and you can not ensure the existence of a certain field, you can write your own filter to add it.

AssafKatz3:
@Alex-Izquierdo I opened [feature request](https://github.com/ansible/ansible-rulebook/issues/443). I couldn't find any instructions for custom filters.

PR #299, opened 2023-01-19
Use cmdline option --print_events

mkanoor:
--print-events allows for the printing of events as they are being read from the source queue.
If --debug is used --print-events is disabled, since --debug logs the event already.

```
ansible-rulebook --inventory ./tests/playbooks/inventory.yml -S ./tests/sources --rulebook rule1.yml --print-events
{   'address': {'city': 'Bedrock', 'state': 'NJ', 'street': '123 Any Street'},
    'age': 10,
    'name': 'Fred Flintstone'}
Shutdown()
```

### DIFF
--- a/ansible_rulebook/cli.py
+++ b/ansible_rulebook/cli.py
@@ -29,6 +29,7 @@ Options:
     --redis_port=<p>            Redis port
     --debug                     Show debug logging
     --verbose                   Show verbose logging
+    --print-events              Print events after reading from source queue
     --version                   Show the version and exit
     --websocket-address=<w>     Connect the event log to a websocket
     --id=<i>                    Identifier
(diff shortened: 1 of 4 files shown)


ansible_rulebook/cli.py
```diff
@@ -134,6 +135,11 @@ def get_parser() -> argparse.ArgumentParser:
         "--controller-token",
         help="Controller API authentication token",
     )
+    parser.add_argument(
+        "--print_events",
```
ttuffin, line 139: To remain consistent with the other args I think this should be `--print-events`.

ttuffin (changes requested):
If both `--print_events` `--debug` are used together the output is printed twice:
```
2023-01-23 08:36:13,907 - ansible_rulebook.engine - DEBUG - None
{'i': 4}
2023-01-23 08:36:13,907 - ansible_rulebook.engine - DEBUG - Received event : {'i': 4}
```

Is this the intended behavior? For brevity, it might be better to suppress print-events when --debug is used?

mkanoor:
@Alex-Izquierdo @ttuffin Changes made please review

ttuffin (approved):

Alex-Izquierdo (approved):
lgtm

PR #307, opened 2023-01-24
Extend documentation with more content

Alex-Izquierdo:
See https://github.com/ansible/ansible-rulebook/blob/main/docs/readme.md to get instructions about how check this PR. 

Includes:
- The introduction from the collection readme
- "Events and facts" section
- Supported datatypes
- Debug action updated
- Clarification about `not` operator
- Matrix chat link
- advices for development of sources
- Some format improvements

### DIFF
--- a/docs/events_and_facts.rst
+++ b/docs/events_and_facts.rst
@@ -1,0 +1,46 @@
+================
+Events and Facts
+================
+
+
+Differences between Events and Facts
+************************************
+
+Internally, there are no differences on how Events and Facts are processed. Both represent the same data, but there are some differences in the way they are used.
+Events are used to represent the occurrence of something, while Facts are used to represent the state of the world. So, a fact is long live event.
+
+An event is automatically discarded as soon it meets the condition within a rule.
+A fact is not discarded, it is kept in the working memory until it is explicitly removed.
+
+Facts can come as a result of an action, for example, cached facts from the playbook, or can be defined explicitly through the ``set_fact`` action.
+
+You can not combine facts and events in the same condition, because there are different events for the rule engine.
+This will never match:
+
+.. code-block:: yaml
+
+    name: An automatic remediation rule
+    condition: event.outage == true and facts.beta_enabled != true
+    action:
+      run_playbook:
+      name: remediate_outage.yml
+
+Instead, the ``all`` operator must be used:
+
+.. code-block:: yaml
+
+    name: An automatic remediation rule
+    condition:
+      all:
+        - event.outage == true
+        - facts.beta_enabled != true
+    action:
+      run_playbook:
+      name: remediate_outage.yml
+
+
+.. note::
+    To use facts you may use either ``event`` or ``facts`` keys interchangeably.
+
+
+You can combine `set_fact <actions.html#set-fact>`_ and `retract_fact <actions.html#retract-fact>`_ actions to manage the global state during the lifecycle of your rulebook.
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -2,6 +2,11 @@
 Getting started
 ===============
 
+First, see the `installation docs <installation.html>`_ to install ansible-rulebook.
+
+Also you may want to install the `ansible.eda collection <https://github.com/ansible/event-driven-ansible/blob/main/COLLECTION.md>`_
+that contains relevant source plugins.
+
 Let's get started with a simple hello world example to familiarize ourselves with the concepts:
 
 .. code-block:: yaml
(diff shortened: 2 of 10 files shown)


docs/getting_started.rst
```diff
@@ -56,3 +61,133 @@ is a more complete example that accepts alerts from Alertmanager:
...
+Once this condition has been met, our defined action will trigger which
+in this case is to trigger a playbook.
+
+One important thing to take note of ansible-rulebook is that it is not
+like ansible-playbook which runs a playbook and once the playbook has
+been completed it will exit. With ansible-rulebook, it will continue to
+run waiting for events and matching those events, it will only exit upon
+a shutdown action or if there is an issue with the event source itself,
+for example if a website you are watching with the url-check plugin
+stops working.
```
ttuffin, line 119: This is not entirely correct. The CLI will also exit when the condition in a ruleset has been met and the action has been run.

docs/getting_started.rst
```diff
@@ -56,3 +61,133 @@ is a more complete example that accepts alerts from Alertmanager:
 This example sets up a webhook to receive events from alertmanager and then matches events
 where the `fastapi` job alert has a status of `firing`.  This runs a playbook that will
 remediate the issue.
+
+
+Let's build an example rulebook that will trigger an action from a
+webhook. We will be looking for a specific payload from the webhook, and
+if that condition is met from the webhook event, then ansible-rulebook
+with trigger the desired action. Below is our example rulebook:
```
ttuffin, line 69:
```suggestion
will trigger the desired action. Below is our example rulebook:
```

docs/events_and_facts.rst
```diff
@@ -0,0 +1,46 @@
+================
+Events and Facts
+================
+
+
+Differences between Events and Facts
+************************************
+
+Internally there are no differences on how Events and Facts are processed, both represent the same data, but there are some differences in the way they are used.
```
ttuffin, line 9:
```suggestion
Internally, there are no differences on how Events and Facts are processed. Both represent the same data, but there are some differences in the way they are used.
```

docs/events_and_facts.rst
```diff
@@ -0,0 +1,46 @@
...
+    action:
+      run_playbook:
+      name: remediate_outage.yml
+
+
+.. note::
+    To use facts you may use either ``event`` or ``facts`` keys interchangeably.
+
+
+You can combine `set_fact <actions.html#set-fact>`_ and `retract_fact <actions.html#retract-fact>`_ actions to manage the global state during the livecycle of your rulebook.
```
ttuffin, line 46:
```suggestion
You can combine `set_fact <actions.html#set-fact>`_ and `retract_fact <actions.html#retract-fact>`_ actions to manage the global state during the lifecycle of your rulebook.
```

ttuffin (changes requested):

Alex-Izquierdo:
Please, take into account that the new content in "getting started" and "introduction" comes from https://github.com/ansible/event-driven-ansible/blob/main/README.md and it wasn't checked. I want to remove it from the collection and leave it here. If there is some mistake, we should leave it until remove it from the other place to keep both equal.

ttuffin (approved):

docs/sources.rst
```diff
@@ -100,6 +127,17 @@ will limit where the ansible playbook can run. A simple example looks like
...
 
+As the plugin have full access to an unbounded queue that is consumed by ansible-rulebbok
+we carefully recommend to use always the method ``asyncio.Queue.put`` to put events as it's a non-blocking call.
+To give free cpu cycles to the event loop to process the events, we recommend to use ``asyncio.sleep(0)``
+inmediately after the ``put`` method.
+
+.. note::
+    ansible-rulebook is intended to be a long running process and react to events over the time.
+    If the ``main`` function exits then the plugin and ansible-rulebook will be terminated.
+    Usually you may want to implement a loop that keeps running and waits for events endlessly.
```
mkanoor, line 138: @Alex-Izquierdo I think we should add a line here that a Shutdown event is automatically inserted by ansible-rulebook when a plugin ends, which will cause the ansible-rulebook to end. If the ruleset uses multiple sources, each of the source has the ability to end, causing all sources to end. A running source will get canceled when the ansible-rulebook is ending. The rulebook can have actions which trigger a Shutdown, when that shutdown happens the running sources would be canceled. If you need logic to do some cleanup in your sources you should catch the exception  asyncio.CancelledError and do a cleanup. e.g. https://github.com/ansible/event-driven-ansible/blob/1f9ce9641361c96438146f5cb6bae10b60fb4464/plugins/event_source/webhook.py#L48

Alex-Izquierdo: Agree. 
- Empathize that if **only one**  source ends, everything ends. 
- Empathize that the shutdown action in the rulebook ends everything. 
- Mention to catch asyncio.CancelledError if the plugin needs to do cleanup.

Alex-Izquierdo: @mkanoor updated

benthomasson (approved):

ttuffin (approved):

PR #308, opened 2023-01-25
Fix version in setup.cfg for 0.10.0

jamesmarshall24:
ansible-rulebook was recently updated and tagged to 0.10.0 which needs an update in setup.cfg


ttuffin (approved):

Alex-Izquierdo:
I don't know why we created a tag without release, but the version should not be managed manually but as part of the release process.

benthomasson:
Fixed in 0.10.1

PR #310, opened 2023-01-25
Adding e2e test for negation (not) operator

ttuffin:
Closes [AAP-8535](https://issues.redhat.com/browse/AAP-8535).


Alex-Izquierdo (changes requested):
There are some testcases that IMO should be tested:
- combined operators inside the expression like: https://github.com/ansible/ansible-rulebook/blob/main/tests/examples/50_negation.yml#L21 (`and` and `or`)
- multiple conditions (`all` statement)

Now that we are merging multiple cases in the same test/rulebook we have more liberty to write rules and cover more potential combinations without an impact in the execution time. Here you are testing datatypes, which I think is good, but you are only relying on the combination with the operator  `and`

I feel we should have more rules like for example one to check a single use of the operator, as the previously commented: `not (event.i > 10 or event.i < 15)`
Or course, we can not test all the potential combinations, and some of them can be verified implicitly, but I feel just one condition is too few. 

If we test multiple cases, each assert should have a message explaining the error.
E.g `assert got == expected, "multiple conditions failed"`

ttuffin:
@Alex-Izquierdo I have updated the test to cover more operators and multiple conditions.

tests/e2e/test_operators.py
```diff
@@ -371,3 +371,30 @@ def test_in_operator(rulebook, expected_count):
...
+        cwd=utils.BASE_DATA_PATH,
+    )
+
+    output = utils.assert_playbook_output(result)
+
+    printed_events = [
+        line for line in output if "Event matched" in line["stdout"]
+    ]
+
+    assert len(printed_events) == 2, "multiple conditions failed"
```
Alex-Izquierdo, line 400: I have not explained myself well. The message is to clarify which case fails if you make multiple assertions one per case. As it stands now, if the first rule fails, this message would be incorrect. In this case, better to remove the message.

tests/e2e/files/rulebooks/operators/test_not_operator.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Test not (negation) operator
+  hosts: all
+  sources:
+    - generic:
+        event_delay: 1
```
Alex-Izquierdo, line 6: We should not need the delay. If you remove it, it doesn't work. I think it's a bug.

Alex-Izquierdo (approved):

tests/e2e/files/rulebooks/operators/test_not_operator.yml
Alex-Izquierdo: It's not a bug, the source plugin ends before. But we can gain one second replacing the delay by `shutdown_after: 1`

ttuffin: Using `shutdown_after` still allows for the possibility that the plugin ends before the run_playbook action has completed. Out of 100 tests it failed 14 times due to this. Using `event_delay` there were no failures and the difference in delay was only a matter of a few ms, so will leave it as-is, if you agree.

Alex-Izquierdo: ok

Alex-Izquierdo (approved):

benthomasson (approved):

PR #312, opened 2023-01-25
[AAP-6465] Support multiple actions for a rule

mkanoor:
https://issues.redhat.com/browse/AAP-6465

A rule can have either a single action or a list of actions. If multiple actions are specified they are executed one after the other in the order they are specified in the rulebook. e.g.
```
   actions:
      - debug:
      - print_event:
           pretty: true
      - echo:
           message: "Multiple Action Message1"
      - echo:
           message: "Multiple Action Message2"
```

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -317,7 +317,18 @@ class RuleSetRunner:
             #    break
 
             action_item = cast(ActionContext, queue_item)
-            await self._call_action(*action_item)
+            for action in action_item.actions:
+                await self._call_action(
+                    action_item.ruleset,
+                    action_item.rule,
+                    action.action,
+                    action.action_args,
+                    action_item.variables,
+                    action_item.inventory,
+                    action_item.hosts,
+                    action_item.facts,
+                    action_item.c,
+                )
 
         logger.info("Stopped waiting for actions on events from %s", self.name)
 
--- a/ansible_rulebook/json_generator.py
+++ b/ansible_rulebook/json_generator.py
@@ -195,7 +195,7 @@ def visit_rule(parsed_rule: Rule, variables: Dict):
     data = {
         "name": parsed_rule.name,
         "condition": generate_condition(parsed_rule.condition, variables),
-        "action": visit_action(parsed_rule.action, variables),
+        "actions": visit_actions(parsed_rule.actions, variables),
         "enabled": parsed_rule.enabled,
     }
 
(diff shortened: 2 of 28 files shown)


mkanoor:
Needs #313

mkanoor:
Needs #313

ttuffin (dismissed):

Alex-Izquierdo (approved):

Alex-Izquierdo:
Ok, I can understand that we want to skip some tests because take too much time, but it doesn't explain why tests fails.

Alex-Izquierdo:
@benthomasson I think the single "action" syntax should have a deprecation date / deprecation warning.  Otherwise, it should not be removed from the docs.

ttuffin:
> @benthomasson I think the single "action" syntax should have a deprecation date / deprecation warning. Otherwise, it should not be removed from the docs.

@Alex-Izquierdo as per the dev sync minutes we are not deprecating `action`. We will have `action` for one action and `actions` for multiple actions.

Alex-Izquierdo:
> Ok, I can understand that we want to skip some tests because take too much time, but it doesn't explain why tests fails.

The fail is not related with any limit in GH. 
Here are documented the timeouts and limits: 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

I have the same errors in my local environment.

Alex-Izquierdo:
> > Ok, I can understand that we want to skip some tests because take too much time, but it doesn't explain why tests fails.
> 
> I don't believe the fail is related with any limit in GH. Here are documented the timeouts and limits: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits
> 
> I have the same errors in my local environment.

If I remember well, the intention was to keep the backward compatibility, so, I guess we want to remove it at some point. If is not the case, and is a permanent design, the documentation must cover it.

mkanoor:
@Alex-Izquierdo @ttuffin There were more files that needed to be changed. I have fixed it now.

Alex-Izquierdo (approved):

ttuffin (approved):

mkanoor:
@Alex-Izquierdo @ttuffin Had to resolve a conflict after merging the regex pr.

ttuffin (approved):

Alex-Izquierdo (approved):

PR #313, opened 2023-01-25
Skip temporal tests in CI

mkanoor:
In CI the temporal examples takes upwards of 5 minutes because the examples have time delays. This is causing PR's to fail.
Maybe we should look to move the temporal tests to e2e workflow or run a scheduled workflow that can run once an hour to catch any issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,4 +74,4 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Run tests
-        run: pytest -m "not e2e" -vv -n auto
+        run: pytest -m "not e2e and not temporal" -vv -n auto


ttuffin (approved):

Alex-Izquierdo:
I don't like the idea of skipping tests based on the type or the feature. There are potentially many testcases that may need to be skipped due to their execution time. In the case of time windows, we have to test more combinations, including actions that will also consume time (for example, the action run_job_template) so the list of tests to skip can grow a lot and we don't want a pytest command too long. 

Instead, I prefer a more general mark to group all the slow tests regardless of the type or the feature. Call it "slow" or "long" or whatever, but everything under the same mark.

Alex-Izquierdo:
The fails are not related with any limit in GH. 
Here are documented the timeouts and limits: 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

I have the same errors in my local environment. 


I can understand that we want to skip some tests because they take too much time, but it can not be the reason to fail. 
Anyway, I think 5 min is quite reasonable.

PR #314, opened 2023-01-26
[AAP-8646] Shutdown can optionally include message and delay

mkanoor:
https://issues.redhat.com/browse/AAP-8646

The Shutdown message can optionally take the following parameters
- message
- delay

When a source ends it puts out a message
```
Source range initiated shutdown at 2023-01-26 16:35:49.250861
```


ansible_rulebook/builtin.py
```diff
@@ -786,8 +788,20 @@ async def shutdown(
             rule=source_rule_name,
             run_at=str(datetime.utcnow()),
             matching_events=_get_events(variables),
+            delay=delay,
+            message=message,
         )
     )
+
+    logger.info(
```
Alex-Izquierdo, line 796: I see a problem here: the message is printed in the stdout only if --verbose or --debug flags are enabled. This would not resolve one of the problems described in the ticket related with the user experience. 

This message, as well as other actions, should be printed to the stdout by the default.

Alex-Izquierdo (changes requested):
Please, update docs :)

ansible_rulebook/builtin.py
mkanoor: @Alex-Izquierdo I have changed it to a print statement now.
There is more work that needs to be done around Shutdown based on the changes discussed in PR #296 
We will be doing that in a different PR.

ansible_rulebook/builtin.py
```diff
@@ -786,8 +788,17 @@ async def shutdown(
...
             run_at=str(datetime.utcnow()),
             matching_events=_get_events(variables),
+            delay=delay,
+            message=message,
         )
     )
+
+    print(
+        "Ruleset: %s rule: %s has initiated shutdown. "
+        "Delay : %.3f seconds, Message :%s"
```
ttuffin, line 798:
```suggestion
        "Ruleset: %s rule: %s has initiated shutdown. "
        "Delay: %.3f seconds, Message: %s"
```

ansible_rulebook/engine.py
```diff
@@ -156,16 +156,32 @@ async def start_source(
...
-    except BaseException:
+        shutdown_msg = (
+            f"Source {source.source_name} task cancelled, "
+            "initiated shutdown at {str(datetime.now())}"
+        )
+        logger.info("Task cancelled " + shutdown_msg)
+    except BaseException as e:
         logger.exception("Source error")
+        shutdown_msg = (
+            f"Shutting down source : {source.source_name} error : {str(e)}"
```
ttuffin, line 179:
```suggestion
            f"Shutting down source: {source.source_name} error: {str(e)}"
```

ttuffin (approved):
Minor cosmetic suggestions for consistent spacing around colons, otherwise lgtm :+1:

ansible_rulebook/engine.py
Alex-Izquierdo: No need to cast str in f-strings, it's implicit:
```f"Shutting down source: {source.source_name} error: {e}"```

**Review by Alex-Izquierdo — APPROVED** [review]

**Comment by mkanoor** [comment]
@Alex-Izquierdo @ttuffin Please review applied suggested changes.

**Review by ttuffin — APPROVED** [review]

File: ansible_rulebook/engine.py [hunk 4]
```diff
@@ -266,7 +282,14 @@ async def _stop(self):
...
 
         await asyncio.wait([self.pa_runner_task, self.action_loop_task])
-        await self.event_log.put(dict(type="Shutdown"))
+        await self.event_log.put(
+            dict(
+                type="Shutdown",
+                message=obj.message,
+                delay=obj.delay,
+                source_plugin=obj.source_plugin,
+            )
```
benthomasson, line 291: Please change the definition of the Shutdown message in messages.py to include these new fields.    We should use well defined types instead of arbitrary dicts for messages.

benthomasson (changes requested):

ansible_rulebook/engine.py
benthomasson: Convert the Shutdown message back to a dict for sending over the wire.   But this well defined type ensures that we have the correct data in the message to start with.

mkanoor: @benthomasson This is being sent to the event log, which has an extra attribute called **type** which doesn't exist in the Shutdown Message. If we have to put a data type for event_log payloads we would have to do that for all the other messages that we are putting in the event_log. If we need to formalize event_log payload we can do that in a separate PR.

ansible_rulebook/messages.py
```diff
@@ -16,5 +16,6 @@
 
 
 class Shutdown(NamedTuple):
-
-    pass
+    message: str = "Not specified"
```
mkanoor, line 19: @benthomasson This is the formal definition of Shutdown which has all the fields, except type which is needed for event_log.

ansible_rulebook/engine.py
benthomasson: ok that's fine if it will go in another PR.

benthomasson (approved):

PR #315, opened 2023-01-27
AAP-8681 update json schema: rulesets minlength

Alex-Izquierdo:
A rulebook can not contain an empty list of rulesets, update the schema accordingly
closes https://issues.redhat.com/browse/AAP-8681


Alex-Izquierdo:
> The PR makes sense, but out of curiosity, does this cause a different validation error for the user to see? With or without this change the validation should fail due to the other missing keys.

If you try to use an empty rulebook the validation should fail at this point. Maybe the application perform other types of validation later, but even in that case the schema must represent the real contract. This is caused because devs of eda-server supposed that rulesets could be empty.

mkanoor (approved):

PR #151, opened 2022-09-02
Removed + syntax of is defined

mkanoor:
To check if an object.attribute exists use **is defined**

e.g
event.j is defined

This aligns us with the Ansible Playbook syntax.

PR #317, opened 2023-01-27
Removed dead code

mkanoor:
We had dropped the use of + for exists and switched to use is defined in PR https://github.com/ansible/ansible-rulebook/pull/151 These are some of the left over related to that.


ttuffin (approved):

Alex-Izquierdo (approved):
LGTM

bzwei (approved):

PR #325, opened 2023-01-30
[AAP-7562] Added support for searching an array of objects in conditions

mkanoor:
https://issues.redhat.com/browse/AAP-7562
Using the guidelines from selectattr and select in Ansible and Jinja we can write conditions that would check if the select/selectattr yields one or more objects to pass the condition. The result of the selectattr and select is a boolean value. It can only answer if in the array of objects there is an object that satisfies the condition, the data can be further proecssed in the playbook if need be.

This also confirms to the ansible playbook guidelines of tests having either a **is** or **is not**

e.g.
levels is an array of integers
```
    event.levels is select('>', 25)
```

addresses is an array of strings
```
    event.addresses is select('regex', 'Main St')
```

addresses is an array of strings
```
   event.addresses is not select('regex', 'Wall St')
```

event.people is an array of person objects, where the attribute we are checking is person.age and checking if the age is > 30.
```
   event.people is selectattr('person.age', '>', 30)
```

event.people is an array of person objects, where the attribute we are checking is person.name and checking if the name is either Barney or Fred
```
   event.people is selectattr('person.name', 'regex', 'Barney|Fred')
```


Alex-Izquierdo (changes requested):
Documentation please

Alex-Izquierdo (dismissed):

Alex-Izquierdo (dismissed):

benthomasson (dismissed):

ttuffin (approved):

Alex-Izquierdo (approved):

PR #326, opened 2023-01-30
Add SSL support when connect to a controller

bzwei:
Resolves AAP-7570


ttuffin:
@bzwei please update the [Usage page](https://ansible-rulebook.readthedocs.io/en/latest/usage.html) in the docs with this new option

bzwei:
Updated.
> @bzwei please update the [Usage page](https://ansible-rulebook.readthedocs.io/en/latest/usage.html) in the docs with this new option

benthomasson (approved):

ansible_rulebook/cli.py
```diff
@@ -186,6 +193,8 @@ def main(args: List[str] = None) -> int:
         if args.controller_token:
             job_template_runner.host = args.controller_url
             job_template_runner.token = args.controller_token
+            if args.controller_no_ssl_verify:
```
mkanoor, line 196: @bzwei it might make more sense to use a single variable called controller_ssl_verify with 3 values
- yes
- no
- provide a ca certificate bundle

https://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification

We had coded this in catalog like here https://github.com/ansible/pinakes/blob/122fc3184b0c6b6745de220bbd42e632531ec93b/pinakes/main/inventory/task_utils/tower_api.py#L28

Alex-Izquierdo: Not sure if the application should handle custom certs. awx does it?

pinging @jamesmarshall24 for feedback.

Alex-Izquierdo (approved):

ansible_rulebook/job_template_runner.py
```diff
@@ -61,6 +68,15 @@ async def _get_page(
     def _auth_headers(self) -> dict:
         return dict(Authorization=f"Bearer {self.token}")
 
+    @cached_property
+    def _sslcontext(self) -> Union[bool, ssl.SSLContext]:
+        if self.host.startswith("https"):
+            if self.verify_ssl.lower() == "yes":
+                return True
+            elif not self.verify_ssl.lower() == "no":
+                return ssl.create_default_context(cafile=self.verify_ssl)
```
mkanoor, line 77: @bzwei should we check if the cafile is really a file and throw an error if its missing

bzwei: the method `create_default_context` will validate the file and throw an error.

mkanoor (approved):

PR #327, opened 2023-01-31
AAP-8740 Add missing ruleset property to actions in schema

ttuffin:
Closes [AAP-8740](https://issues.redhat.com/browse/AAP-8740).

According to the [docs](https://ansible-rulebook.readthedocs.io/en/latest/actions.html#run-playbook) it should be possible to specify the ruleset for `set_facts` and `post_events` when used under the run_playbook or run_job_template action, but the "ruleset" property is missing from the schema for both of these actions.

Note there are some other discrepancies between the actions params listed in the docs and the JSON schema, but to avoid blowing out the scope for this change those shall be addressed separately in AAP-8744.


benthomasson (approved):

Alex-Izquierdo (approved):

bzwei:
@ttuffin You need to add the same line for run_module.

ttuffin:
> @ttuffin You need to add the same line for run_module.

@bzwei I checked with @mkanoor about `run_module` and it seems that action currently does not produce any artifacts, so set_facts and post_events doesn't work with `run_module`, yet. This is probably something we need to look at separately, if you agree.

PR #332, opened 2023-02-01
Pin black version

Alex-Izquierdo:
Pin black version and fix CI job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,8 @@ jobs:
 
       - name: black
         uses: psf/black@stable
+        with:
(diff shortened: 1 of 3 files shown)


cutwater:
Can you please give some context for this change, like link to the bug in `black`?

Alex-Izquierdo:
> Can you please give some context for this fix, like link to the bug in `black`?
Is not because a bug, but to be protected against uncontrolled upgrades. In later iterations we can update the formatter and code accordingly.

cutwater (approved):

ttuffin (approved):

PR #333, opened 2023-02-01
[AAP-8602] Brought back get_facts for debug action

mkanoor:
https://issues.redhat.com/browse/AAP-8602

The debug action fetches facts from permanent storage, the Drools engine has a temporary storage to store matching events for rules and a permanent storage for facts.


Alex-Izquierdo:
e2e test for debug action must be updated.

Alex-Izquierdo:
Now we have the "facts" key duplicated, one empty, which is a bit confused, don't think you? 

Also could you update this? 
https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/test_actions.py#L60

mkanoor:
@Alex-Izquierdo That facts is coming off, it's a redundant variable that floats thru the system we had discussed this in one of the Dev Sync. It gets inserted into the queue as an empty dict and flows thru the system and shows up in different places. Once that's removed it will go away by itself.
https://github.com/ansible/ansible-rulebook/blob/96c08e305f2287780e22e45c415b52c96d72f7b8/ansible_rulebook/rule_generator.py#L109 I think @bzwei was planning on removing it in the code.

Alex-Izquierdo:
@mkanoor Ok, anyway, we should verify the good facts are printed here: https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/test_actions.py#L60

mkanoor:
@Alex-Izquierdo can we wait on that test till we get the first facts out otherwise we would have 2 facts. The facts is only used for debug actions, it is not used anywhere else.

Alex-Izquierdo (approved):

ansible_rulebook/builtin.py
```diff
@@ -76,6 +76,9 @@ async def debug(event_log, **kwargs):
     print("kwargs:")
     pprint(kwargs)
     print(get_horizontal_rule("="))
+    print("facts:")
+    pprint(lang.get_facts(kwargs["source_ruleset_name"]))
```
bzwei, line 80: I tested this code with the rulebook from the Jira issue, the facts are an empty array []. Is it expected?

mkanoor: @bzwei The only time the rule engine stores anything in permanent storage is when you do set_facts a good example is this one https://github.com/ansible/ansible-rulebook/blob/4a3c9930c153b138df56311147b3a373fac84378/tests/examples/04_set_fact.yml#L14 but you would have to change the following action from print_event to debug to see the data.

bzwei (approved):

PR #335, opened 2023-02-01
Mention event key size limits in sources doc

mkanoor:
Since ansible-rulebook will not process events with over 255 keys, we should document one of the ways to alleviate this issue using ansible.eda.json_filter


Alex-Izquierdo (commented):
I would put this new section before the development section since it's related with the usage of the source plugin.

Alex-Izquierdo (approved):

benthomasson (changes requested):
We should remove the limit on key sizes.  It was for durable rules.

mkanoor:
closed in favor of #341

PR #337, opened 2023-02-02
Remove unused facts in code

bzwei:
A followup of #333

### DIFF
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -52,7 +52,6 @@ async def none(
     inventory: Dict,
     hosts: List,
     variables: Dict,
-    facts: Dict,
     project_data_file: str,
     source_ruleset_name: str,
     source_rule_name: str,
@@ -99,7 +98,6 @@ async def print_event(
     inventory: Dict,
     hosts: List,
     variables: Dict,
-    facts: Dict,
     project_data_file: str,
     source_ruleset_name: str,
     source_rule_name: str,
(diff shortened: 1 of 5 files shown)


mkanoor:
@bzwei I think we should hold on this PR till we get clarification from @benthomasson on this line
https://github.com/ansible/ansible-rulebook/blob/720c67f435fe859b6a9026407a6e2ffad07056d2/ansible_rulebook/engine.py#L387

If that's intentional or a bug. If that line is a bug and it should be
```
if not facts:
    facts = lang.get_facts(ruleset)
```

Then we just have to undo the changes from PR #333 in the debug action and every action will have a facts. This facts is going to be different from events. And we should keep the facts and events separate and pass it to the playbooks and job templates. We had discussed this in the Dev Sync this week about removing facts since they are a copy of events. If the above line is a bug then the facts will be different and we can keep both facts and events.

bzwei:
This line evidently is a bug. But it seems the facts is never used.
> @bzwei I think we should hold on this PR till we get clarification from @benthomasson on this line
> 
> https://github.com/ansible/ansible-rulebook/blob/720c67f435fe859b6a9026407a6e2ffad07056d2/ansible_rulebook/engine.py#L387
> 
> If that's intentional or a bug. If that line is a bug and it should be
> 
> ```
> if not facts:
>     facts = lang.get_facts(ruleset)
> ```
> 
> Then we just have to undo the changes from PR #333 in the debug action and every action will have a facts. This facts is going to be different from events. And we should keep the facts and events separate and pass it to the playbooks and job templates. We had discussed this in the Dev Sync this week about removing facts since they are a copy of events. If the above line is a bug then the facts will be different and we can keep both facts and events. This also resolves the mystery of why facts and events are the same.

Alex-Izquierdo (approved):

benthomasson (approved):
get_facts can be removed.  It is probably a bad idea to export all the facts now.

PR #338, opened 2023-02-02
[AAP-8555] Error parsing not condition

mkanoor:
When using multiple logical statments in a condition with or/and we need ( and ) to resolve ambiguities. Otherwise we should be able to parse simple conditions with a not without braces. The not has a priority over the and and or.

https://issues.redhat.com/browse/AAP-8555
e.g. all of these are valid

```
   condition: not event.i == 5
```

```
   condition: not event.msg == "Barney"
```

```
   condition: not event.j >= 10
```

```
   condition: not event.my_boolean
```

But if you use and/or you need surrounding round braces
```
   condition: not (event.i > 50 or event.i < 10)
```

### DIFF
--- a/ansible_rulebook/condition_parser.py
+++ b/ansible_rulebook/condition_parser.py
@@ -149,7 +149,6 @@ all_terms = (
 condition = infix_notation(
     all_terms,
     [
-        ("not", 1, OpAssoc.RIGHT, lambda toks: NegateExpression(*toks[0])),
         (
             one_of("* /"),
             2,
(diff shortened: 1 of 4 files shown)


ttuffin:
Hi @mkanoor, the note under the [negation example](https://ansible-rulebook.readthedocs.io/en/latest/conditions.html#negation-example) in the docs needs to be updated to reflect this change.

ttuffin (approved):

Alex-Izquierdo (approved):

PR #341, opened 2023-02-03
Removed event key count limits

mkanoor:
There is a limit of 255 keys in an event payload.
The key size limit was used for durable rules. Now that we have switched to drools we don't have the key count limit


Alex-Izquierdo:
Why don't remove https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/util.py#L80

bzwei (approved):

PR #345, opened 2023-02-03
Remove facts and fact from extra_vars

bzwei:
Also add ruleset and rule to extra_vars

Resolves AAP-8787

### DIFF
--- a/docs/events_and_facts.rst
+++ b/docs/events_and_facts.rst
@@ -40,11 +40,11 @@ Instead, the ``all`` operator must be used:
 
 
 .. note::
-    To use facts you may use either ``event`` or ``facts`` keys interchangeably.
+    To use facts you may use either ``events`` or ``facts`` keys interchangeably.
 
 
 You can combine `set_fact <actions.html#set-fact>`_ and `retract_fact <actions.html#retract-fact>`_ actions to manage the global state during the lifecycle of your rulebook.
 
 The text above describes how to use ``events`` or ``facts`` in a rulebook. A single matched ``event`` or multiple matched ``events`` are also
 sent to a playbook through extra_vars under namespace ``ansible_eda`` when a run_playbook or run_job_template action is executed. So in a playbook
-you should reference them as ``ansible_eda.event`` or ``ansible_eda.events``.
+you should reference them as ``ansible_eda.event`` or ``ansible_eda.events``. Facts are not sent to playbooks.
--- a/tests/asts/test_set_facts.yml
+++ b/tests/asts/test_set_facts.yml
@@ -32,10 +32,10 @@
               name: playbooks/validate_args_playbook.yml
               extra_vars:
                 my_vars:
-                  fact.alpha: 1
-                  fact.beta.location: "{{ Naboo }}"
-                  fact.beta.name: R2D2
-                  fact.msg: hello world
+                  event.alpha: 1
+                  event.beta.location: "{{ Naboo }}"
+                  event.beta.name: R2D2
+                  event.msg: hello world
         condition:
           AllCondition:
           - EqualsExpression:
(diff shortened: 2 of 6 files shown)


tests/asts/test_set_facts.yml
```diff
@@ -32,10 +32,10 @@
               name: playbooks/validate_args_playbook.yml
               extra_vars:
                 my_vars:
-                  fact.alpha: 1
-                  fact.beta.location: "{{ Naboo }}"
-                  fact.beta.name: R2D2
-                  fact.msg: hello world
+                  event.alpha: 1
```
mkanoor, line 35: @bzwei Does it make sense to not call this event, maybe something else like group. Because we already have lot of confusion between event/fact events/facts and this example adds to that confusion. These are all user defined variables so we can call it anything we want.

bzwei: @mkanoor We have to construct keys under my_vars exactly the same as what are under ansible_eda in order for this [playbook](https://github.com/ansible/ansible-rulebook/blob/main/tests/playbooks/validate_args_playbook.yml#L2) to validate the results by set_facts.

Alex-Izquierdo (dismissed):

mkanoor:
@bzwei Does this doc need changing https://github.com/ansible/ansible-rulebook/blob/main/docs/events_and_facts.rst

Alex-Izquierdo (dismissed):

bzwei:
@mkanoor Technically no, but I added a few lines to explain the difference. One is used in rulebook, and the other is used in a playbook. Please review. 
> @bzwei Does this doc need changing https://github.com/ansible/ansible-rulebook/blob/main/docs/events_and_facts.rst

docs/events_and_facts.rst
```diff
@@ -40,7 +40,11 @@ Instead, the ``all`` operator must be used:
 
 
 .. note::
-    To use facts you may use either ``event`` or ``facts`` keys interchangeably.
+    To use facts you may use either ``events`` or ``facts`` keys interchangeably.
```
mkanoor, line 43: @bzwei Now that we have removed fact and facts, you can only use events. Shouldn't this line get deleted?

bzwei: This work only changes what are passed to playbook as extra_vars. It does not change how fact and facts are used in a rulebook.

bzwei:
New commit added:

Only add namespace ansible_eda to extra_vars sent to playbooks

Namespace ansible_eda in rulebook is implied. It should never
be explicitly referenced.

Also process var_root before Jinja substitution.

Resolves AAP-8967

docs/extra_vars.rst
```diff
@@ -5,9 +5,9 @@ Extra vars for playbooks or job templates
 Extra vars needed by each playbook or job template should be explicitly
 specified in the extra_vars section under actions in a rulebook. The values of
 extra vars can be literals or references to rulebook variables. In addition to
-the user specified extra vars, the rulebook engine automatically inserts 
-fact/event (single match) or facts/events (multiple matches) under the top
-level key ansible_eda.
+the user specified extra vars, the rulebook engine automatically inserts
```
ttuffin, line 8: Just a note that I am refactoring the docs about variables in #359 and this includes moving the extra_vars info to a new page, so I have incorporated this new text into that refactor. The extra_vars page will no longer exist after the refactor.

ttuffin (dismissed):
Some suggestions, otherwise lgtm.

docs/extra_vars.rst
bzwei: ok. Thanks

ttuffin (approved):

bzwei:
Extract the second commit to new PR #372

mkanoor (approved):

ttuffin (approved):

PR #347, opened 2023-02-03
Intermittent test failures

### LINKED ISSUE #346: Intermittent test failure for time window example
* ansible-rulebook version: 0.10.1
* Python version: 3.9
* Operating System: Mac

### Description

When running pytest one of the examples which uses a time window of 10 seconds gets a QueueEmpty exception. Since the rule engine is not not going to respond till after 10 seconds we might need to control how long to wait

### What I Did
pytest
```
>           event = event_log.get_nowait()

/Users/madhukanoor/devsrc/ansible-rulebook/tests/test_examples.py:1398: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Queue at 0x105b4f4f0 maxsize=0>

    def get_nowait(self):
        """Remove and return an item from the queue.
    
        Return an item if one is immediately available, else raise QueueEmpty.
        """
        if self.empty():
>           raise QueueEmpty
E           asyncio.queues.QueueEmpty

/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/queues.py:189: QueueEmpty
```

mkanoor:
The example test_54_time_window fails intermittently, since the test introduces a delay of 10 seconds, the queue get might get the occasional QueueEmpty exception. Attempting to fix it by adding a retry if we get the QueueEmpty exception.

Fixes #346

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -47,6 +47,18 @@ def load_rulebook(rules_file):
     return ruleset_queues, event_log
 
 
+async def get_queue_item(queue, timeout=0.5, times=1):
+    for _ in range(times):
+        try:
+            item = queue.get_nowait()
+        except asyncio.QueueEmpty:
+            await asyncio.sleep(timeout)
+            continue
+        if item:
+            return item
+    return None
+
+
 def validate_events(event_log, **kwargs):
     shutdown_events = 0
     job_events = 0
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -21,7 +21,7 @@ from ansible_rulebook.exception import VarsKeyMissingException
 from ansible_rulebook.messages import Shutdown
 from ansible_rulebook.util import load_inventory
 
-from .test_engine import load_rulebook, validate_events
+from .test_engine import get_queue_item, load_rulebook, validate_events
 
 
 class SourceTask:
@@ -1395,7 +1395,7 @@ async def test_54_time_window():
             load_inventory("playbooks/inventory.yml"),
         )
 
-        event = event_log.get_nowait()
+        event = await get_queue_item(event_log, 10, 2)
         assert event["type"] == "Action", "1"
         assert event["action"] == "print_event", "1"
         assert event["matching_events"] == {


ttuffin (approved):

tests/test_engine.py
```diff
@@ -47,6 +47,18 @@ def load_rulebook(rules_file):
     return ruleset_queues, event_log
 
 
+async def get_queue_item(queue, timeout=0.5, times=1):
```
Alex-Izquierdo, line 50: Why not making it more time consumption efficient based only on the timeout? something like:
```
async def wait_for_queue_item(queue: asyncio.Queue = None, timeout=0.5):
    start = time.time()
    while True:
        if time.time() - start > timeout:
            return None
        try:
            item = queue.get_nowait()
        except asyncio.QueueEmpty:
            await asyncio.sleep(0.1)
            continue
        if item:
            return item
```

Alex-Izquierdo (approved):

tests/test_engine.py
```diff
@@ -47,6 +47,18 @@ def load_rulebook(rules_file):
...
+    for _ in range(times):
+        try:
+            item = queue.get_nowait()
+        except asyncio.QueueEmpty:
+            await asyncio.sleep(timeout)
+            continue
+        if item:
```
Alex-Izquierdo, line 57: What if the item returned evaluates to False? Don't you wanted it?

mkanoor: @Alex-Izquierdo Its always a dict coming into the event_log queue, it is never an individual scalar value

tests/test_engine.py
mkanoor: @Alex-Izquierdo For the test I am not expecting any activity for 10 seconds because the timer will pop after 10 seconds so doing a sleep for 10 seconds in 0.1 second interval causes unnecessary churn leading to "Ideal Wakeups"

PR #349, opened 2023-02-06
Adding e2e tests for run_playbook action

ttuffin:
Closes [AAP-8660](https://issues.redhat.com/browse/AAP-8660).

### DIFF
--- a/tests/e2e/files/playbooks/run_playbook_test_playbook.yml
+++ b/tests/e2e/files/playbooks/run_playbook_test_playbook.yml
@@ -1,0 +1,51 @@
+---
+- name: A playbook to run simulated post-processing tasks and emulate a first run failure
+  hosts: localhost
+  gather_facts: false
+  vars:
+    tempfile: "/tmp/eda-e2etest-tempfile"
+  tasks:
+    - name: Remediate a host
+      when: remediation_required
+      block:
+        - name: Check if temp file exists
+          ansible.builtin.stat:
+            path: "{{ tempfile }}"
+          register: file_status
+
+        - name: Create temp file
+          ansible.builtin.file:
+            path: "{{ tempfile }}"
+            state: touch
+            mode: "0644"
+          when: not file_status.stat.exists
+
+        - name: Fail if tempfile does not exist
+          ansible.builtin.fail:
+            msg: "Remediation failed on {{ target_host }}"
+          when: not file_status.stat.exists
+
+        - name: Set the result fact
+          ansible.builtin.set_fact:
+            results:
+              remediation:
+                remediation_result: "Success"
+                remediated_host: "{{ target_host }}"
+            cacheable: true
+
+        - name: Echo result
+          ansible.builtin.debug:
+            msg: "Remediation successful"
+
+        - name: Delete the file
+          ansible.builtin.file:
+            path: "{{ tempfile }}"
+            state: absent
+          when: file_status.stat.exists
+
+    - name: Mark complete
+      ansible.builtin.set_fact:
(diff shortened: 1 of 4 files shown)


tests/e2e/files/playbooks/post_processing.yml
```diff
@@ -0,0 +1,51 @@
+---
```
Alex-Izquierdo, line 1: The name of this file is confusing. This playbook is for a specific testcase, is not for general purposes as "print_event". The name should include what test belongs to.

tests/e2e/files/rulebooks/actions/test_run_playbook.yml
```diff
@@ -0,0 +1,56 @@
+---
+- name: Ruleset 1
+  hosts: all
+  sources:
+    - generic:
+        shutdown_after: 6
```
Alex-Izquierdo, line 6: Can you please include this value as envvar injected in the test? This would be a preparation for: https://issues.redhat.com/browse/AAP-9047

Alex-Izquierdo (changes requested):

tests/e2e/test_actions.py
```diff
@@ -94,3 +95,48 @@ def test_actions_sanity():
...
+
+@pytest.mark.e2e
+def test_run_playbook():
+    """
+    Execute a rulebook that contains multiple run_playbook actions
+    to validate all params that are available with the action.
+    """
+
+    rulebook = utils.BASE_DATA_PATH / "rulebooks/actions/test_run_playbook.yml"
+    os.environ["SOURCE_SHUTDOWN_AFTER"] = "6"
```
Alex-Izquierdo, line 108: This set the environment variable for the main process, generating potential side effects. You must create a copy of the environment, update it and inject it only for this specific command. Example: https://github.com/ansible/event-driven-ansible/blob/main/tests/integration/event_source_webhook/test_webhook_source.py#L44

Alex-Izquierdo (changes requested):

tests/e2e/test_actions.py
ttuffin: @Alex-Izquierdo done. Pls re-review

Alex-Izquierdo (dismissed):

ttuffin:
@Alex-Izquierdo please re-review. I have updated the test to use the new env fixture and added a new commit to include `pytest-check` and updated the test.

Alex-Izquierdo (approved):

tests/e2e/files/playbooks/run_playbook_test_playbook.yml
```diff
@@ -0,0 +1,51 @@
...
+  hosts: localhost
+  gather_facts: false
+  vars:
+    tempfile: "/tmp/eda-e2etest-tempfile"
+  tasks:
+    - name: Remediate a host
+      when: remediation_required
+      block:
+        - name: Check if temp file exists
+          ansible.builtin.stat:
```
mkanoor, line 12: @ttuffin We have seen that for some reason the stat call on localhost causes the ansible-playbook to do an auto discovery of python. And the results are different on Mac and Linux.
We were able to fix this in PR #365 by adding the python interpreter in the inventory.yml

mkanoor (approved):

tests/e2e/files/playbooks/run_playbook_test_playbook.yml
ttuffin: Thanks @mkanoor, thats correct, we are checking for strings in stdout so it shouldn't be affected. Will keep it in mind though.

PR #351, opened 2023-02-06
Don't serialize execution of job templates

mkanoor:
Based on the dev sync call last week we only want to serialize the execution of playbooks and modules but not the job templates.


bzwei:
@mkanoor I happened to create #352 to do the same thing. It includes a test for job_templates and have a Jira issue linked.

PR #352, opened 2023-02-06
Remove run_job_template from the queueable list

bzwei:
This allows run_job_template action to run in parallel unconditionally

Also add a test to run lauch and monitor a job template execution

Resolves AAP-8991

### DIFF
--- a/ansible_rulebook/job_template_runner.py
+++ b/ansible_rulebook/job_template_runner.py
@@ -1,3 +1,17 @@
+#  Copyright 2022 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


ansible_rulebook/job_template_runner.py
```diff
@@ -82,12 +96,12 @@ async def run_job_template(
         name: str,
         organization: str,
         job_params: dict,
-        event_handler: Callable[[dict], Any],
+        event_handler: Callable[[dict], Any] = None,
```
Alex-Izquierdo, line 99: If it has None as default value, the type hint is wrong.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #358, opened 2023-02-09
[AAP-9235] Allow for actions to finish in an ordered shutdown

mkanoor:
Gracefully handle shutdowns by allowing queued actions to complete. We found issues on ppc64le env where the processes were running slower and leading to actions not getting completed before the Shutdown was getting issued.

There are 2 kinds of Shutdowns

- graceful
- now

A shutdown of kind=graceful, which is the default allows for actions to complete and then ends the processes.
A shutdown of kind=now, doesn't wait for actions to complete and it ends all tasks

We run ansible playbooks using the ansible_runner which has a cancel_callback to cancel a running playbook, when we get a  shutdown, we cancel running playbooks.

https://issues.redhat.com/browse/AAP-9235

  
 --shutdown-delay command line can be used to specify the maximum time in seconds
      to wait after a shutdown is issued. The process shuts down if all actions are completed before this time. 
      Equivalent env var is EDA_SHUTDOWN_DELAY

### DIFF
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -41,6 +41,7 @@ from .exception import (
     ShutdownException,
 )
 from .job_template_runner import job_template_runner
+from .messages import Shutdown
 from .util import get_horizontal_rule
 
 logger = logging.getLogger(__name__)
@@ -449,6 +450,7 @@ async def call_runner(
 ):
 
     host_limit = ",".join(hosts)
+    shutdown = False
 
     loop = asyncio.get_running_loop()
 
@@ -475,30 +477,42 @@ async def call_runner(
                 val["run_at"] = event_data.get("created")
                 await event_log.put(val)
         except CancelledError:
-            pass
+            logger.info("Ansible Runner Queue task cancelled")
+
+    def cancel_callback():
+        return shutdown
 
     tasks = []
 
     tasks.append(asyncio.create_task(read_queue()))
 
     with concurrent.futures.ThreadPoolExecutor(max_workers=1) as task_pool:
-        await loop.run_in_executor(
-            task_pool,
-            partial(
-                ansible_runner.run,
-                private_data_dir=private_data_dir,
-                limit=host_limit,
-                verbosity=verbosity,
-                event_handler=event_callback,
-                json_mode=json_mode,
-                **runner_args,
-            ),
-        )
+        try:
+            await loop.run_in_executor(
+                task_pool,
+                partial(
+                    ansible_runner.run,
+                    private_data_dir=private_data_dir,
+                    limit=host_limit,
+                    verbosity=verbosity,
+                    event_handler=event_callback,
+                    cancel_callback=cancel_callback,
+                    json_mode=json_mode,
+                    **runner_args,
+                ),
+            )
+        except CancelledError:
+            logger.debug("Ansible Runner Thread Pool executor task cancelled")
+            shutdown = True
+            raise
+        finally:
+            # Cancel the queue reading task
+            for task in tasks:
+                if not task.done():
+                    logger.debug("Cancel Queue reading task")
+                    task.cancel()
 
-    # Cancel the queue reading task
-    for task in tasks:
-        task.cancel()
-    await asyncio.gather(*tasks)
+            await asyncio.gather(*tasks)
 
 
 async def untar_project(output_dir, project_data_file):
@@ -764,8 +778,9 @@ async def shutdown(
     source_ruleset_name: str,
     source_rule_name: str,
     ruleset: str,
-    delay: float = 0.0,
+    delay: float = 60.0,
     message: str = "Default shutdown message",
+    kind: str = "graceful",
 ):
     await event_log.put(
         dict(
@@ -778,16 +793,16 @@ async def shutdown(
             matching_events=_get_events(variables),
             delay=delay,
             message=message,
+            kind=kind,
         )
     )
 
     print(
-        "Ruleset: %s rule: %s has initiated shutdown. "
+        "Ruleset: %s rule: %s has initiated shutdown of type: %s. "
         "Delay: %.3f seconds, Message: %s"
-        % (source_ruleset_name, source_rule_name, delay, message)
+        % (source_ruleset_name, source_rule_name, kind, delay, message)
     )
-    await asyncio.sleep(delay)
-    raise ShutdownException()
+    raise ShutdownException(Shutdown(message=message, delay=delay, kind=kind))
 
 
 actions: Dict[str, Callable] = dict(
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -53,6 +53,14 @@ from ansible_rulebook.util import collect_ansible_facts, substitute_variables
 logger = logging.getLogger(__name__)
 
 
+all_source_queues = []
+
+
+def broadcast(shutdown: Shutdown):
+    for queue in all_source_queues:
+        queue.put_nowait(shutdown)
+
+
 class FilteredQueue:
     def __init__(self, filters, queue: asyncio.Queue):
         self.filters = filters
@@ -76,8 +84,10 @@ async def start_source(
     source_dirs: List[str],
     variables: Dict[str, Any],
     queue: asyncio.Queue,
+    shutdown_delay: float = 60.0,
 ) -> None:
 
+    all_source_queues.append(queue)
     try:
         logger.info("load source")
         if (
@@ -160,13 +170,13 @@ async def start_source(
     except KeyboardInterrupt:
         shutdown_msg = (
             f"Source {source.source_name} keyboard interrupt, "
-            f"initiated shutdown at {str(datetime.now())}"
+            + f"initiated shutdown at {str(datetime.now())}"
         )
         pass
     except asyncio.CancelledError:
         shutdown_msg = (
             f"Source {source.source_name} task cancelled, "
-            "initiated shutdown at {str(datetime.now())}"
+            + f"initiated shutdown at {str(datetime.now())}"
         )
         logger.info("Task cancelled " + shutdown_msg)
     except BaseException as e:
@@ -177,7 +187,13 @@ async def start_source(
         logger.error(shutdown_msg)
         raise
     finally:
-        await queue.put(Shutdown(shutdown_msg, 0.0, source.source_name))
+        broadcast(
+            Shutdown(
+                message=shutdown_msg,
+                source_plugin=source.source_name,
+                delay=shutdown_delay,
+            )
+        )
 
 
 async def run_rulesets(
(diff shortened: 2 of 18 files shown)


tests/e2e/test_actions.py
```diff
@@ -2,14 +2,15 @@
...
+import os
 import subprocess
 
 import pytest
 
 from . import utils
 
 LOGGER = logging.getLogger(__name__)
-DEFAULT_TIMEOUT = 15
+DEFAULT_TIMEOUT = int(os.environ.get("EDA_E2E_DEFAULT_TIMEOUT", "15"))
```
Alex-Izquierdo, line 13: As a preparation for https://issues.redhat.com/browse/AAP-9026 which is a previous step of https://issues.redhat.com/browse/AAP-9047 I would implement a config.py and a settings instance for the e2e tests and manage the default timeout in only place.

Alex-Izquierdo: Maybe the refactor of E2E tests should come in a different PR.

Alex-Izquierdo:
Should the graceful shutdown have a default timeout overwritable in the rulebook?

bzwei:
@mkanoor What changes had been made for engine.py other than extracting ruleset runner to a separate file? Because of the content move I cannot tell the actual change.

Alex-Izquierdo (changes requested):
Please, Can you remove the EDA_E2E_DEFAULT_TIMEOUT implementation? It doesn't belong to the scope of this PR and more changes are necessary for https://issues.redhat.com/browse/AAP-9047

mkanoor:
@Alex-Izquierdo @bzwei Please review added test cases for terminating long running playbooks, which just sleep and are terminated during a shutdown.

ansible_rulebook/exception.py
```diff
@@ -14,8 +14,11 @@
 
 
 class ShutdownException(Exception):
+    def __init__(self, obj):
```
Alex-Izquierdo, line 17: Seems that `obj` can be only a `shutdown` object. IMO using a more self describing variable name and the respective type hint is convenient.

Alex-Izquierdo (commented):
Please, correct me if I'm wrong. If a source ends, a default shutdown action is scheduled. How can be defined the default timeout in that case? I feel the rulebook should contain a top level key and an environment variable for it. 

I only see SIGINT handled. What about SIGTERM?

ansible_rulebook/builtin.py
```diff
@@ -483,30 +485,44 @@ async def read_queue():
                 val["run_at"] = event_data.get("created")
                 await event_log.put(val)
         except CancelledError:
-            pass
+            logger.info("Ansible Runner Queue task cancelled")
+
+    def cancel_callback():
+        if shutdown:
+            return True
+        return False
```
Alex-Izquierdo, line 493:
```suggestion
    def cancel_callback():
        return shutdown
```

Alex-Izquierdo:
@mkanoor The Jira issue in the PR's description is wrong. We need a new ticket to keep track this specific implementation

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -388,7 +388,11 @@
                     "type": ["object","null"],
                     "properties": {
                         "delay": {"type": "number" },
-                        "message": {"type": "string" }
+                        "message": {"type": "string" },
+                        "kind" : {
+                           "type": "string",
+                           "pattern": "^(now|graceful)$"
+                        }
```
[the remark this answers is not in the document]
bzwei: Not sure we really need the kind property since now is equivalent to delay = 0.

ansible_rulebook/engine.py
```diff
@@ -52,6 +52,14 @@
 logger = logging.getLogger(__name__)
 
 
+all_source_queues = []
+
+
+async def broadcast_shutdown(obj):
+    for queue in all_source_queues:
+        queue.put_nowait(obj)
```
bzwei, line 60: either remove `async` from the method def or use `queue.put()`.  better to use `queue.put()`

bzwei:
@mkanoor how long does it takes for pytest to run all tests? If it is too long can we adjust time parameters to achieve the same goal?

mkanoor:
> @mkanoor how long does it takes for pytest to run all tests? If it is too long can we adjust time parameters to achieve the same goal?

@bzwei its not just a matter of timeout, we need some predictability that all actions in a graceful shutdown have gotten a chance to end. The current implementation is based on listening on the Source queue and reacting to a Shutdown message with no information about active tasks. There is no ability to cancel long running playbooks. We do have information about the Action Queue and the items pending in it once the shutdown gets issued. We have to let all the actions finish within the specified time window.

mkanoor:
> Please, correct me if I'm wrong. If a source ends, a default shutdown action is scheduled. How can be defined the default timeout in that case? I feel the rulebook should contain a top level key and an environment variable for it.
> 
> I only see SIGINT handled. What about SIGTERM?

There is no signal handling yet, once we get the Shutdowns sorted out we can easily add a signal handler and issue different kinds of Shutdown messages based on different signals. As far as the source containing a default shutdown delay or a ruleset containing a default shutdown delay we can discuss that. Currently the source ending by is considered to be a graceful shutdown, only actions or signals (in future) can initiate a now shutdown.

ansible_rulebook/schema/ruleset_schema.json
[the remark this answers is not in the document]
mkanoor: @bzwei The **kind** makes it explicit which kind of a shutdown it is. If we assume delay of 0.0 means an immediate shutdown it won't be clear to the user. If the kind="now" then the delay gets ignored.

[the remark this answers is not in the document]
bzwei: @mkanoor It is not assumed. Isn't delay=0 actually causing an immediate shutdown?

Alex-Izquierdo:
@mkanoor Ok, signals can be discussed later, but the capability to define the default timeout I think must come with this change. Don't think you?

mkanoor:
> @mkanoor Ok, signals can be discussed later, but the capability to define the default timeout I think must come with this change. Don't think you?

@Alex-Izquierdo there are 2 distinct things 
1. Shutdown Event (emanates from sources)
2. Shutdown Action (emanates from a rulebook)

The Shutdown Action puts a Shutdown Event, which gets broadcast to all running rulesets. The shutdown action has all the configuration parameters available to it, so we don't need one for specific rulebook.

The Shutdown event that gets generated when a Source ends, is where the default is being currently used. We would need to provide a value for the source instead of the rulebook. In the Generic Plugin we had added a shutdown after, should we formalize a parameter for the source plugin which can be passed into the source and then used when it shuts down.
It might need more discussion.

Alex-Izquierdo:
Although the shutdown event is bound to a source plugin it affects to the whole process, so  IMO, the timeout must be defined at that scope and applied to all sources. You don't want to apply different timeouts depending on the source, the time that you want to wait to kill the process depends on your rulebook (if it contains long-running playbooks, for example) or your environment. 

Let's discuss it then.

Alex-Izquierdo (approved):

mkanoor:
> Although the shutdown event is bound to a source plugin it affects to the whole process, so IMO, the timeout must be defined at that scope and applied to all sources. You don't want to apply different timeouts depending on the source, the time that you want to wait to kill the process depends on your rulebook (if it contains long-running playbooks, for example) or your environment.
> 
> Let's discuss it then.

Another point here is if there are multiple rulesets in a rulebook and they specify different values for shutdown delay, which value should we use, they can't be different values but either the max or min of all the ruleset values. Preferably the max of all ruleset values.

Alex-Izquierdo:
Good point! The default timeout should be defined for the global process: an environment variable, cmd flag or static config (if it would be implemented). It can not be at the rulebook scope because we can have multiple rulesets.

tests/e2e/test_actions.py
```diff
@@ -118,6 +120,8 @@ def test_run_playbook(update_environment):
         env=env,
     )
 
+    print(result.stdout)
```
Alex-Izquierdo, line 123: I think this is not needed.

ttuffin: It was just while we troubleshoot the problem yesterday. It can be removed along with `debug=True`

mkanoor: @Alex-Izquierdo I have removed it for now, but I am thinking we might need a way to enable debug output when we see errors on Travis, that doesn't happen in local Development environment, especially with race conditions.

Alex-Izquierdo: Travis? do you mean github action? If a fails occurs, pytest print the stacktrace and the values compared. If you find something that needs to be printed, I think it's better using the logger.

mkanoor:
@Alex-Izquierdo the shutdown delay is a high water mark, how long you are willing to wait. If there are no pending actions the shutdown happens right away it doesn't wait for the 60 seconds default. Also providing a shutdown delay will prompt us to second guess how long your tests are going to run and fine tune it to 5 seconds, 10 seconds etc when the system does it for you for free, max TTL for all actions after a shutdown is issued is 60 seconds.  As soon as the actions wrap up the shutdown happens you don't have to second guess.

Alex-Izquierdo:
Don't forget to update the changelog

mkanoor:

ansible_rulebook/messages.py
```diff
@@ -12,10 +12,14 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import NamedTuple
+from dataclasses import dataclass
 
+DEFAULT_SHUTDOWN_DELAY = 60.0
```
Alex-Izquierdo, line 17: I really think this value might be taken from an environment variable. 
```
import os
DEFAULT_SHUTDOWN_DELAY = float(os.environ.get("EDA_DEFAULT_SHUTDOWN_DELAY", 60.0))
```

Alex-Izquierdo (approved):
I would like the default delay would be configurable with an envvar (and docs updated accordingly), otherwise, LGTM

docs/usage.rst
```diff
@@ -35,6 +35,7 @@ The `ansible-rulebook` CLI supports the following options:
     --controller-ssl-verify CONTROLLER_SSL_VERIFY
                             How to verify SSL when connecting to the controller, yes|no|<path to a CA bundle>, default to yes for https connection
     --print-events        Print events to stdout, redundant and disabled with -vv
+    --shutdown-delay      Maximum number of seconds to wait after a graceful shutdown is issued, default is 60. Can also be set via an env var called EDA_SHUTDOWN_DELAY. The process will shutdown if all actions complete before this time period
```
mkanoor, line 38: @Alex-Izquierdo The docs have been updated to use EDA_SHUTDOWN_DELAY

ttuffin (approved):

PR #359, opened 2023-02-10
Refactor documentation on variables

ttuffin:
- Create new page to cover all variables usage
- Extra vars docs combined with new variables page

Pending merge of #345 which has some minor conflicts.


docs/variables.rst
```diff
@@ -0,0 +1,131 @@
...
+        $ MY_NAME="Bob" ansible-rulebook -i inventory.yml --rulebook hello_there.yml --vars vars.yml --env-vars MY_NAME
+
+Which would return the following output:
+
+    .. code-block:: console
+
+        2023-02-09 13:37:73.311337 : Hi, I'm Bob
+
+.. note::
+    - Quotation marks around Jinja substituted variables are **mandatory** when used as the first element in a value.
```
Alex-Izquierdo, line 75: I don't understand the last part of this sentence. Quotation marks around Jinja expressions are mandatory *always*. Conditions do not use jinja. Maybe it's a matter to remark that difference.

docs/variables.rst
```diff
@@ -0,0 +1,131 @@
...
+    - Note the variables that use Jinja substitution in the source config and action message, in contrast to
+      the variable in the condition which instead uses the "**vars.**" prefix. Jinja substitution will not work
```
Alex-Izquierdo, line 77: There is a note in conditions to remark that Jinja is not supported for conditions. Maybe is good place to link it

https://github.com/ansible/ansible-rulebook/blame/main/docs/conditions.rst#L28

docs/variables.rst
bzwei: An example will help.

ttuffin: @Alex-Izquierdo it is best practice, but it is not mandatory. This is true in Ansible and ansible-rulebook. For example, the following will work:
`message: The action is {{ msg }}`

But this will fail:
`message: {{ msg }} is the action`

It is mentioned in the Ansible [docs](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#gotchas) but since its a common 'gotcha' that folks run into I think we should mention it in our docs as well. Ansible has a clear error message when you hit do it wrong, but ansible-rulebook does not.

I can add an example.

ttuffin: Or we change the note to recommend always enclosing values that contain Jinja substituted variables in quotation marks.

Alex-Izquierdo: IMO better and failsafe saying use always quotation marks

Alex-Izquierdo (approved):
Minor comments, otherwise LGTM

bzwei (approved):

PR #362, opened 2023-02-10
[AAP-9218] Honor limit set in run_job_template's job_args.limit field

bzwei:
hosts defined in this field should overwrite rulebook's hosts field but yield to hosts in event's meta field

Resolves AAP-9218

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -21,6 +21,7 @@ from datetime import datetime
 from pprint import PrettyPrinter, pformat
 from typing import Any, Dict, List, Optional, cast
 
+import dpath
 from drools import ruleset as lang
 from drools.dispatch import establish_async_channel, handle_async_messages
 from drools.exceptions import (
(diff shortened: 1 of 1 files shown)


ansible_rulebook/engine.py
```diff
@@ -362,6 +363,17 @@ async def _call_action(
         result = None
         if action in builtin_actions:
             try:
+                if action == "run_job_template":
+                    limit = dpath.get(
```
Alex-Izquierdo, line 367: action_args and job_args are typed as dicts, why do you use dpath?

bzwei: We don't have to use it, but dpath simplifies the code when I need to dig into nested dict. In this case I don't have to test whether action_args['job_args'] exists. dpath works on dicts https://pypi.org/project/dpath/

ttuffin (approved):

AdamMack2007:
@bzwei 

When running this on version 0.11.0 I am still getting an issue.

Here is the rulebook:
```[amack@ansible-eda ansible-event-driven-automation]$ vi rulebook.yml 

- name: Example
  hosts: all
  sources:
    - ansible.eda.webhook:
        host: 0.0.0.0
        port: 5000
  rules:
    - name: Get device info if BGP neighbor down
      condition: event.payload.message == "BGP neighbor down"
      throttle:
         once_within: 5 minutes
         group_by_attributes:
           - event.payload.meta.hosts
      action:
        run_job_template:
          name: Parsers
          job_args:
            limit: "{{ event.payload.meta.hosts }}"
          organization: MackNet

```
When condition is triggered the rulebook output looks correct for the 'limit' field:

```2023-03-27 17:40:09,010 - ansible_rulebook.rule_set_runner - INFO - action args: {'name': 'Parsers', 'job_args': {'limit': 'leaf3'}, 'organization': 'MackNet'}```

But in Controller the limit shows as 
`{{ event.payload.meta.hosts }}`

PR #365, opened 2023-02-13
Fix broken test, was missing localhost in events

mkanoor:
The ansible-playbook was skipping the playbook since none of the hosts matched with this warning.
```
PLAY [Print Hello World and set as fact] ***************************************
skipping: no hosts matched
```
Fixed the event payload to include meta hosts.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -572,15 +572,15 @@ async def test_20_is_not_defined():
 
 
 PLAYBOOK_RULES = [
-    "examples/21_run_playbook.yml",
-    "examples/33_run_playbook_retry.yml",
-    "examples/34_run_playbook_retries.yml",
+    ("examples/21_run_playbook.yml", 12),
+    ("examples/33_run_playbook_retry.yml", 33),
+    ("examples/34_run_playbook_retries.yml", 33),
 ]
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("rule", PLAYBOOK_RULES)
-async def test_21_run_playbook(rule):
+@pytest.mark.parametrize("rule, ansible_events", PLAYBOOK_RULES)
+async def test_21_run_playbook(rule, ansible_events):
     ruleset_queues, event_log = load_rulebook(rule)
 
     queue = ruleset_queues[0][1]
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (approved):

tests/test_examples.py
```diff
@@ -572,36 +572,45 @@ async def test_20_is_not_defined():
...
 @pytest.mark.asyncio
-@pytest.mark.parametrize("rule", PLAYBOOK_RULES)
-async def test_21_run_playbook(rule):
+@pytest.mark.parametrize("rule, ansible_events", PLAYBOOK_RULES)
+async def test_21_run_playbook(rule, ansible_events):
     ruleset_queues, event_log = load_rulebook(rule)
 
     queue = ruleset_queues[0][1]
-    queue.put_nowait(dict(i=1))
+    queue.put_nowait(dict(i=1, meta=dict(hosts="localhost")))
```
bzwei, line 587: no need to add meta. load_inventory of playbooks/inventory.yml is sufficient

tests/test_examples.py
```diff
@@ -572,36 +572,45 @@ async def test_20_is_not_defined():
...
         event_log,
         ruleset_queues,
         dict(),
-        dict(),
+        load_inventory("playbooks/inventory.yml"),
     )
 
     event = event_log.get_nowait()
     assert event["type"] == "Job", "0"
-    for i in range(4):
```
bzwei, line 599: change `range(4)` to `range(ansible_events)` is simpler and enough.

tests/test_examples.py
```diff
@@ -572,36 +572,45 @@ async def test_20_is_not_defined():
...
+            ansible_event_count += 1
+        else:
+            break
+
+    assert ansible_event_count == ansible_events
     assert event["type"] == "Action", "1"
     assert event["action"] == "run_playbook", "2"
-    assert event["matching_events"] == {"m": {"i": 1}}
+    assert event["matching_events"] == {
+        "m": {"i": 1, "meta": {"hosts": "localhost"}}
```
bzwei, line 612: no need meta as explained above

Alex-Izquierdo: Setting the interpreter is a bad idea because it might not match the path of the environment. Instead I think it's better setting the auto_silent mode: 
https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html

Alex-Izquierdo: Oh, I see you do: `ansible_python_interpreter: "{{ansible_playbook_python}}"` If it's works I guess is enough

bzwei (approved):

Alex-Izquierdo (approved):

PR #373, opened 2023-02-15
[AAP-9299] Fixes stack trace when a playbook is not found

mkanoor:
When a playbook cannot be found, we throw an index error
```
  File "/Users/madhukanoor/devsrc/ansible-rulebook/ansible_rulebook/engine.py", line 522, in execute
    await action_method(**action_args)
  File "/Users/madhukanoor/devsrc/ansible-rulebook/ansible_rulebook/builtin.py", line 328, in run_playbook
    await post_process_runner(
  File "/Users/madhukanoor/devsrc/ansible-rulebook/ansible_rulebook/builtin.py", line 635, in post_process_runner
    fact_folder = _get_latest_artifact(
  File "/Users/madhukanoor/devsrc/ansible-rulebook/ansible_rulebook/builtin.py", line 845, in _get_latest_artifact
    return files[0]
IndexError: list index out of range
```
Throw an exception if the runner results file can't be found. Also prevent looking for the facts file when the runner has failed to execute the playbook.

https://issues.redhat.com/browse/AAP-9299

- Add a log info message to always report the **rc** and **status**
- If rc is not 0, read the stderr or stdout in the private_data_dir and report it as an error
- If the playbook has syntax error it would be visible in the error log. Previously this was getting suppressed
- if the rc is not 0 then we should not post_event to set_fact

### DIFF
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -35,7 +35,11 @@ from drools import ruleset as lang
 
 from .collection import find_playbook, has_playbook, split_collection_name
 from .conf import settings
-from .exception import ShutdownException
+from .exception import (
+    PlaybookNotFoundException,
+    PlaybookStatusNotFoundException,
+    ShutdownException,
+)
 from .job_template_runner import job_template_runner
 from .util import get_horizontal_rule
 
@@ -573,9 +577,9 @@ async def pre_process_runner(
                 os.path.join(project_dir, name),
             )
         else:
-            logger.error(
-                "Could not find a playbook for %s from %s", name, os.getcwd()
-            )
+            msg = f"Could not find a playbook for {name} from {os.getcwd()}"
+            logger.error(msg)
+            raise PlaybookNotFoundException(msg)
 
     return (private_data_dir, playbook_name)
 
@@ -597,6 +601,12 @@ async def post_process_runner(
 
     rc = int(_get_latest_artifact(private_data_dir, "rc"))
     status = _get_latest_artifact(private_data_dir, "status")
+    logger.info("Playbook rc: %d, status: %s", rc, status)
+    if rc != 0:
+        error_message = _get_latest_artifact(private_data_dir, "stderr")
+        if not error_message:
+            error_message = _get_latest_artifact(private_data_dir, "stdout")
+        logger.error(error_message)
 
     result = dict(
         type="Action",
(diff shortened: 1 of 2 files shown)


ttuffin (dismissed):

Alex-Izquierdo:
Hi @mkanoor I'm not sure if it's a good idea bind the logic to the return code of ansible. Let's say that you have a playbook that performs many actions, some of them populate facts used by rulebook, for example: service_stopped. If a later task fail, we would be ignoring a real event.

mkanoor:
@Alex-Izquierdo There are all kinds of issues related to handling failures. Today when a playbook fails for whatever reason
1. Invalid Syntax
2. Missing Playbook
3. Some other error
We assume we have to retry that again, the situation doesn't change we try it whatever the retry limit specifies.
The second aspect is when a failure happens, is there a guarantee that we can rely on that data generated by the playbook and blindly send it as a fact into the rule engine. If we know the playbook was successfully executed it makes sense to send up the facts/events into the rule engine. If it fails is there a point in sending partial, missing, empty fact into the rule engine.

Alex-Izquierdo:
We not detecting earlier if the playbook is not found. Just its printed the message from ansible-runner:
`ERROR! the playbook: ./playbooksmissing.yml could not be found`

Alex-Izquierdo (approved):

mkanoor:
@bzwei There are no error codes coming from Ansible Runner that says that something is missing, its a non zero return code. But in the stdout there is an error message, it's not in stderrr (which is empty). But for our retry logic we treat non zero return code as a trigger to retry. Then we retry multiple times and fail with the same error multiple times and then fail. If there were clear error codes from Ansible Runner for when a file is missing it would help but I haven't found one.

bzwei:
@mkanoor We have checking files logic https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/builtin.py#L559. If no playbook is found, we should not just log the error but stop proceeding to ansible_runner.

mkanoor:
> @mkanoor We have checking files logic https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/builtin.py#L559. If no playbook is found, we should not just log the error but stop proceeding to ansible_runner.

@bzwei I have added an exception to fix that code path if the playbook is missing. We still need the other changes when there are other errors in the playbook. for e.g Syntax Errors in Playbook will not be caught in this path.

bzwei (approved):

mkanoor:

Alex-Izquierdo (approved):

ttuffin (approved):

PR #374, opened 2023-02-15
Fix intermittently failing run_playbook test

ttuffin:
Depends on #358.

With the introduction of ordered shutdown in PR 358, an intermittent race condition arose with the run_playbook test due to Ruleset 2 bailing before it received the set_fact from ruleset 1. PR 358 adds ordered shutdown, but if there are no pending items in the action queue then the ruleset will end immediately. In this test, once the time specified under `shutdown_after` in ruleset 1 has expired then a shutdown command will be issued to all rulesets, and since ruleset 2 does not have any pending items it will end immediately, sometimes before the `set_fact` from ruleset 1 has had a chance to run. Increasing `shutdown_after` is an option, but this increases the test time unnecessarily. 

This change moves the condition that performs the set_fact so that it occurs at the beginning of the run, so we can be sure that ruleset 2 will not end before the set_fact and therefore avoid the race condition. We may run into a race condition with other arch's, such as ppc, in which case we would need to increase `SOURCE_SHUTDOWN_AFTER` for that particular arch.

### DIFF
--- a/tests/e2e/files/rulebooks/actions/test_run_playbook.yml
+++ b/tests/e2e/files/rulebooks/actions/test_run_playbook.yml
@@ -3,14 +3,30 @@
   hosts: all
   sources:
     - generic:
-        shutdown_after: "{{ SOURCE_SHUTDOWN_AFTER }}"
+        shutdown_after: "{{ DEFAULT_SHUTDOWN_AFTER }}"
         payload:
+          - alert:
+              message: "upgrade complete"
+            meta:
+              hostname: "nala"
           - alert:
               message: "upgrade failure"
             meta: 
               hostname: "simba"
 
   rules:
+    - name: Run post_processing playbook and set_facts to Ruleset 2
+      condition: event.alert.message == "upgrade complete"
+      action:
+        run_playbook:
+          name: ./playbooks/run_playbook_test_playbook.yml
+          set_facts: true
+          ruleset: "Ruleset 2"
+          extra_vars:
+            processed_host: "{{ event.meta.hostname }}"
+            remediation_required: false
+          json_mode: true
+
     - name: Run post_processing playbook and post_events to same ruleset
       condition: event.alert.message == "upgrade failure"
       action:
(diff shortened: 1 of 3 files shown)


tests/e2e/files/rulebooks/actions/test_run_playbook.yml
```diff
@@ -40,16 +55,24 @@
           json_mode: true
           verbosity: 4
 
+    - name: Validate set_facts from same ruleset
+      condition: >
+            event.post_processing_state == "Complete" and
```
Alex-Izquierdo, line 60: Can you use the same level of indentation (2 spaces)?

tests/e2e/files/rulebooks/actions/test_run_playbook.yml
```diff
@@ -40,16 +55,24 @@
...
   sources:
     - generic:
         startup_delay: 30  # never start, only triggered by Ruleset 1
   rules:
-    - name: Validate the set_facts operation
+    - name: Validate the set_facts from ruleset 1
       condition: >
             event.post_processing_state == "Complete" and
-            event.post_processing_host == "simba"
+            event.post_processing_host == "nala"
```
Alex-Izquierdo, line 75: Can you use the same level of indentation (2 spaces)?

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

ttuffin:
Tests fail intermittently due to the post_events and set_facts not always finishing before the source plugin ends. This will be resolved once 358 is merged.

ttuffin:
@mkanoor @Alex-Izquierdo pls review now that 358 has been merged.

mkanoor (approved):

ttuffin:
Failing due to missing `dynaconf` package from test requirements file. Will be fixed once #381 is merged.

mkanoor (approved):

Alex-Izquierdo (approved):

PR #375, opened 2023-02-16
Adding e2e tests for variables

ttuffin:
Closes [AAP-8369](https://issues.redhat.com/browse/AAP-8369).

### DIFF
--- a/tests/e2e/test_variables.py
+++ b/tests/e2e/test_variables.py
@@ -1,0 +1,79 @@
+"""
+Module with tests for the use of variables
+"""
+import logging
+import subprocess
+
+import pytest
+from pytest_check import check
+
+from . import utils
+from .settings import SETTINGS
+
+LOGGER = logging.getLogger(__name__)
+DEFAULT_CMD_TIMEOUT = SETTINGS["cmd_timeout"]
+DEFAULT_EVENT_DELAY = SETTINGS["event_delay"]
+
+
+@pytest.mark.e2e
+def test_variables_sanity(update_environment):
+    """
+    Execute a rulebook that requires multiple variables to be defined
+    at runtime. Variable values are supplied through environment variables
+    and file variables.
+    """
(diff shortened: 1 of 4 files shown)


tests/e2e/test_variables.py
```diff
@@ -0,0 +1,78 @@
...
+        utils.BASE_DATA_PATH / "rulebooks/variables/test_variables_sanity.yml"
+    )
+    vars_file = (
+        utils.BASE_DATA_PATH / "extra_vars/test_variables_extra_vars.yml"
+    )
+    cmd = utils.Command(
+        rulebook=rulebook,
+        envvars="SOURCE_EVENT_DELAY,E2E_AGENT_ID,E2E_AGENT_NAME",
+        vars_file=vars_file,
+        debug=True,
```
Alex-Izquierdo, line 42: Many things are printed in the debug mode, making the assertions useless. Don't rely in this option here.

Alex-Izquierdo (changes requested):

tests/e2e/test_variables.py
ttuffin: @Alex-Izquierdo I have thought about this and can't think of another way to validate the rule name. The only other option is to reduce from debug to verbose, which is still quite noisy. Do you have any other ideas?

Alex-Izquierdo: If I remember well, ruleset and rulename are exposed within ansible_eda variable if you use run_playbook. Otherwise, it's fine to me to skip this case. 
Eventlog also exposes it through the websocket, but I think it's making the test unnecessarily complex just for that case.

ttuffin:
@Alex-Izquierdo test updated to check rule name via run_playbook and to use default config.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #377, opened 2023-02-16
[AAP-9301] Added msg/var support in debug action

mkanoor:
https://issues.redhat.com/browse/AAP-9301

The debug action supports msg and var attributes to display single/multiple messages or variables akin to ansible debug builtin. The msg supports an array of messages.
This removes the need for a separate echo command. The echo is equivalent to
```
action:
   debug:
     msg: .....
```

### DIFF
--- a/ansible_rulebook/schema/ruleset_schema.json
+++ b/ansible_rulebook/schema/ruleset_schema.json
@@ -109,8 +109,7 @@
                                 {"$ref": "#/$defs/print-event-action"},
                                 {"$ref": "#/$defs/debug-action"},
                                 {"$ref": "#/$defs/none-action"},
-                                {"$ref": "#/$defs/shutdown-action"},
-                                {"$ref": "#/$defs/echo-action"}
+                                {"$ref": "#/$defs/shutdown-action"}
                          ]
                      }
                 },
@@ -125,8 +124,7 @@
                         {"$ref": "#/$defs/print-event-action"},
                         {"$ref": "#/$defs/debug-action"},
                         {"$ref": "#/$defs/none-action"},
-                        {"$ref": "#/$defs/shutdown-action"},
-                        {"$ref": "#/$defs/echo-action"}
+                        {"$ref": "#/$defs/shutdown-action"}
                     ]
                 }
             },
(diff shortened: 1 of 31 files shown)


docs/actions.rst
```diff
@@ -360,31 +359,56 @@ When a rule's condition(s) are satisfied we get the results back as:
...
+               - "Message 1 {{ event }}"
+               - Second Message
+
+    .. code-block:: yaml
+
+       name: debug with var
+       condition: event.i >= 5
+       action:
+          debug:
+            var: "{{ event.i }}"
```
bzwei, line 408: From this example and line 399 I don't see the difference var vs msg. Why not just use msg?

mkanoor: @bzwei var doesn't need the curly braces, we should be able to just use event.i. And use dpath to pick up the data.
In the ansible documentation it says that var should not be using curly braces.
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/debug_module.html#parameter-var
I am going to be replacing the braces and just running dpath on anything in that field. By the time it comes to the debug method it has already gone thru jinja substitution.

mkanoor: @bzwei I have removed the {{ }} which is optional

bzwei (approved):

docs/actions.rst
```diff
@@ -360,31 +359,57 @@ When a rule's condition(s) are satisfied we get the results back as:
...
      - Description
      - Required
-   * - message
-     - A terse message to display to stdout
-     - Yes
+   * - msg
+     - A simple string or an array of strings, which can have references to event or events
+     - No
+   * - var
+     - The variable to print, which can have references to event or events. Using {{ }} is optional.
```
Alex-Izquierdo, line 373: As far as I can see in the code, curly brackets are not supported here.

ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -125,8 +124,7 @@
                         {"$ref": "#/$defs/print-event-action"},
                         {"$ref": "#/$defs/debug-action"},
                         {"$ref": "#/$defs/none-action"},
-                        {"$ref": "#/$defs/shutdown-action"},
-                        {"$ref": "#/$defs/echo-action"}
```
Alex-Izquierdo, line 129: We should not remove the echo action here if the we don't remove it from the rest of the code. I suggest leaving it for a while, and given the case, remove it completely if a later PR.

Alex-Izquierdo (changes requested):
Changelog must be updated as well.

docs/actions.rst
mkanoor: @Alex-Izquierdo Every action args goes thru a substitution first and anything that is left is handled in the var for a debug.

mkanoor:
@Alex-Izquierdo Changeling updated

Alex-Izquierdo (dismissed):

mkanoor:
@Alex-Izquierdo @ttuffin Please review I had to rebase to align changes in CHANGELOG.md

Alex-Izquierdo (approved):

ttuffin (approved):

PR #378, opened 2023-02-17
Changes and Adds operator e2e tests

Alex-Izquierdo:
Refactor operator tests according to previous discussion to include multiple cases in the same rulebook to improve the execution time in addition to replace the run_playbook action in favor of echo. Makes shutdown delay configurable. Extended usecases. Implement pytest-match to be able to run the rest of the assertions of a given test even if some one fails. 

It alsfo fixes the dynaconf pkg, which was in requirements_lint instead of requirements_test

Closes https://issues.redhat.com/browse/AAP-8658


tests/e2e/test_operators.py
```diff
@@ -5,397 +5,454 @@
...
+    with check:
+        assert (
+            "Ruleset: Test relational operators rule: Finish"
+            " - test shutdown msg has initiated shutdown" in result.stdout
+        ), "Shutdown message failed"
+
+    assert len(result.stdout.splitlines()) == 21, "Unexpected output"
+
+
+def test_membership_operators(update_environment):
```
ttuffin, line 228: Missing e2e mark.

tests/e2e/test_operators.py
```diff
@@ -5,397 +5,454 @@
...
+                    for line in result.stdout.splitlines()
+                    if "test contains operator #1 passed" in line
+                ]
+            )
+            == 2
+        ), "contains operator failed"
+    assert len(result.stdout.splitlines()) == 27, "Unexpected output"
+
+
+def test_logical_operators(update_environment):
```
ttuffin, line 358: Missing e2e mark.

tests/e2e/test_operators.py
```diff
@@ -5,397 +5,454 @@
...
+    with check:
+        assert (
+            len(
+                [
+                    line
+                    for line in result.stdout.splitlines()
+                    if "test contains operator #1 passed" in line
+                ]
+            )
+            == 2
```
ttuffin, line 353: There are four rules for the "contains" operator, but only two are matching. The third is not matching because it's looking for 6.99, but the list contains 6.90. The fourth is trying to match an empty list with an empty string but this doesn't seem to work.

tests/e2e/files/rulebooks/operators/test_logical_operators.yml
```diff
@@ -0,0 +1,132 @@
...
+        echo:
+          message: "Test and operator #1 passes"
+
+    - name: Test or operator simple
+      condition: event.environment is regex(".*-prod(uction)?") or event.overload > 100
+      action:
+        echo:
+          message: "Test or operator #1 passes"
+
+    - name: Test and-or operator simple
```
ttuffin, line 106: From here onwards there are discrepancies between each rule name and the message that is echoed

tests/e2e/test_operators.py
```diff
@@ -5,397 +5,454 @@
...
+        assert (
+            len(
+                [
+                    line
+                    for line in result.stdout.splitlines()
+                    if "Test and-or operator #1 passes" in line
+                ]
+            )
+            == 1
+        ), "logical and-or operator failed"
```
ttuffin, line 422: Each and-or fail message should have more detail to outline which and-or test failed.

ttuffin (changes requested):

Alex-Izquierdo:
This was a old branch by mistake. Close it and create a new one with the final version

PR #380, opened 2023-02-20
Added ansible_eda_date_time to use in action arguments

mkanoor:
Added 2 new variables
* ansible_eda_date_time_local   - Stores Local time when the action ran
* ansible_eda_date_time_utc     - Stores UTC time when the action ran

This is similar to the **ansible_date_time** that is used in ansible-playbooks. In our use case it is generated for every action and stores the date time in local and UTC format in 2 different variables. This can be used in action args and also allows us to add a time component to internal events like set_fact and post_event actions.
e.g.

```
    - name: r3
      condition: event.url3 is regex("example\.com/others/foo")
      action:
        debug:
            first : "{{ ansible_eda_date_time_local.time }} regex works"
```

### DIFF
--- a/ansible_rulebook/eda_times.py
+++ b/ansible_rulebook/eda_times.py
@@ -1,0 +1,66 @@
+#  Copyright 2022 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


mkanoor:
@ttuffin @Alex-Izquierdo Is there a way to allow for date_time in the variables field. It will be different for every run of an action so not sure if we can compare for exactness.

Alex-Izquierdo:
@mkanoor Is this new data populated within `ansible_eda` variable sent to playbooks? I think the name is confusing, because we agreed to not use "ansible_eda" word for variables inside rulebooks.

Alex-Izquierdo:
e2e tests fails because new data is printed in the debug action

PR #384, opened 2023-02-22
Add e2e tests for string search operators

ttuffin:
String search operators include search, match and regex (with negation)

### DIFF
--- a/tests/e2e/test_operators.py
+++ b/tests/e2e/test_operators.py
@@ -437,3 +437,116 @@ def test_logical_operators(update_environment):
             )
             == 4
         ), "Testcase #6 failed"
+
+
+@pytest.mark.e2e
+def test_string_match():
+    """
+    Execute a rulebook that performs match operations on a string
+    """
+
+    rulebook = (
+        utils.BASE_DATA_PATH / "rulebooks/operators/test_string_match.yml"
+    )
+    cmd = utils.Command(rulebook=rulebook)
+
+    LOGGER.info(f"Running command: {cmd}")
+    result = subprocess.run(
+        cmd,
+        timeout=DEFAULT_CMD_TIMEOUT,
+        capture_output=True,
+        cwd=utils.BASE_DATA_PATH,
+        text=True,
+    )
+
+    with check:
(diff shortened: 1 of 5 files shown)


mkanoor (approved):

tests/e2e/test_operators.py
```diff
@@ -437,3 +437,70 @@ def test_logical_operators(update_environment):
...
+        pytest.param(
+            "test_string_search_search.yml",
+            id="string_search_search",
+        ),
+        pytest.param(
+            "test_string_search_regex.yml",
+            id="string_search_regex",
+        ),
+    ],
+)
```
Alex-Izquierdo, line 459: Hi @ttuffin you are binding the assertion logic for three different rulebooks, If we need to modify or update some of these tests it would require a refactor splitting them instead of just adding or update the testcase, it doesn't make sense, for me is a bad use of the DRY principle. I think it should be three diferent tests or only one rulebook.

ttuffin: @Alex-Izquierdo actually the idea is that they _should_ be bound. My reasoning was that they all perform a similar operation (string search) using a different approach (match, search regex). Therefore I wanted to ensure they are fed the same dataset and get to the same end result via their respective approaches. If a test is added/removed/modified for one then the same should be done with the other two. I tried one rulebook but it was becoming unmaintainable. It does mean having to update three different rulebooks, but this will at least be enforced when the test fails to ensure they remain consistent. Splitting to three different tests will mean we could lose that consistency and they will diverge. If we find later on that is what we want, then they can easily be split out at that time?

Alex-Izquierdo: `match` has different behavior than `regex` or `search`, it must be decoupled.

ttuffin: Tests decoupled

ttuffin:
@Alex-Izquierdo @mkanoor pls review. The latest changes include:

- Decouple search and regex tests from match test
- Update one test each for search and match to use regex syntax
- Update docs to include note that regex syntax is supported in match and search.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #387, opened 2023-02-23
AAP-8657 sanity test for multiple actions

Alex-Izquierdo:
- Add test case for multiple actions
- Inject shutdown after from global config
- Remove range plugin in favor of generic, ensure second ruleset ends as expected
- Asserts with pytest-check

Closes https://issues.redhat.com/browse/AAP-8657

### DIFF
--- a/tests/e2e/files/rulebooks/actions/test_actions_sanity.yml
+++ b/tests/e2e/files/rulebooks/actions/test_actions_sanity.yml
@@ -3,7 +3,8 @@
   hosts: all
   sources:
     - generic:
-        event_delay: 1
+        event_delay: "{{ DEFAULT_EVENT_DELAY }}"
+        shutdown_after: "{{ DEFAULT_SHUTDOWN_AFTER }}"
         payload:
           - action: "run_playbook"
           - action: "run_module"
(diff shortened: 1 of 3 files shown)


tests/e2e/files/rulebooks/actions/test_actions_sanity.yml
```diff
@@ -110,13 +112,27 @@
...
             msg: "Retracted fact in same ruleset, this should not be printed"
 
+    - name: Test multiple actions
+      condition: event.action == "multiple_actions"
+      actions:
+        - run_module:
+            name: ansible.builtin.debug
+            module_args:
+              msg: "Multiple action #1 triggered"
+        - debug:
```
ttuffin, line 122: @Alex-Izquierdo should we also test if the same action is used more than once? I.e. two separate debug actions.

Alex-Izquierdo: It's a *sanity* test. That case is already applied in your PR: https://github.com/ansible/ansible-rulebook/pull/375/files#diff-eafffb1901d6d199cb5a7a9823449471157c108b5f418525ca8e271ed32c0faaR18

ttuffin (approved):

mkanoor (approved):

PR #390, opened 2023-02-23
[AAP-7240] Validate is not defined in conditions

mkanoor:
https://issues.redhat.com/browse/AAP-7240
A condition cannot just contain is not defined
```
    - event.x is not defined
```
The above will raise a Validation Error

A valid use of is not defined is alongside an **and** condition
```
    - event.x is not defined and event.y == 5
```

### DIFF
--- a/ansible_rulebook/rules_parser.py
+++ b/ansible_rulebook/rules_parser.py
@@ -18,10 +18,16 @@ import ansible_rulebook.rule_types as rt
 from ansible_rulebook.condition_parser import (
     parse_condition as parse_condition_value,
 )
+from ansible_rulebook.condition_types import (
+    Condition,
+    Identifier,
+    OperatorExpression,
+)
 from ansible_rulebook.job_template_runner import job_template_runner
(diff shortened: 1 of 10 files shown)


ansible_rulebook/rules_parser.py
```diff
@@ -167,21 +173,50 @@ def parse_action(action: Dict) -> rt.Action:
...
         keys = list(condition.keys())
         if len(condition) == 1 and keys[0] in ["any", "all", "not_all"]:
             when = keys[0]
-            return rt.Condition(
+            cond = rt.Condition(
                 when,
                 [parse_condition_value(c) for c in condition[when]],
                 timeout,
             )
+            _validate(cond)
```
Alex-Izquierdo, line 189: I suggest to use a more self-descriptive name like `_validate_condition`

mkanoor:
Based on the discussion with @benthomasson this is working as designed and we should not impose restrictions suggested in this PR which breaks the retract fact

PR #392, opened 2023-02-27
Updated to use the newer drools_jpy

mkanoor:
version 0.2.3

Fixed the following issues
- selectattr operator with negation using greater/less than operators
- select operator and comparing ints and floats

https://issues.redhat.com/browse/AAP-9647
https://issues.redhat.com/browse/AAP-9616

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 ### Fixed
 
 - get_java_version, add compatibility with macs and tests for check_jvm
+- selectattr operator with negation using greater/less than operators
+- select operator and comparing ints and floats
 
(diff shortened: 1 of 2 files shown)


ttuffin (dismissed):

ttuffin:
@mkanoor should we update the changelog when we move to a new version of drools?

Alex-Izquierdo:
IMO changelog should be updated with the changes or new features that comes with drools following the same rules that we follow for other regular changes. Just saying that drools is updated is meaningless.

Alex-Izquierdo (approved):

ttuffin (approved):

PR #396, opened 2023-02-28
[AAP-9186] Use NativeTemple in Jinja2 to preserve types

mkanoor:
https://issues.redhat.com/browse/AAP-9186

https://jinja.palletsprojects.com/en/3.0.x/nativetypes/ 

Jinja2 support native types if we use NativeTemplate instead of regular template. 
This helps us preserve the data type in action args and source args.

When using native template it uses ast.literal_eval to convert it to native type. This is problematic for us if we blindly send every string into Jinja to evaluate.

```
import ast
x = ast.literal_eval("78")
type(x)
<class 'int'>
```
As shown above it converts integers wrapped as strings into integers which will cause issues for us. So I only pass it to Jinja if the string has a {{ and }} otherwise we don't pass it into jinja for evaluation.

### DIFF
--- a/ansible_rulebook/util.py
+++ b/ansible_rulebook/util.py
@@ -26,6 +26,7 @@ from typing import Any, Dict, List, Union
 import ansible_runner
 import jinja2
 import yaml
+from jinja2.nativetypes import NativeTemplate
 from packaging import version
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


ansible_rulebook/util.py
```diff
@@ -39,9 +40,12 @@ def get_horizontal_rule(character):
 
 
 def render_string(value: str, context: Dict) -> str:
-    return jinja2.Template(value, undefined=jinja2.StrictUndefined).render(
-        context
-    )
+    if "{{" in value and "}}" in value:
```
mkanoor, line 43: This is to ensure that we don't get burnt by the ast.literal_eval which will convert the integer wrapped as a string to integer.
https://jinja.palletsprojects.com/en/3.0.x/nativetypes/#jinja2.nativetypes.NativeTemplate.render

ttuffin (approved):

PR #401, opened 2023-03-01
[AAP-8337] Make inventory optional

mkanoor:
https://issues.redhat.com/browse/AAP-8337

Since ansible-rulebook can be used with Automation Controller we don't need to make inventory mandatory. This is needed for the eda-server to launch the ansible-rulebook and execute a job on the Automation Controller.

Check if the rulebook contains any actions with run_module or run_playbook and raise an exception
```
ansible_rulebook.exception.InventoryNeededException: Rule r1 has an action run_playbook which needs inventory to be defined
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,89 @@
+import pytest
+
+from ansible_rulebook.condition_types import Condition as Conditions
+from ansible_rulebook.rule_types import (
+    Action,
+    Condition,
+    EventSource,
+    EventSourceFilter,
+    Rule,
+    RuleSet,
+)
+
+
+@pytest.fixture
+def create_conditions(**kwargs):
+    def _conditions(**kwargs):
+        return Conditions(kwargs.pop("value", "2 > 1"))
+
+    return _conditions
+
+
+@pytest.fixture
+def create_condition(create_conditions, **kwargs):
+    def _condition(**kwargs):
+        when = kwargs.pop("when", "all")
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo:
I think this is not enough, according to the ticket and discussions related, we should take care if the rulebook contains run_playbook or run_module actions

benthomasson (dismissed):

mkanoor:

tests/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+def create_condition(create_conditions, **kwargs):
+    def _condition(**kwargs):
+        when = kwargs.pop("when", "all")
+        value = kwargs.pop("value", [create_conditions()])
+        return Condition(when, value)
+
+    return _condition
+
+
+@pytest.fixture
```
bzwei, line 32: @mkanoor What is the difference between `@pytest.fixture()` and `@pytest.fixture`?

Alex-Izquierdo (dismissed):

tests/conftest.py
mkanoor: @bzwei They are inter changeable and uses the default scope

mkanoor: @bzwei Nice catch, I have made them all consistent now.

bzwei (approved):

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

bzwei (approved):

Issue #397, opened 2023-03-01
rulebook json schema is invalid


ssbarnea:
The current JSON Schema file is invalid for at least two reasons and I would suggest running some schema validation using both python-jsonschema and the ajv library in order to ensure that the schema file can be consumed by most clients.

Identified problems:

* `$id` field should be a valid permalink to the file itself, now is a broken 404 url pointing to https://www.ansible.com/schema/ruleset.json - easiest fix would be to use https://raw.githubusercontent.com/ansible/ansible-rulebook/main/ansible_rulebook/schema/ruleset_schema.json

* `$schema` field is having a trailing hash and it should be exactly `https://json-schema.org/draft-04/schema` instead.

Blocking: https://github.com/ansible/ansible-lint/pull/3103

PR #405, opened 2023-03-02
Switch JSON schema to draft-7

ssbarnea:
As the support for the very old draft-4 was removed from at least
one critical tool (ajv), we switch to draft-7 which also happens to
be the version used by [all the other](https://github.com/ansible/ansible-lint/tree/main/src/ansiblelint/schemas) Ansible related schemas.

Related: https://github.com/ajv-validator/ajv/issues/472
Related: https://github.com/ansible/ansible-lint/issues/3079
Related: https://github.com/ansible/ansible-rulebook/issues/397

### DIFF
--- a/ansible_rulebook/schema/ruleset_schema.json
+++ b/ansible_rulebook/schema/ruleset_schema.json
@@ -1,9 +1,10 @@
 {
-    "$schema": "https://json-schema.org/draft-04/schema",
+    "$schema": "https://json-schema.org/draft-07/schema",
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (changes requested):
This must be updated as well:
https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/validators.py#L32

Alex-Izquierdo (approved):

ttuffin (approved):

PR #406, opened 2023-03-02
Update pre-commit hooks

ssbarnea:
This was needed because outdated isort one was not working with py311


benthomasson (approved):

Alex-Izquierdo:
We need to update this line as well: https://github.com/ansible/ansible-rulebook/blob/main/.github/workflows/ci.yml#L40

PR #410, opened 2023-03-02
Fix $id in schema

ssbarnea:
- corrects wrong $id in schema
- adds pre-commit hook that validates schema using ajv library as
  in order to prevent future regression. Note that AJV was able to
  detect $id bug while python-jsonschema was not.

### DIFF
--- a/ansible_rulebook/schema/ruleset_schema.json
+++ b/ansible_rulebook/schema/ruleset_schema.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft-07/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
     "$id": "https://raw.githubusercontent.com/ansible/ansible-rulebook/main/ansible_rulebook/schema/ruleset_schema.json",
     "type": "array",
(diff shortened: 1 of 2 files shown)


ansible_rulebook/schema/ruleset_schema.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://json-schema.org/draft-07/schema",
+    "$schema": "http://json-schema.org/draft-07/schema#",
```
Alex-Izquierdo, line 2: Why must be http?

ssbarnea: It must be http because that is what the file content itself is advertising as its self-url (`$id` field). Until the official URL is moved to https, we need to use that form.

ssbarnea: There is also a thread on https://json-schema.slack.com/archives/C5CF75URH/p1677766289988399?thread_ts=1677763627.735939&cid=C5CF75URH which you might find useful.

benthomasson (approved):

PR #413, opened 2023-03-02
Add e2e test for disabled rules

ttuffin:
Closes [AAP-9792](https://issues.redhat.com/browse/AAP-9792).

### DIFF
--- a/tests/e2e/files/rulebooks/test_disabled_rules.yml
+++ b/tests/e2e/files/rulebooks/test_disabled_rules.yml
@@ -1,0 +1,53 @@
+---
+- name: Ruleset 1
+  hosts: all
+  sources:
+    - generic:
+        event_delay: 0.5
+        payload:
+          - action: "trigger_enabled_rule"
+          - action: "trigger_disabled_rule"
+
+  rules:
+    - name: Enabled rule
+      enabled: true
+      condition: event.action == "trigger_enabled_rule"
+      actions:
+        - post_event:
+            ruleset: Ruleset 2
+            event:
+              action: "post_to_disabled_rule"
+        - set_fact:
+            fact:
+              action: "fact_to_disabled_rule"
+        - debug:
+            msg: "Enabled rule fired correctly"
+
+    - name: Disabled rule (from source event)
+      enabled: false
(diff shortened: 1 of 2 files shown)


tests/e2e/files/rulebooks/test_disabled_rules.yml
```diff
@@ -0,0 +1,53 @@
+---
+- name: Ruleset 1
+  hosts: all
+  sources:
+    - generic:
+        event_delay: 0.5
```
Alex-Izquierdo, line 6: This should be a variable

tests/e2e/files/rulebooks/test_disabled_rules.yml
```diff
@@ -0,0 +1,53 @@
...
+        payload:
+          - action: "trigger_enabled_rule"
+          - action: "trigger_disabled_rule"
+
+  rules:
+    - name: Enabled rule
+      enabled: true
```
Alex-Izquierdo, line 13: I wonder if it supports variable substitution.

ttuffin: Good question. The schema doesn't allow it. @mkanoor, what do you think about allowing variable substitution or the `enabled` param?

tests/e2e/files/rulebooks/test_disabled_rules.yml
ttuffin: The default event delay is too high so we won't use that. Do you mean to specify the value in the test, or add a new key in default.yml?

tests/e2e/files/rulebooks/test_disabled_rules.yml
mkanoor: @ttuffin I am not sure about it, so far only the name has substitution enabled. We can discuss in the dev sync call

Alex-Izquierdo: I think it could be good. Let's discuss it there.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #419, opened 2023-03-06
[AAP-7661] Support inventory in ini/yaml formats

### LINKED ISSUE #241: Using an Ansible INI file as inventory source is not supported
* ansible-rulebook version:
```
__version__ = '0.9.4'
b1098dab-ac28-4b68-b604-d4b7bcebc07c
```
* Python version: `Python 3.10.6`
* Operating System: `Ubuntu 22.04.1 LTS`

### Description

```
TASK [Gathering Facts] *********************************************************
fatal: [...]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: Could not resolve hostname ...: Name or service not known", "unreachable": true}
ok: [127.0.0.1]
```

### What I Did

```
~$ ansible-rulebook --verbose --inventory inventory --rulebook otx.yml
INFO:ansible_rulebook.app:Starting sources
INFO:ansible_rulebook.app:Starting rules
INFO:ansible_rulebook.engine:run_ruleset
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
INFO:ansible_rulebook.engine:ruleset define: {"name": "otx events", "hosts": ["all"], "sources": [{"EventSource": {"name": "Match all messages", "source_name": "ansible.eda.otx", "source_args": {"count": "1"}, "source_filters": []}}], "rules": [{"Rule": {"name": "Send to playboox", "condition": {"AllCondition": [{"IsDefinedExpression": {"Event": "otx"}}]}, "action": {"Action": {"action": "run_playbook", "action_args": {"name": "otx_ufw.yml"}}}, "enabled": true}}]}
INFO:ansible_rulebook.engine:load source
INFO:ansible_rulebook.engine:load source filters
INFO:ansible_rulebook.engine:Calling main in ansible.eda.otx
INFO:ansible_rulebook.engine:Waiting for event from otx events
INFO:ansible_rulebook.rule_generator:calling Send to playboox
INFO:ansible_rulebook.engine:call_action run_playbook
INFO:ansible_rulebook.engine:substitute_variables [{'name': 'otx_ufw.yml'}] [{'event': {'otx': {'ip': '193.105.210.113'}}, 'fact': {'otx': {'ip': '193.105.210.113'}}}]
INFO:ansible_rulebook.engine:action args: {'name': 'otx_ufw.yml'}
INFO:ansible_rulebook.builtin:running Ansible playbook: otx_ufw.yml
INFO:ansible_rulebook.builtin:ruleset: otx events, rule: Send to playboox
INFO:ansible_rulebook.builtin:Calling Ansible runner

PLAY [otx events] **************************************************************

TASK [Gathering Facts] *********************************************************
fatal: [...]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: Could not resolve hostname ...: Name or service not known", "unreachable": true}
ok: [127.0.0.1]

TASK [debug msg] ***************************************************************
ok: [127.0.0.1] => {
    "msg": "debug task: 193.105.210.113"
}

PLAY RECAP *********************************************************************
...                        : ok=0    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0   
127.0.0.1                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
^C
```

```
~$ ansible-inventory -i inventory --list
{
    "_meta": {
        "hostvars": {
            "127.0.0.1": {
                "ansible_connection": "local"
            }
        }
    },
    "all": {
        "children": [
            "ungrouped"
        ]
    },
    "ungrouped": {
        "hosts": [
            "127.0.0.1"
        ]
    }
}
```

```
~$ cat otx*.yml
---
- name: otx events
  hosts: all
  sources:
    - name: Match all messages
      ansible.eda.otx:
        count: "1"
  rules:
    - name: Send to playboox
      condition: event.otx is defined
      action:
        run_playbook:
          name: otx_ufw.yml
...
---
- name: otx events
  hosts: all
  tasks:
    - name: debug msg
      ansible.builtin.debug:
        msg: "debug task: {{ event.otx.ip }}"
...
```

mkanoor:
ansible-rulebook doesn't do anything with the inventory data. It just reads it and writes into the inventory file for ansible_runner. So instead of treating every inventory as a yaml file and load it as a dictionary, this PR just loads the inventory data as a blob irrespective of the format and sends it to ansible-runner.

Fixes #241
Closes: https://issues.redhat.com/browse/AAP-7661

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -63,7 +63,7 @@ async def run(parsed_args: argparse.ArgumentParser) -> None:
             int(parsed_args.id), parsed_args.websocket_address
         )
     else:
-        inventory = {}
+        inventory = ""
         variables = load_vars(parsed_args)
         rulesets = load_rulebook(parsed_args, variables)
         if parsed_args.inventory:
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -54,7 +54,7 @@ KEY_EDA_VARS = "ansible_eda"
 
 async def none(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -114,7 +114,7 @@ async def debug(event_log, **kwargs):
 
 async def print_event(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -148,7 +148,7 @@ async def print_event(
 
 async def set_fact(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -176,7 +176,7 @@ async def set_fact(
 
 async def retract_fact(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -203,7 +203,7 @@ async def retract_fact(
 
 async def post_event(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -229,7 +229,7 @@ async def post_event(
 
 async def run_playbook(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -326,7 +326,7 @@ async def run_playbook(
 
 async def run_module(
     event_log,
-    inventory: Dict,
+    inventory: str,
     hosts: List,
     variables: Dict,
     project_data_file: str,
@@ -346,7 +346,6 @@ async def run_module(
     extra_vars: Optional[Dict] = None,
     **kwargs,
 ):
-
     temp_dir, module_name = await pre_process_runner(
         event_log,
         inventory,
@@ -523,7 +522,7 @@ async def untar_project(output_dir, project_data_file):
 
 async def pre_process_runner(
     event_log,
-    inventory: Dict,
+    inventory: str,
     variables: Dict,
     ruleset: str,
     rulename: str,
@@ -554,7 +553,7 @@ async def pre_process_runner(
         f.write(yaml.dump(playbook_extra_vars))
     os.mkdir(inventory_dir)
     with open(os.path.join(inventory_dir, "hosts"), "w") as f:
-        f.write(yaml.dump(inventory))
+        f.write(inventory)
     os.mkdir(project_dir)
 
     logger.debug("project_data_file: %s", project_data_file)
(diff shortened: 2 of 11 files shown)


tests/e2e/utils.py
```diff
@@ -11,7 +11,7 @@
 
 BASE_DATA_PATH = Path(f"{__file__}").parent / Path("files")
 DEFAULT_SOURCES = Path(f"{__file__}").parent / Path("../sources")
-DEFAULT_INVENTORY = BASE_DATA_PATH / "inventories/default_inventory.yml"
+DEFAULT_INVENTORY = BASE_DATA_PATH / "inventories/default_inventory.ini"
```
Alex-Izquierdo, line 14: Why not instead of overriding the current inventory, we create a new one so we are able to test both?

ttuffin (approved):

tests/e2e/utils.py
mkanoor: @Alex-Izquierdo The debug action tries to do a pretty print of all its arguments which introduces new line characters.
So if the data is in yaml format we used to store this as a dictionary. When I read it as raw data it introduces new line characters and breaks this comparison.

https://github.com/ansible/ansible-rulebook/blob/70eac893fff643236c3db4334afc7ee07ac860e6/tests/e2e/test_actions.py#L100

https://github.com/ansible/ansible-rulebook/blob/70eac893fff643236c3db4334afc7ee07ac860e6/ansible_rulebook/builtin.py#L95

I struggled with keeping the yaml inventory file and tried doing all kinds of pprint tricks to get that test to pass.
This is the only test that matches the debug output.
I gave up and switched to INI file which is easier to match.

Alex-Izquierdo: yes, for this test you can use the ini, but leave the yaml for the rest, so we are able to test both. It's just a matter to define the inventory in this test, instead of the default

mkanoor:
@Alex-Izquierdo Another issue we need to handle is that there can be multiple inventory files like ansible-playbook in that case we can't load the data at all, we would have to keep the inventory file names as is and then add them into the ansible-runner inventory folder. So some of these debug test which are based on data would have to change to use file names.

mkanoor:
@Alex-Izquierdo @ttuffin Please review keep both the INI file and yaml file for e2e tests

Alex-Izquierdo (approved):
Thanks a lot for attending my request.

Alex-Izquierdo:
> @Alex-Izquierdo Another issue we need to handle is that there can be multiple inventory files like ansible-playbook in that case we can't load the data at all, we would have to keep the inventory file names as is and then add them into the ansible-runner inventory folder. So some of these debug test which are based on data would have to change to use file names.

Agree. I think we need some logic to be able to load a folder instead of a file, this is true also for dynamic inventories.

ttuffin (approved):

PR #425, opened 2023-03-07
[AAP-9956] Fixes running actions in parallel

mkanoor:
https://issues.redhat.com/browse/AAP-9956

As events come in we trigger actions without waiting for the previous actions to end


ttuffin (dismissed):

ansible_rulebook/rule_set_runner.py
```diff
@@ -208,13 +233,8 @@ async def _drain_actionplan_queue(self):
                     )
                     self.active_actions.add(task)
                     task.add_done_callback(self.active_actions.discard)
```
bzwei, line 235: It is not a bug, but will the logic slightly more readable to remove this line and delete the task from active_actions inside the new callback?

mkanoor: @bzwei They have to be distinct because in the other call back I am checking to see if the active_actions is empty to cancel the action plan task.

bzwei: @mkanoor I meant you could do `active_actions.discard(task)` in the new callback.

mkanoor: @bzwei Updated.

ttuffin (approved):

bzwei (approved):

PR #427, opened 2023-03-09
Fix broken links to collection

ttuffin:
Docs in the collection were updated which broke these links in the ansible-rulebook docs. Updating the links to point to the base repo.


Alex-Izquierdo (changes requested):
The right link should be https://github.com/ansible/event-driven-ansible#install

Alex-Izquierdo (approved):

mkanoor (approved):

PR #437, opened 2023-03-13
Adding e2e tests for shutdown action and process termination

ttuffin:
Closes [AAP-9229](https://issues.redhat.com/browse/AAP-9229).

Also covered is concurrent execution of actions in the same, and different, rulesets. [AAP-6101](https://issues.redhat.com/browse/AAP-6101).

### DIFF
--- a/tests/e2e/test_runtime.py
+++ b/tests/e2e/test_runtime.py
@@ -3,6 +3,7 @@ Module with tests for general CLI runtime
 """
 import logging
 import subprocess
+import time
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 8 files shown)


tests/e2e/test_runtime.py
```diff
@@ -102,3 +103,114 @@ def test_disabled_rules():
...
+    while i < 10:
+        line = process.stdout.readline()
+        if "{'action': 'long_loop'}" in line:
+            process.send_signal(subprocess.signal.SIGINT)
+            process.wait()
+            break
+        i += 1
+        sleep(0.5)
+    else:
+        process.kill()
```
Alex-Izquierdo, line 214: I don't understand this loop, it adds an arbitrary timeout outside the scope of the conf. Why are you not using the DEFAULT_CMD_TIMEOUT?

```
    start = time.time()
    while line := process.stdout.readline():
        if "{'action': 'long_loop'}" in line:
            break
        time.sleep(0.1)
        if time.time() - start > DEFAULT_CMD_TIMEOUT:
            process.kill()
```

ttuffin: It's a good idea, I'll use that :+1:

Alex-Izquierdo (approved):

mkanoor (approved):

PR #446, opened 2023-03-16
Fixed worker mode and activation id in parse args

mkanoor:
In an earlier fix we had made -r a required parameter which breaks when used in worker mode. In worker mode there is no rulebook the rulebook comes in as a payload on the websocket channel.

Undoes the changes from PR #431

the --id parameter is a string since it can be a GUID. Previously it was an integer

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -145,7 +145,10 @@ def load_vars(parsed_args) -> Dict[str, str]:
 def load_rulebook(
     parsed_args: argparse.ArgumentParser, variables: Optional[Dict] = None
 ) -> List[RuleSet]:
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo:
Yes, but that's it's not accurate and it's prone error. The good way imo would add logic to require the rulebook if the worker flag is not provided, which is what we want if I'm correct.

mkanoor:
@Alex-Izquierdo @ttuffin 
Please review added checks to validate combination of parameters
-w should have --id and --W
if no -w we need a -r

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #449, opened 2023-03-16
Fixes handling of activation payload from eda-server

mkanoor:
1. Handle EndOfResponse Message
2.  Removed Inventory from Payload
3. Make Project Data optional
4. Activation ID is a string
5. Share common args between worker mode and non worker mode
6. Fixed tests

### DIFF
--- a/ansible_rulebook/common.py
+++ b/ansible_rulebook/common.py
@@ -1,0 +1,29 @@
+#  Copyright 2023 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from dataclasses import dataclass, field
+from typing import Dict, List
+
+from ansible_rulebook.rule_types import RuleSet
+
+
+@dataclass
+class StartupArgs:
+    rulesets: List[RuleSet] = field(default_factory=list)
+    variables: Dict = field(default_factory=dict)
+    controller_url: str = field(default="")
+    controller_token: str = field(default="")
+    controller_ssl_verify: str = field(default="")
+    project_data_file: str = field(default="")
+    inventory: str = field(default="")
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -27,7 +27,9 @@ from ansible_rulebook.collection import (
     load_rulebook as collection_load_rulebook,
     split_collection_name,
 )
+from ansible_rulebook.common import StartupArgs
 from ansible_rulebook.engine import run_rulesets, start_source
+from ansible_rulebook.job_template_runner import job_template_runner
 from ansible_rulebook.rule_types import RuleSet, RuleSetQueue
 from ansible_rulebook.util import load_inventory
 from ansible_rulebook.validators import Validate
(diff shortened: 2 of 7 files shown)


ansible_rulebook/common.py
```diff
@@ -0,0 +1,29 @@
...
+
+from dataclasses import dataclass, field
+from typing import Dict, List
+
+from ansible_rulebook.rule_types import RuleSet
+
+
+@dataclass
+class ActivationPayload:
+    rulesets: List[RuleSet] = field(default_factory=list)
```
Alex-Izquierdo, line 23: Why is this a list of rulesets and not the whole rulebook?

hsong-rh (dismissed):

ansible_rulebook/common.py
hsong-rh: List of rulesets is the content of the whole rulebook

hsong-rh:
> @hsong-rh That issue was related to how we were doing action validation as the ruleset was being parsed. I have cleaned up that code to not do validation as the rules are being parsed, instead do it after the rules are parsed. We have a single function called validate_actions that is run to check if the actions are valid it used to check if inventory is there for run_playbook and modules. The validate_actions has been enhanced to check for controller info if the action is run_job_template. I have added a new exception called ControllerNeededException. Please try again

@mkanoor The problem solved.

ansible_rulebook/common.py
mkanoor: @Alex-Izquierdo @hsong-rh That was just a bad class name on my part. I wanted a data class that can share the arguments if they were coming from websocket or the command lines. I have renamed it to StartupArgs and the ruleset that you mention is the parsed rulesets from a passed in rulebook. In case of command line we get the file name which is parsed and loaded into this RuleSet object, in case of websocket we get the entire ruleset blob which is parsed and then used.

hsong-rh:
The change is also tested and worked in the docker env with command:
`docker run -t localhost/ansible-rulebook ansible-rulebook -w -W  ws://host.docker.internal:8000/api/eda/ws/ansible-rulebook --id 1`.

Since the output from `run_job_template` is different from `run-playbook`, aap-eda side will create a minor PR to fix it.

hsong-rh (approved):

hsong-rh:
@mkanoor I found another two issues during the testing:

1. The output action for `run_job_template` missed `matching_events` field:

```
{'type': 'Action', 'action': 'run_job_template', 'activation_id': '93', 'job_template_name': 'Hello World', 'organization': 'Default', 'job_id': '16e63411-8dd0-4658-a800-d636befc68a3', 'ruleset': 'Test run job templates', 'rule': 'Run job template', 'status': 'successful', 'run_at': '2023-03-20T18:14:55.036753Z'}
```
2. When tested with bad job template, the action failed to send:
```
2023-03-20 14:34:08,228 - ansible_rulebook.websocket - ERROR - websocket error on {'type': 'Action', 'action': 'run_job_template', 'activation_id': '95', 'playbook_name': 'Hello, New  World', 'status': 'failed', 'run_at': '2023-03-20 18:34:08.228786', 'reason': {'error': ControllerApiException('Failed to post to https://10.0.154.18/api/v2/job_templates/Hello, New  World++Default/launch/. Status: 404, Body: {"detail":"Not found."}')}}
  File "/Users/hsong/work/reviews/ansible-rulebook-madhu/ansible_rulebook/websocket.py", line 94, in send_event_log_to_websocket
    await websocket.send(json.dumps(event))
  File "/Users/hsong/.pyenv/versions/3.10.4/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/Users/hsong/.pyenv/versions/3.10.4/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/hsong/.pyenv/versions/3.10.4/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/hsong/.pyenv/versions/3.10.4/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type ControllerApiException is not JSON serializable
```

benthomasson (approved):

PR #450, opened 2023-03-17
Update actions sanity test to cover multiple sequential actions

ttuffin:
Closes [AAP-10172](https://issues.redhat.com/browse/AAP-10172).

- Update multiple_actions test and assert the complete output to ensure it's executed in the correct order
- Changed ruleset 2 to use startup delay since it's only triggered by ruleset 1 and does not need a shutdown after delay
- Test finishes slightly quicker with the addition of the shutdown_action in the multiple_actions condition

### DIFF
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -83,10 +83,10 @@ class RuleSetRunner:
             tasks.append(self.source_loop_task)
             await asyncio.wait([self.action_loop_task])
         except asyncio.CancelledError:
-            logger.info("Cancelled error caught in run_ruleset")
+            logger.debug("Cancelled error caught in run_ruleset")
             for task in tasks:
                 if not task.done():
-                    logger.info("Cancelling (2) task %s", task.get_name())
+                    logger.debug("Cancelling (2) task %s", task.get_name())
                     task.cancel()
 
         try:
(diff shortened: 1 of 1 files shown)


tests/e2e/files/rulebooks/actions/test_actions_sanity.yml
```diff
@@ -122,24 +122,22 @@
...
+            msg: "Sequential action #3: debug"
 
 - name: Second ruleset
   hosts: all
   sources:
     - generic:
-        shutdown_after: "{{ DEFAULT_SHUTDOWN_AFTER }}"
-        repeat_delay: 1
-        repeat_count: 20
+        startup_delay: 30  # never start, only triggered by Ruleset 1
```
Alex-Izquierdo, line 140: This value still needs taken from configuration, because in slow envs (like ppc) might require more time.

ttuffin: @Alex-Izquierdo we have a default value for startup delay of 1 that is already used by the shutdown_now test. Do you want me to create a new, non-default key in the config?

Alex-Izquierdo: sure, ok.

ttuffin: Due to wider changes required will take care of it in a follow up PR.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #451, opened 2023-03-17
Update tests to use default_startup_delay from config

ttuffin:
Depends on #450 

- default_startup_delay set to 30 in the config
- Tests that use this value now take it from the config
- Previously value was only used by shutdown_now test which now has its own key

### DIFF
--- a/tests/e2e/config/default.yml
+++ b/tests/e2e/config/default.yml
@@ -1,7 +1,8 @@
 default_event_delay: 1
 default_shutdown_after: 5
-default_startup_delay: 1
+default_startup_delay: 30
 operators_shutdown_after: 1
+shutdown_now_startup_delay: 1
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo (dismissed):

ttuffin:
@Alex-Izquierdo @mkanoor pls review. I have added configurable values for shutdown_now and disabled_rules tests. I also updated the process shutdown test to use a hardcoded value of 0.05, which matches the hardcoded value used in the operators tests. Tests with hardcoded values do not rely on the accuracy of these values for the tests to pass, it is just to prevent an event storm hence they do not need to be configurable.

mkanoor (dismissed):

Alex-Izquierdo (approved):

mkanoor (approved):

PR #452, opened 2023-03-17
Fix websocket protocol

benthomasson:


benthomasson:
I had a hard time turning off SSL verification. I'll revisit that next week.

ansible_rulebook/job_template_runner.py
```diff
@@ -85,10 +85,10 @@ def _auth_headers(self) -> dict:
     @cached_property
     def _sslcontext(self) -> Union[bool, ssl.SSLContext]:
         if self.host.startswith("https"):
-            if self.verify_ssl.lower() == "yes":
+            if self.verify_ssl:
                 return True
-            elif not self.verify_ssl.lower() == "no":
-                return ssl.create_default_context(cafile=self.verify_ssl)
```
benthomasson, line 91: I have no idea how this line worked.  I think cafile is supposed to be a file, but appears to be "no"???

mkanoor: yes

PR #457, opened 2023-03-20
correct JAVA_HOME for Fedora like systems

evgeni:
tested in CentOS Stream 8 and Fedora 37, the folder is called `jre-17-openjdk`, not `java-17-openjdk`


Alex-Izquierdo (approved):
True. Thanks.

ttuffin (approved):

PR #459, opened 2023-03-22
[AAP-10473] Adds support for calling eda.builtin.insert_meta_info     

mkanoor:
This builtin filter is added to every source and embellishes the event data with the following meta data

```
       {'meta': {'source': {'name': 'my_source',
                            'type': 'ansible.eda.range'}},
                'received_at': '2023-03-23T19:11:15.802274Z',
                'uuid': 'eb7de03f-6f8f-4943-b69e-3c90db346edf',
       }
```

https://issues.redhat.com/browse/AAP-10473

For internal events coming from set_fact/post_event/retract_fact we set the source name and type to **internal**.
The retract fact is currently broken because of the way Drools decides to delete the fact, since the meta portion never matches it doesn't retract the fact. Some of the tests had to be skipped because of this issue. Talking to the Drools team to see if the retract fact can skip the meta portion when doing a match.

### DIFF
--- a/ansible_rulebook/util.py
+++ b/ansible_rulebook/util.py
@@ -22,7 +22,7 @@ import subprocess
 import sys
 import tempfile
 import typing
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import ansible_runner
 import jinja2
@@ -30,7 +30,11 @@ from jinja2.nativetypes import NativeTemplate
 from packaging import version
 from packaging.version import InvalidVersion
 
+from ansible_rulebook.exception import InvalidFilterNameException
+
 logger = logging.getLogger(__name__)
+
+EDA_BUILTIN_FILTER_PREFIX = "eda.builtin."
 
 
 def get_horizontal_rule(character):
@@ -208,3 +212,27 @@ def check_jvm():
             file=sys.stderr,
         )
         sys.exit(1)
+
+
+def has_builtin_filter(name: str) -> bool:
+    return _builtin_filter_path(name)[0]
+
+
+def find_builtin_filter(name: str) -> Optional[str]:
+    found, path = _builtin_filter_path(name)
+    if found:
+        return path
+    return None
+
+
+def _builtin_filter_path(name: str) -> Tuple[bool, str]:
+    if not name.startswith(EDA_BUILTIN_FILTER_PREFIX):
+        return False, ""
+    filter_name = name.split(".")[-1]
+
+    if not filter_name:
+        raise InvalidFilterNameException(name)
+
+    dirname = os.path.dirname(os.path.realpath(__file__))
+    path = os.path.join(dirname, "event_filters", filter_name + ".py")
+    return os.path.exists(path), path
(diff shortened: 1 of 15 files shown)


Alex-Izquierdo (dismissed):

Alex-Izquierdo:
I have serious doubts about depending on the collection for a logic that is mandatory in the core of ansible-rulebook. Seems that it should be a builtin filter. @benthomasson @mkanoor

Alex-Izquierdo (dismissed):
lgtm

mkanoor:
@hsong-rh based on our discussion when we have multiple actions they will send the same matching event for all the actions individually. So the uuid is important in that case. I have added uuid to the meta data.

Single Action A1 : E1, E2.  E1 and E2 will have unique uuid's

Multiple Actions A1: E1, E2
                            A2: E1, E2
                            A3: E1, E2

With multiple actions for a single rule the E1,E2 will be reported for each action, the UUID will help keep them unique. On the server side for optimization we only need to create 2 events instead of 6 events.

mkanoor:
@Alex-Izquierdo @benthomasson @hsong-rh Please review

ansible_rulebook/util.py
```diff
@@ -208,3 +212,27 @@ def check_jvm():
...
+    if not name.startswith(EDA_BUILTIN_FILTER_PREFIX):
+        return False, ""
+    filter_name = name.split(".")[-1]
+
+    if not filter_name:
+        raise InvalidFilterNameException(name)
+
+    dirname = os.path.dirname(os.path.realpath(__file__))
+    path = os.path.join(dirname, "event_filters", filter_name + ".py")
+    return os.path.exists(path), path
```
Alex-Izquierdo, line 238: The logic of these three functions seems a bit odd for me. Returning a boolean with the result seems more a pattern of golang rather than python. I feel it could be simpler if `has_builtin_filter` just check if the source has a valid builtin filter name and `find_builtin_filter` (which seems to be more like `get_builtin_filter`) returns an exception if the source filter doesn't exist.

Alex-Izquierdo: I mean something like:
```python
def has_builtin_filter(name: str) -> bool:
    return name.startswith(EDA_BUILTIN_FILTER_PREFIX):
    
def get_builtin_filter_path(name: str) -> str:
    if not has_builtin_filter(name):
        raise InvalidFilterNameException(f"Invalid builtin filter name: {name}")
    filter_name = name.split(".")[-1]
    dirname = os.path.dirname(os.path.realpath(__file__))
    path = os.path.join(dirname, "event_filters", filter_name + ".py")
    if not os.path.exists(path):
        raise SourceFilterNotFoundException(f"Could not find source plugin: {name}")
    return path
```

Because if the rulebook contains a wrong "eda.builtin.missing" filter you want to fail at that point. 

And then you can just do:

```python
            elif has_builtin_filter(source_filter.filter_name):
                source_filter_module = runpy.run_path(
                    get_builtin_filter_path(source_filter.filter_name)
                )
```
But this is only a suggestion, I won't to block the PR.

mkanoor: @Alex-Izquierdo I don't think that solves the problem. The has_builtin_filter has to also check if the file exists so that we can continue the search to the next place where the filter can be found. If you just check the name and raise an exception when the file is missing, you cannot continue to looking for the filter in other places.

The has_builtin_filter allows you to check 
1. Does the name start with eda.builtin?
2. Does the file exist?

If both those tests pass we know we have found the filter. If not we can continue looking for the filter in other places.

PR #462, opened 2023-03-24
Fixed: inconsistent version of dpath

mkanoor:
ansible-rulebook needs dpath at version 2.1.4 or above otherwise it fails with the dpath.get error


Dostonbek1 (approved):
lgtm

PR #463, opened 2023-03-25
[AAP-10581] Compare 2 different attributes in an event

mkanoor:
https://issues.redhat.com/browse/AAP-10581
```
   event.x == event.y
```

Needs a new drools_jpy == 0.2.8


Alex-Izquierdo (approved):
It would be feasible to add a mention in the docs as well as some usecase in our e2e tests? like `test_operators` for instance?

benthomasson (approved):

PR #473, opened 2023-04-11
[AAP-11110] Added status for internal actions

mkanoor:
We were setting the status only for external actions like
 * run_playbook
 * run_module
 * run_job_template

Now we have started setting the stats for internal actions like
 * none
 * debug
 * print_event
 * set_fact
 * post_event
 * retract_fact
 * shutdown

Once the internal action is executed successfully we set the status to successful

https://issues.redhat.com/browse/AAP-11110


hsong-rh:
Tested, server side received the `status` value.
Before they show as `Unknown` like:
image
After 
image

hsong-rh (approved):

PR #474, opened 2023-04-11
Added example of multiple conditions

mkanoor:
Multiple conditions have special handling in Drools which keep the events in cache till they are used and then re-evaluated when the events match.

### DIFF
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -148,7 +148,7 @@ Once all the conditions have been met, it will return you all the events that ma
 which can be used in action.
 
     .. note::
-        Note that this case the engine will consider **all the different events** until meet the conditions,
+        Note that in this case the engine will consider **all the different events** until the conditions are met,
         regardless of whether those events come from one or multiple sources.
         Multiple conditions with ``all`` are not equivalent to a single condition with the ``and`` operator.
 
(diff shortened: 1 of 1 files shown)


docs/conditions.rst
```diff
@@ -148,7 +148,7 @@ Once all the conditions have been met, it will return you all the events that ma
 which can be used in action.
 
     .. note::
-        Note that this case the engine will consider **all the different events** until meet the conditions,
+        Note that in this case the engine will consider **all the different events** until meet the conditions,
```
ttuffin, line 151:
```suggestion
        Note that in this case the engine will consider **all the different events** until the conditions are met,
```

docs/conditions.rst
```diff
@@ -177,7 +267,7 @@ Multiple conditions where **any** one of them has to match
           debug:
 
     .. note::
-        Note that this case the engine will consider **all the different events** until meet the conditions,
+        Note that in this case the engine will consider **all the different events** until one of then meets one of the condition,
```
ttuffin, line 270:
```suggestion
        Note that in this case the engine will consider **all the different events** until one of them meets one of the conditions,
```

ttuffin (changes requested):

Alex-Izquierdo (approved):

PR #478, opened 2023-04-13
[AAP-10911] Pass in the job url for the job run

mkanoor:
This allows the UI to display the Job that ran for an action and allow the user to see the job details in the Controller.

The job_url looks like 'https://x.x.x.x/#/jobs/playbook/945/details'

https://marvelapp.com/prototype/9ag4e0f/screen/91118854


ansible_rulebook/builtin.py
```diff
@@ -932,3 +933,11 @@ def _embellish_internal_event(event: Dict, method_name: str) -> Dict:
     return insert_meta(
         event, **dict(source_name=method_name, source_type="internal")
     )
+
+
+def _controller_job_url(data: dict) -> str:
+    if "id" in data:
+        return (
+            f"{job_template_runner.host}/#/jobs/playbook/{data['id']}/details"
```
bzwei, line 941: Why `playbook` is in the path?

mkanoor: @bzwei that's the way the Controller UI shows it. The api path is different from the ui path.

ansible_rulebook/builtin.py
```diff
@@ -817,6 +817,7 @@ async def event_callback(event: dict) -> None:
             rule_uuid=source_rule_uuid,
             status=controller_job["status"],
             run_at=controller_job["created"],
+            job_url=_controller_job_url(controller_job),
```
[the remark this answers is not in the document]
mkanoor: @bzwei In the server side its going into a column called url, so maybe instead of it being specific it should just be url. Whatever that URL points to the UI will try to open it.

Alex-Izquierdo (approved):
lgtm

ansible_rulebook/builtin.py
mkanoor: @bzwei it works both ways I removed the playbook which causes the tower to issue a redirect. Since we might be running workflows in the future its better to remove the playbook from the path.

mkanoor:
@Alex-Izquierdo @bzwei @hsong-rh Please review

bzwei (approved):

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #479, opened 2023-04-13
Fix ansible-galaxy temp directory when running in openshift.

benthomasson:
ansible-galaxy fails in openshift because of the user-changing security scheme because it tries to make a temporary file in a directory that is write protected.  This is a fix to get ansible-galaxy working in openshift by writing the temp files to /tmp.

### DIFF
--- a/ansible_rulebook/collection.py
+++ b/ansible_rulebook/collection.py
@@ -12,11 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (dismissed):

ddonahue007 (dismissed):
lgtm

ddonahue007 (approved):
lgtm

mkanoor (approved):

PR #480, opened 2023-04-14
Added rule_run_at to group multiple actions in a single rule

mkanoor:
A single rule can fire multiple actions, when these actions are sent to the aap-server, it doesn't know if the rule has 1 action or more than 1 actions. The rule_run_at allows us to group all the actions into a single rule. This also helps on the server side since we are capturing when the rule engine, Drools tells us that a rule has matched and we should fire the actions.

### DIFF
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -36,7 +36,7 @@ from ansible_rulebook.rule_types import (
     EngineRuleSetQueuePlan,
 )
 from ansible_rulebook.rules_parser import parse_hosts
-from ansible_rulebook.util import substitute_variables
+from ansible_rulebook.util import run_at, substitute_variables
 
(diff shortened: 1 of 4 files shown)


ansible_rulebook/rule_set_runner.py
```diff
@@ -268,6 +274,7 @@ def _run_action(
                 action_item.ruleset_uuid,
                 action_item.rule,
                 action_item.rule_uuid,
+                group_uuid,
```
hsong-rh, line 277: should we define `group_uuid` as part of `action_item`? I feel it's more clear and future we can easily refactor `_call_action()` with two parameters `action` and `action_item`.

hsong-rh (approved):
Tested, server side receives the `rule_run_at` data

ddonahue007:
lgtm

ddonahue007 (approved):

PR #482, opened 2023-04-17
Find job_template by name

bzwei:


.pre-commit-config.yaml
```diff
@@ -5,10 +5,6 @@ repos:
     rev: 22.6.0
     hooks:
       - id: black
-  - repo: https://github.com/pycqa/isort
```
Alex-Izquierdo, line 8: Why do you remove the isort check?

ansible_rulebook/job_template_runner.py
```diff
@@ -91,6 +80,35 @@ def _sslcontext(self) -> Union[bool, ssl.SSLContext]:
...
+        slug = f"{self.JOB_TEMPLATE_SLUG}/"
+        params = {"search": name}
+
+        async with aiohttp.ClientSession(
+            headers=self._auth_headers()
+        ) as session:
+            while True:
+                response = await self._get_page(session, slug, params)
+                json_body = json.loads(response["body"])
+                for jt in json_body["results"]:
```
Alex-Izquierdo, line 93: I believe you are assuming that the request went fine and that could not happen. Should it check if the request returned 200 and `results` is in the response? otherwise you could have unexpected "jsondecode or keyerror" exceptions if there are issues with the awx connection, don't?

Alex-Izquierdo: ok, I see the status code is checked in the `_get_page` method. Should it check if the response is deserializable and it contains the `results` key?

.pre-commit-config.yaml
Alex-Izquierdo: Sorry, I didn't read the PR description. 

I can not reproduce the issue with pre-commit. Let's talk in private.

bzwei: update the version

ansible_rulebook/job_template_runner.py
bzwei: Controller list API always returns results as an array. I think we can trust the return has the desired format.

Alex-Izquierdo (approved):
lgtm

mkanoor (approved):
Tested with a job template name which has + / and | characters in the name. The resource name approach fails and this one works.

PR #486, opened 2023-04-21
[AAP-11478] Added support for --websocket-ssl-verify

mkanoor:
The websocket url can be "wss" or "ws" if we are using self signed certs in dev mode we might need to disable certificate validation

https://issues.redhat.com/browse/AAP-11478


Alex-Izquierdo (approved):
LGTM

hsong-rh (approved):

PR #493, opened 2023-04-27
fix invalid yaml in doc

tarilabs:
wrong indentation in yaml snippet in doc would make it fail to be valid yaml.

Current doc:

image

After this change:

image


ttuffin (approved):
Thank you for logging this PR.

Alex-Izquierdo (approved):

PR #496, opened 2023-04-27
[AAP-11677] Fixes web socket connection issues

mkanoor:
Due to recent changes in the backend, the web socket connection is closed prematurely from the server side leading to a misleading error on the ansible-rulebook side.
With this PR we are logging an error and adding an exception to help the user understand where the issue is.

https://issues.redhat.com/browse/AAP-11677


Alex-Izquierdo (approved):
lgtm

ttuffin (approved):

PR #501, opened 2023-05-03
[AAP-11788] Pulling job instead of job_events to get job status

bzwei:
We no longer need to log job events. Therefore pulling job for status is much faster than parsing status from job events.

Fixes AAP-11788:EDA should not hit job_events endpoint on controller jobs


Alex-Izquierdo (approved):
LGTM

I'm not sure if this should be optional. If you use the CLI as an standalone process you might want to get the output.

PR #502, opened 2023-05-05
Do not place the whole exception object in event_log

bzwei:
Exception object cannot be serialized and sent through websocket. Catch JobTemplateNotFoundException explicitly

Fixes AAP-11969: ansible-rulebook worker mode fails to handle JobTemplateNotFoundException

### DIFF
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -28,7 +28,10 @@ from drools.exceptions import (
 
 from ansible_rulebook.builtin import actions as builtin_actions
 from ansible_rulebook.conf import settings
-from ansible_rulebook.exception import ShutdownException
+from ansible_rulebook.exception import (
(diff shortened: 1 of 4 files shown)


ansible_rulebook/rule_set_runner.py
```diff
@@ -418,15 +418,15 @@ async def _call_action(
...
                 logger.error("Error calling action %s, err %s", action, str(e))
-                result = dict(error=e)
+                error = e
             except BaseException as e:
                 logger.error(e)
                 raise
         else:
             logger.error("Action %s not supported", action)
-            result = dict(error=f"Action {action} not supported")
+            error = Exception(f"Action {action} not supported")
```
mkanoor, line 427: @bzwei Can we add a new Exception called ActionFailureException instead of using the generic Exception

mkanoor (approved):

PR #503, opened 2023-05-05
Added support for websockets over unix domain socket

mkanoor:
Unix domain sockets can be used for websocket connection using the following url style

ws+unix://%2Frun%2Fansible-automation-platform%2Feda%2Fdaphne.sock/api/eda/ws/ansible-rulebook'

Since there is an embedded path in the URL the host part needs to be quoted.

https://issues.redhat.com/browse/AAP-11706

This PR would allow us to connect to the unix socket used by Daphne and we can avoid going thru NGINX and avoiding the ssl issue.

### DIFF
--- a/ansible_rulebook/websocket.py
+++ b/ansible_rulebook/websocket.py
@@ -19,6 +19,7 @@ import os
 import ssl
 import tempfile
 from asyncio.exceptions import CancelledError
+from urllib.parse import unquote, urljoin, urlparse
 
 import websockets
 import yaml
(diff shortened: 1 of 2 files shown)


mkanoor:
If we can use tcp port for a local Daphne service we won't need to use unix sockets.

Alex-Izquierdo (commented):
I think we should tell in the documentation what is the expected format for the variable if it is an unix socket. 
Also the changelog should be updated.

PR #506, opened 2023-05-12
[AAP-11904] Added execution-strategy command line option to prevent overloading controllers

mkanoor:
Added a new command line option called **--execution-strategy**

* sequential (default)
* parallel

This will help us from overloading the controller when a simple rulebook runs run_job_template with no throttling
This applies to **all actions** not just run_job_template or run_playbook or run_module.
A properly written rulebook that can handling throttling with either once_after or once_within can run things in parallel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Added
 - Sending heartbeat to the server with the session stats
+- Added command line option --execution-strategy
 
 ### Fixed
 
--- a/ansible_rulebook/cli.py
+++ b/ansible_rulebook/cli.py
@@ -164,6 +164,14 @@ def get_parser() -> argparse.ArgumentParser:
         help="Send heartbeat to the server after every n seconds"
         "Default is 0, no heartbeat is sent",
     )
+    parser.add_argument(
+        "--execution-strategy",
+        default="sequential",
+        choices=["sequential", "parallel"],
+        help="Actions can be executed in sequential order or in parallel."
+        "Default is sequential, actions will be run only after the "
+        "previous one ends",
+    )
     return parser
 
 
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -247,9 +247,14 @@ class RuleSetRunner:
                     self.active_actions.add(task)
                     task.add_done_callback(self._handle_action_completion)
                 else:
-                    self._run_action(
+                    task = self._run_action(
                         action_item.actions[0], action_item, rule_run_at
                     )
+                    if (
+                        self.parsed_args
+                        and self.parsed_args.execution_strategy == "sequential"
+                    ):
+                        await task
         except asyncio.CancelledError:
             logger.debug(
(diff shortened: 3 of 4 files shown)


Alex-Izquierdo (commented):
Changelog must be updated. 

Is this related with a bug blocker? Otherwise I don't think a feature like this should be shipped after the code freeze for 2.4

ansible_rulebook/rule_types.py
```diff
@@ -63,12 +64,18 @@ class Rule(NamedTuple):
...
+    PARALLEL = 2
+
+
 class RuleSet(NamedTuple):
     name: str
     hosts: Union[str, List[str]]
     sources: List[EventSource]
     rules: List[Rule]
     gather_facts: bool
+    execution_mode: ExecutionMode
```
benthomasson, line 78: Change the name to ExecutionStrategy and make it a command line option.

mkanoor:
@benthomasson Changed to use --execution-strategy on the command line. We don't need any extra attribute in the rulebook. The default value is sequential so actions will wait for the previous action to end.

Alex-Izquierdo:
I'm not sure, but maybe @benthomasson meant that the option should be configurable *also* through the command line and not instead of? removing the capability to set it in the rulebook is a bad idea IMO. The strategy can be something more related with the logic of the rulebook itself rather with the environment, in many cases, the strategy would be bounded to the rulebook, so it makes sense if it can be declared in the definition. Also, parallelization capabilities in ansible-playbook are defined through the code. 

In addition, I think the scope should be per rule. There are customers that might want to customize some rules. 
For example, in a rule, you have two actions, one to perform the remediation and other one to send a notification. There you need sequential strategy. Other rule could be to run two different remediations, then you might want the parallel strategy.

Alex-Izquierdo:
Last but not least, I think the usual expectation for the customer will be parallel and IMO that should be the default value.

Alex-Izquierdo:
Even if you go ahead with this approach, please don't forget that EDA UI should have an option to run the activation with this flag, otherwise all EDA users are forced to use the sequential strategy.

benthomasson:
Rulebooks might need to be declared as supporting parallel execution. So keeping the option in the rulebook could be good.

benthomasson:
Command line option should override the rulebook option except that explicit sequential rulebooks should never run in parallel.

mkanoor:
@benthomasson Should the command line option only be used to provide the default execution strategy when none is specified in the ruleset. If we change the execution strategy on the fly based on the command line parameter the ruleset might not have been designed with that execution strategy in mind. A rulebook that is designed to be run in parallel might not work properly when run in sequential mode.

benthomasson:
This PR is fine as it is.   All rulebooks should work with the sequential strategy and EDA-controller should be able to set it.
In the future we could add an option to allow parallel strategy for some rulebooks if they declare it.

benthomasson (approved):

ddonahue007 (approved):
lgtm

Alex-Izquierdo (approved):

Alex-Izquierdo:
Test failing due to race conditions, retrying...

PR #512, opened 2023-05-23
[AAP-11904] Fixed sequential execution for multiple actions

mkanoor:
This was broken for multiple actions

### DIFF
--- a/tests/e2e/test_actions.py
+++ b/tests/e2e/test_actions.py
@@ -49,6 +49,7 @@ def test_actions_sanity(update_environment):
     cmd = utils.Command(
         rulebook=rulebook,
         inventory=inventory,
+        execution_strategy="parallel",
         envvars="DEFAULT_SHUTDOWN_AFTER,"
         "DEFAULT_EVENT_DELAY,"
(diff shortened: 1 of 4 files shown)


mkanoor:

tests/e2e/test_actions.py
```diff
@@ -49,6 +49,7 @@ def test_actions_sanity(update_environment):
     cmd = utils.Command(
         rulebook=rulebook,
         inventory=inventory,
+        execution_strategy="parallel",
```
Alex-Izquierdo, line 52: Why this is needed? If sequential is the default strategy doesn't seem a good idea to run the test with parallel. I guess it is because we have fails, don't? In that case I prefer to update the affected tests. 

Also we would need more tests to ensure both strategies. I will create a ticket for it.

mkanoor: @Alex-Izquierdo That particular test is written with parallel execution in mind. So the only way that test can pass is if we set the execution strategy to parallel

Alex-Izquierdo (approved):

ttuffin (approved):

PR #516, opened 2023-05-25
propose dev reload (similar to Quarkus)

tarilabs:
Following up on similar idea from https://github.com/ansible/ansible-rulebook/pull/513,
and the "_Report on: Kubernetes + Event Driven Ansible using Rulebooks_" document shared with the team,
this PR proposes a "dev reload" similar to [Quarkus dev mode](https://quarkus.io/guides/maven-tooling#dev-mode),
that will monitor for rulebook changes and auto-reload when supplied on cli with `--dev`.

In the following, I provide a demo of its use.

"alternative 0" with the command line parameter NOT supplied, behaves as before:
Screenshot 2023-05-25 at 14 44 15

Now I launch the same, but with `--dev` command line parameter:

Screenshot 2023-05-25 at 14 46 06

it does not auto-terminate, and waits for file changes.
Now I want to modify from using the generic source to the webhook source:

Screenshot 2023-05-25 at 14 46 48

I wanted to send analogous events as those from the generic source, but through the webhook by using the postman application, but my rule is not triggering anymore 🤔 ...
[this was intentional for the purpose of the demo]

Ah yes! 😉  I should update the condition to reflect accordingly:

Screenshot 2023-05-25 at 14 47 05

In the previous screenshot, I just:

1. changed the rulebook
2. saved the rulebook
3. the ansible-rulebook dev mode downscaled and restarted automatically

As I send the same event from postman:

Screenshot 2023-05-25 at 14 47 11

It is now working as expected.

I hope this demo exemplify the intended workflow of the proposed dev mode for ansible-rulebook, inspired by Quarkus dev mode.

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -243,11 +243,16 @@ async def monitor_rulebook(rulebook_file):
     finally:
         observer.stop()
         observer.join()
-        raise HotReloadException(
-            "Rulebook file changed, "
-            + "raising exception so to asyncio.FIRST_EXCEPTION "
-            + "in order to reload"
-        )
+        # we need to check if the try-clause completed because
+        # while-loop terminated successfully, in such case we
+        # follow on the hot-reload use case, or if we got into
+        # this finally-clause because of other errors.
+        if event_handler.is_modified():
+            raise HotReloadException(
+                "Rulebook file changed, "
+                + "raising exception so to asyncio.FIRST_EXCEPTION "
+                + "in order to reload"
+            )
 
 
 async def run_rulesets(


ansible_rulebook/engine.py
```diff
@@ -281,12 +321,22 @@ async def run_rulesets(
             logger.info("Cancelling " + task.get_name())
             task.cancel()
 
+    should_reload = False
+    if (
+        monitor_task
+        and monitor_task.exception().__class__ == DevReloadException
```
Alex-Izquierdo, line 327: The right way I think it would be:
```suggestion
        and isinstance(monitor_task.exception(), DevReloadException)
```

ansible_rulebook/cli.py
```diff
@@ -172,6 +172,12 @@ def get_parser() -> argparse.ArgumentParser:
         "Default is sequential, actions will be run only after the "
         "previous one ends",
     )
+    parser.add_argument(
+        "--dev",
```
Alex-Izquierdo, line 176: I understand that this comes from quarkus, but I feel the term "dev" is quite generic and confusing. I suggest replace it by "hot-reload" which is more specific.

ansible_rulebook/engine.py
```diff
@@ -217,14 +221,45 @@ async def start_source(
...
+
+    def is_modified(self):
+        return self.modified
+
+
+async def monitor_rulebook(rulebook_file):
+    event_handler = RulebookFileChangeHandler()
+    to_observe = os.path.abspath(rulebook_file)
+    observer = Observer()
+    observer.schedule(event_handler, to_observe, recursive=True)
```
Alex-Izquierdo, line 239: I think `recursive` flag is not necessary, am I right?

Alex-Izquierdo (commented):
It produces the following log entry:
```
2023-05-25 18:58:15,645 - asyncio - ERROR - _GatheringFuture exception was never retrieved
future: <_GatheringFuture finished exception=DevReloadException('Rulebook file changed, raising exception so to asyncio.FIRST_EXCEPTION in order to reload')>
```
To prevent it, you need the `return_exceptions=True` here: https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/engine.py#L285

This is a question for @mkanoor:
`asyncio.gather` returns an awaitable, is there a reason to not use `await` in https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/engine.py#L285 ?

Alex-Izquierdo:
In addition to the linting issues, It might need to refactor some test due to recursion issues. Maybe a good approach could be encapsulate the logic of the `run` [function](https://github.com/tarilabs/ansible-rulebook/blob/c4f944eabdf4ea5e28366db50e2d249eadd4ef29/ansible_rulebook/app.py#L62) to avoid recursive calls.

ansible_rulebook/engine.py
tarilabs: forgot to remove it as I was trying it in different places earlier 😅  thanks for spotting it!

tarilabs: actually, no.
It seems to be required to capture the modification of itself, as I've just tried omitting the `recursive=True` and without it, does not trigger the `RulebookFileChangeHandler`

ansible_rulebook/cli.py
tarilabs: done in 81683a9

tarilabs:
> It produces the following log entry:
> 
> ```
> 2023-05-25 18:58:15,645 - asyncio - ERROR - _GatheringFuture exception was never retrieved
> future: <_GatheringFuture finished exception=DevReloadException('Rulebook file changed, raising exception so to asyncio.FIRST_EXCEPTION in order to reload')>
> ```
> 
> To prevent it, you need the `return_exceptions=True` here: https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/engine.py#L285

thank you, done in  81683a9

> This is a question for @mkanoor: `asyncio.gather` returns an awaitable, is there a reason to not use `await` in https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/engine.py#L285 ?

Not sure as I haven't touched that line, I assume it was that way as anyway the result of gather is not needed?
eg., even if it was
```python
unused = await asyncio.gather(*ruleset_tasks, return_exceptions=True)
```

the `unused` was not used anywhere, so not necessary to await? At least I intended it that way? 🙃 

> In addition to the linting issues, It might need to refactor some test due to recursion issues. Maybe a good approach could be encapsulate the logic of the `run` [function](https://github.com/tarilabs/ansible-rulebook/blob/c4f944eabdf4ea5e28366db50e2d249eadd4ef29/ansible_rulebook/app.py#L62) to avoid recursive calls.

Thank you, and I am facing this problem, but I'm not sure I fully understood your feedback.
Could you provide a bit more context/example? Also always available on call if preferred

Thank you so far 🙏

ansible_rulebook/engine.py
Alex-Izquierdo: Interesting, I did a couple of tests and for me it worked, maybe it depends on the platform, not sure.

tarilabs:
I would assume I just need to now fix the remaining linting issues... checking

ansible_rulebook/app.py
```diff
@@ -141,6 +152,9 @@ async def run(parsed_args: argparse.ArgumentParser) -> None:
     await event_log.put(dict(type="Exit"))
     if error_found:
         raise Exception("One of the source plugins failed")
+    elif should_reload is True:
```
Alex-Izquierdo, line 155: Minor: This would be useful only if the variable can have other values different that an boolean that could be evaluated to `True` which is not the case as the variable is to store a boolean. The recommended practice would be:
```
elif should_reload:
```

Ref:
https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations

tarilabs: > Minor: This would be useful only if the variable can have other values different that an boolean that could be evaluated to `True` which is not the case as the variable is to store a boolean. 


Thank you for highlighting this, but I beg to differ: the `should_reload` returns the cancelled future in _some_ test cases, which is why it originally recursed infinitely (it went into this branch because cancelled, was valorized as a future) and I fixed it with 8d795da I believe

Alex-Izquierdo: Oh, sorry, you are right, I forgot the whole code from last week and I need a one more coffee :)

Alex-Izquierdo (approved):
I have tested it a little with good results. LGTM

ansible_rulebook/app.py
tarilabs: no worries! 👍 🚀

benthomasson:
I really like hot-reload features for development.    As long as people are developing against non-production systems this is probably OK.

benthomasson (changes requested):
Add a big warning banner in the output and a critical level log that says that the rules will be reloaded.   Also add a critical level log that says that rules were reloaded when a hot reload happens.

benthomasson (approved):
This feature is for development environment only and does not need production testing other than that it doesn't break production execution.

tarilabs:
friendly bump / ping @mkanoor @Alex-Izquierdo can this be merged after Ben's approval https://github.com/ansible/ansible-rulebook/pull/516#pullrequestreview-1593703429 or something need further amendment in your pov, please?

ansible_rulebook/cli.py
```diff
@@ -173,6 +173,12 @@ def get_parser() -> argparse.ArgumentParser:
         "Default is sequential, actions will be run only after the "
         "previous one ends",
     )
+    parser.add_argument(
+        "--hot-reload",
+        help="Should perform hot-reload on rulebook file changes",
```
mkanoor, line 178: @tarilabs I am thinking this option is only available when the ansible-rulebook is running in non worker mode. Should we mention that in the help. Since it will be ignored when run in worker mode.

tarilabs: > Should we mention that in the help

@mkanoor in the inline-help per this suggestion, or somewhere else?

```suggestion
        help="Will perform hot-reload on rulebook file changes (when running in non-worker mode). This option is ignored in worker mode.",
```

ansible_rulebook/engine.py
```diff
@@ -217,14 +220,45 @@ async def start_source(
...
+    observer.start()
+    try:
+        while not event_handler.is_modified():
+            await asyncio.sleep(1)
+    finally:
```
mkanoor, line 243: @tarilabs when we cancel tasks because the rulebook is ending normally the monitor task is going to get cancelled,  in that case should we be executing this finally clause which will raise the execption HotReloadException, should the HotReloadException have a check to ensure that the event_handler.is_modified is True before raising the exception.

Alex-Izquierdo:
I think that as a feature that we are going to support from now on, it should have an e2e test.

tarilabs:
> I think that as a feature that we are going to support from now on, it should have an e2e test.

Since this has been raised from May 25, and I will have less opportunity to work on this "full time", can I ask you for a little guidance how to best setup and e2e test that interacts with this "as non-worker mode" and make file change that triggers the reload, please?

In other words, I'm very happy to add an e2e test, but I would really need a "bootstrap" guidance.

Alex-Izquierdo:
Hi @tarilabs I don't want to go on too long here. At a quick thought I would say that you can run the process in the background, make some simple modification in the rulebook and check that you have the expected output from the expected restart. 

This test can serve as a reference: 
https://github.com/ansible/ansible-rulebook/blob/98345aaf154e68d49acdf482e79edc45ab7f04aa/tests/e2e/test_runtime.py#L198

In this case the process is simply a fork, you can also [launch it with async](https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/test_websocket.py#L45). As you like. 

Our strategy in many places is always the same: use the debug action in the rulebook to print some string that is verified in the test. If you inspect other existing tests you will see that we use the `pytest_check` plugin to do the trick to run multiple assertions in the same test without stop the test. This is to optimize the execution time and verify multiple cases in the same rulebook instead of creating a new process for each testcase. Here you would not need it. 

Anycase, you have to update the command helper with the new flag:
https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/utils.py#L18

I don't have much time either to work on it but if it's too much trouble I can try to take some time to do the test and make a PR to your branch.

tarilabs:
@Alex-Izquierdo added test as required and per your suggestions with 43ed370 . The test is passing locally, now 🤞 on CI :)

tests/e2e/test_runtime.py
```diff
@@ -224,3 +224,62 @@ def test_terminate_process_sigint():
...
+    process = subprocess.Popen(
+        cmd,
+        cwd=utils.BASE_DATA_PATH,
+        text=True,
+        stdout=subprocess.PIPE,
+    )
+
+    fin = open(rulebook, "rt")
+    original_data = fin.read()
+    fin.close()
```
Alex-Izquierdo, line 252: It is fine, but there is a better a more pythonic way to do it with a context manager:
```suggestion
    with open(rulebook, "rt") as file:
        original_data = file.read()
```

In that way you ensure the file is properly closed even if there is some error.  

Also you can do it as well in oneline if you use the [pathlib library](https://docs.python.org/3/library/pathlib.html) with the [read_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text) method.

tests/e2e/test_runtime.py
```diff
@@ -224,3 +224,62 @@ def test_terminate_process_sigint():
...
+        time.sleep(0.1)
+        if time.time() - start > DEFAULT_CMD_TIMEOUT:
+            process.kill()
+
+    assert found_rule_1_in_out
+
+    data = original_data.replace('- action: "value_a"', '- action: "value_b"')
+    fout1 = open(rulebook, "wt")
+    fout1.write(data)
+    fout1.close()
```
Alex-Izquierdo, line 270: It is fine, but there is a better a more pythonic way to do it with a context manager:
```suggestion
    with open(rulebook, "rt") as file:
        original_data = file.write()
```

In that way you ensure the file is properly closed even if there is some error.  

Also you can do it as well in oneline if you use the [pathlib library](https://docs.python.org/3/library/pathlib.html) with the [write_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text) method.

Alex-Izquierdo (dismissed):
Thank you very much for attending my request. I think the e2e test is fine, anyway I leave you a couple of suggestions if you want to improve it.

tests/e2e/test_runtime.py
tarilabs: This comment uses "open read" I believe is a typo. I will manage this suggestion in separate commit "from my end" not from github.

Alex-Izquierdo (dismissed):

ansible_rulebook/engine.py
tarilabs: I'm not entirely sure the described scenario applies during "development time" use-case, but I get we want to be sure the exception is raised IFF the try clause actually ended correcly hence semantically to raise an exception on confirmation we are in the scenario of `event_handler.is_modified() is True` (that would happen when try-clause completes without other errors)

Will work into that, thanks

tests/e2e/test_runtime.py
Alex-Izquierdo: ups, things of the copy/paste. Good catch!

ansible_rulebook/engine.py
tarilabs: implemented with c7512df

Alex-Izquierdo (approved):

mkanoor (approved):

PR #517, opened 2023-05-25
[AAP-12513] Fix intermittent test failures because of missing events in drools

mkanoor:
Upgraded to drools_jpy 0.3.3 which had a missing event 
Fixed code coverage which was getting overwritten
Fixed a test that was using same file name in 2 different tests when run in parallel they used to fail.

https://issues.redhat.com/browse/AAP-12513

Hopefully this fixes the intermittent test failures.


Alex-Izquierdo (approved):
LGTM

PR #519, opened 2023-05-26
[AAP-11109] Support for complete and partial retract facts

mkanoor:
https://issues.redhat.com/browse/AAP-11109

We always use partial match but if for some reason the user wants to use a complete match they can set partial : False

### DIFF
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -219,9 +219,16 @@ async def retract_fact(
     rule_run_at: str,
     ruleset: str,
     fact: Dict,
+    partial: bool = True,
     name: Optional[str] = None,
 ):
-    lang.retract_fact(ruleset, _embellish_internal_event(fact, "retract_fact"))
+
+    if not partial:
+        exclude_keys = ["meta"]
+    else:
+        exclude_keys = []
+
+    lang.retract_matching_facts(ruleset, fact, partial, exclude_keys)
     await event_log.put(
         dict(
             type="Action",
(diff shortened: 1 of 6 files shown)


ansible_rulebook/builtin.py
```diff
@@ -219,9 +219,16 @@ async def retract_fact(
...
     name: Optional[str] = None,
 ):
-    lang.retract_fact(ruleset, _embellish_internal_event(fact, "retract_fact"))
+
+    if not partial:
+        exclude_keys = ["meta"]
+    else:
+        exclude_keys = []
+
+    lang.retract_matching_facts(ruleset, fact, partial, exclude_keys)
```
Alex-Izquierdo, line 231: I suggest to use named parameters in this call for a better readability.

Alex-Izquierdo (commented):
I'm confusing, in the jira ticket @benthomasson says that we should use always a partial.

tests/e2e/test_actions.py
```diff
@@ -153,12 +153,11 @@ def test_actions_sanity(update_environment):
             "Fact matched in different ruleset: sent" in result.stdout
         ), "set_fact action across rulesets failed"
 
-    # TODO: Retract fact doesn't work in Drools with the presence of meta data
-    # with check:
-    #    assert (
-    #        "Retracted fact in same ruleset, this should not be printed"
-    #        not in result.stdout
-    #    ), "retract_fact action failed"
+    with check:
```
mkanoor, line 156: @ttuffin I am bring this back after the retract fact gets fixed.

mkanoor:
@Alex-Izquierdo The default is always a partial match, since a user wouldn't know all the event attributes. In the odd case that a customer has a small payload they might want to use complete match. This is just for completeness. Since we add meta data to the event payload the complete match will always fail, if we see that the request is complete match, we can ask Drools to ignore the meta keys.

Alex-Izquierdo (approved):

PR #520, opened 2023-05-26
[AAP-12600] Synchronize creation of job request

mkanoor:
We create 1 job request at a time with an asyncio lock. Once the request has been created we can create other requests. Once the requests are created we monitor them separately. The controller can accept hundreds of requests serially.

### DIFF
--- a/ansible_rulebook/job_template_runner.py
+++ b/ansible_rulebook/job_template_runner.py
@@ -31,6 +31,8 @@ from ansible_rulebook.exception import (
 
 logger = logging.getLogger(__name__)
 
+request_create_lock = asyncio.Lock()
+
 
 class JobTemplateRunner:
     JOB_TEMPLATE_SLUG = "/api/v2/job_templates"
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
@mkanoor  Could you describe what is your plan to implement throttling requests and what is the role of this PR? Without the rest of the changes it is difficult to deduce.

mkanoor:
@Alex-Izquierdo This PR is in response to QE asking for job launch requests to be serialized. So we don't run parallel job launch requests. This PR needs #521 and you would have a rulebook with **execution_stratgey: parallel** and we would send hundreds of requests to the controller. We will create the job launch request one after the other instead of creating them in parallel. Creating them in parallel overwhelms the controller, once a request is created the controller will start them as free slots are available. This doesn't make sense when the execution_strategy is sequential since we are running one job launch and waiting for the job to end before we submit a new one.

Alex-Izquierdo (approved):
With this approach I think we could want to customize the timeout for the ClientSession which is by default of 5 minutes. https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts
Otherwise if there is some network issue or awx has problems, the whole rulebook process would be blocked. 

Also I suggest considering a [custom connection pool](https://docs.aiohttp.org/en/latest/client_advanced.html#limiting-connection-pool-size) for later iterations, which would give more control at the time of throttling requests and doesn't block the whole process.

PR #521, opened 2023-05-26
[AAP-11904] Added execution_strategy for rulesets in rulebooks

mkanoor:
Continuation of PR #506
The rulesets in a rulebook can have an optional attribute called

* execution_strategy = sequential | parallel

Default is sequential

This is still at the ruleset level and can't be configured at the rule level, which is a bit more challenging to implement.

### DIFF
--- a/ansible_rulebook/conf.py
+++ b/ansible_rulebook/conf.py
@@ -19,6 +19,7 @@ class _Settings:
     def __init__(self):
         self.identifier = str(uuid.uuid4())
         self.gc_after = 1000
+        self.default_execution_strategy = "sequential"
 
 
 settings = _Settings()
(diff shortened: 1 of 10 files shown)


ansible_rulebook/conf.py
```diff
@@ -19,6 +19,7 @@ class _Settings:
     def __init__(self):
         self.identifier = str(uuid.uuid4())
         self.gc_after = 1000
+        self.execution_strategy = "sequential"
```
Alex-Izquierdo, line 22: I suggest to rename this variable as `default_execution_strategy` to avoid potential confusions.

hsong-rh (approved):

Issue #344, opened 2023-02-03
Leverage inventory plugins for actions


quietjoy:
* ansible-rulebook version: `0.10.1`
* Python version: `3.11.0`
* Operating System: `fedora:37`

### Description

I am trying to leverage the aws inventory plugin for an action. I would like my action to run against ec2 instances deployed in aws. The aws inventory plugin allows me to dynamically look up the instances I want to run against using the aws API. 

I would also like to leverage the aws ssm connection plugin.

### What I Did

I am running ansible-rulebook pointed to my aws inventory. This is the same inventory I want my action to use.

```
ansible-rulebook --rulebook rulebook.yml -i inventory/aws_ec2.yml --verbose
```

My rulebook...
```
---
- name: Listen for events on a webhook
  hosts: all
  sources:
    - ansible.eda.webhook:
        host: 0.0.0.0
        port: 5000
  rules:
    - name: Prepare build server
      condition: event.payload.message == "Prep Build Server"
      action:
        run_playbook:
          name: tasks/build_server.yml
```

The file `tasks/build_server.yml`

```
- hosts: build_server
  vars:
    ansible_connection: aws_ssm
    ansible_aws_ssm_bucket_name: ansible-eda-demo
    ansible_aws_ssm_region: us-east-1
  tasks:
  - name: Install all packages for bender
    dnf:
      name:
      - python3-pip
      - buildah
      - podman
      - ansible
      - docker
  - name: Start dockerd
    systemd:
      name: docker
      state: started
  - name: install bender
    pip:
      name:
      - ansible-bender
```

My inventory leveraging the [aws inventory plugin](https://docs.ansible.com/ansible/latest/collections/amazon/aws/aws_ec2_inventory.html)
```
plugin: aws_ec2
regions:
  - us-east-1
filters:
  instance-state-name: running
compose:
  ansible_host: instance_id
keyed_groups:
  - key: tags.Name
    separator: ''
```

Here is the output when ansible-rulebook attempts to run the action `build_server`.

```markdown
2/3/2023, 2:16:41 PM | PLAY RECAP *************** | eda-demo
2/3/2023, 2:16:41 PM | skipping: no hosts matched | eda-demo
2/3/2023, 2:16:41 PM | [WARNING]: * Failed to parse /tmp/run_playbookmfk3b58g/inventory/hosts with | eda-demo
2/3/2023, 2:16:41 PM | yaml plugin: Plugin configuration YAML file, not YAML inventory | eda-demo
[...]

PR #523, opened 2023-06-01
If we can't find the inventory path, assume we should pass it through…

matburt:
… verbatim to Runner

This isn't an exact solution to #344 but it allows ad-hoc hosts and maybe gets it a little closer. I wonder if it would be better to just let Runner do its own thing?

### DIFF
--- a/ansible_rulebook/util.py
+++ b/ansible_rulebook/util.py
@@ -81,11 +81,12 @@ def substitute_variables(
         return value
 
 
-def load_inventory(inventory_file: str) -> Any:
-
-    with open(inventory_file) as f:
-        inventory_data = f.read()
-    return inventory_data
+def load_inventory(inventory_spec: str) -> Any:
+    if os.path.exists(inventory_spec):
+        with open(inventory_spec) as f:
+            inventory_data = f.read()
+        return inventory_data
+    return inventory_spec
 
 
 def collect_ansible_facts(inventory: str) -> List[Dict]:


ansible_rulebook/util.py
```diff
@@ -81,11 +81,12 @@ def substitute_variables(
...
-
-    with open(inventory_file) as f:
-        inventory_data = f.read()
-    return inventory_data
+def load_inventory(inventory_spec: str) -> Any:
+    if os.path.exists(inventory_spec):
+        with open(inventory_spec) as f:
+            inventory_data = f.read()
+        return inventory_data
+    return inventory_spec
```
Alex-Izquierdo, line 89: Why are you returning the string if the path does not exist? I believe that could generate errors because later the code expects to receive the content of the inventory file, like for example in https://github.com/ansible/ansible-rulebook/blob/aa7dceffd909886196f7bf122a2da20d76eabfc9/ansible_rulebook/util.py#L91

matburt: Yeah this PR is mostly just a discussion point... I'm not sure why we *need* to parse inventory like this rather than just passing things along in a more idiomatic way to Runner underneath.

This particular PR at least allow ad-hoc inventory to be passed in. There's a ton of logic in Runner and Ansible itself to manage inventory that I don't think ansible-rulebook needs to do much if anything to it.

mkanoor: @matburt Based on the [inventory notes ](https://ansible.readthedocs.io/projects/runner/en/stable/intro/#inventory)for the runner should we also allow for an env_var ANSIBLE_INVENTORY to satisfy the --inventory requirement in ansible-rulebook if the rulebook has run_playbook action? That could be the default value for [--inventory](https://github.com/ansible/ansible-rulebook/blob/67c54134be6d2bb95005463d4b1eec37b5be6c91/ansible_rulebook/cli.py#L89). It might also help if we can check the value of --inventory to see if its a file/directory/string and then do the needful based on if its 

- file that exists get copied to the Ansible-runner inventory project directory
- directory that exists gets copied along with sub directories to the Ansible-runner inventory 
- string data gets copied to inventory/hosts file like it does today

PR #525, opened 2023-06-06
[AAP-12909] Validate Controller URL and Token at Startup

mkanoor:
If the caller has provided a controller URL and token we validate that it is correct at startup so we dont have to fail when the events arrive if the url or token is invalid.

At startup we get the config information from the Controller using /api/v2/config and log the version of the controller we are using.

https://issues.redhat.com/browse/AAP-12909

The issue mentions failing on audit_rule, with this PR we fail as soon as the process starts. The Activation is marked as failed and the history has the details for bad url or bad token. This doesn't have to be replicated for every rule and event. If the URL and token is bad it will stay bad for the entire run of the activation since the URL and token cannot be changed when the activation is running.

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -87,7 +87,7 @@ async def run(parsed_args: argparse.ArgumentParser) -> None:
         startup_args.controller_ssl_verify = parsed_args.controller_ssl_verify
 
     validate_actions(startup_args)
-    set_controller_params(startup_args)
+    await validate_controller_params(startup_args)
 
     if parsed_args.websocket_address:
         event_log = asyncio.Queue()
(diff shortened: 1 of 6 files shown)


ttuffin (approved):

ansible_rulebook/app.py
```diff
@@ -245,9 +245,12 @@ def validate_actions(startup_args: StartupArgs) -> None:
                     )
 
 
-def set_controller_params(startup_args: StartupArgs) -> None:
+async def validate_controller_params(startup_args: StartupArgs) -> None:
```
Alex-Izquierdo, line 248: I think I have found duplicated code: 
https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/cli.py#L227

Alex-Izquierdo (approved):
A minor comment about duplicated code. Otherwise LGTM. 

Just a note: this PR doesn't fixed totally the bug, just prevents one of the cases. We would need to addres if AWX comes unreachable during the execution of ansible-rulebook

ansible_rulebook/app.py
mkanoor: @Alex-Izquierdo Thats a different issue we need to ensure that we have dependent arguments on the command line, if you give us the controller_url you also need to give us the token. I had tried using the parser with dependent arguments so we can say if arg1 is provided we need arg2 too. When arg2 is missing we want to exit the process. Since I couldn't implement the parser we have to have that code to check the dependent arguments.

PR #529, opened 2023-06-07
[AAP-12600] Limit the number of concurrent connections to controller

mkanoor:
Using aiohttp sessions we can limit the number of concurrent connections to the controller.
The current limit is 30 connections, it can be changed via an env var EDA_CONTROLLER_CONNECTION_LIMIT
When used in conjunction with execution_stratgey: parallel we should be able to create job requests without overloading the controller and getting a 502.

https://issues.redhat.com/browse/AAP-12600

### DIFF
--- a/ansible_rulebook/job_template_runner.py
+++ b/ansible_rulebook/job_template_runner.py
@@ -35,7 +35,6 @@ logger = logging.getLogger(__name__)
 class JobTemplateRunner:
     JOB_TEMPLATE_SLUG = "/api/v2/job_templates"
     CONFIG_SLUG = "/api/v2/config"
-    VALID_POST_CODES = [200, 201, 202]
     JOB_COMPLETION_STATUSES = ["successful", "failed", "error", "canceled"]
 
     def __init__(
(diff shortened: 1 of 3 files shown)


ansible_rulebook/job_template_runner.py
```diff
@@ -47,41 +46,33 @@ def __init__(
...
-            raise ControllerApiException(
-                "Failed to get from %s. Status: %s, Body: %s"
-                % (
-                    url,
-                    response_text["status"],
-                    response_text.get("body", "empty"),
-                )
+        self._session = None
+
+    def _create_session(self):
```
Alex-Izquierdo, line 51: I think the session should be closed at the end. 
I suggest:
```
      def _close_session(self):
          if self._session and not self._closed:
              self._session.close()
```
```
            self._session = aiohttp.ClientSession(
                connector=aiohttp.TCPConnector(limit=limit),
                headers=self._auth_headers(),
                raise_for_status=True,
            )
            atexit.register(self._close_session)
```

It requires [atexit](https://docs.python.org/3/library/atexit.html) module

Alex-Izquierdo: My suggested approach is not the most optimal, because the object will keep in memory until the end of the program and the garbage collector will not be able to collect it. But I think it is minor and I can not imagine an easy way to do it better.

Alex-Izquierdo (approved):
Lgtm

Alex-Izquierdo (approved):

ttuffin (approved):

PR #532, opened 2023-06-08
[AAP-13408] Add support for custom actions

benthomasson:
Add action loading from a directory

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -92,6 +92,8 @@ async def run(parsed_args: argparse.Namespace) -> None:
         startup_args.controller_url = parsed_args.controller_url
         startup_args.controller_token = parsed_args.controller_token
         startup_args.controller_ssl_verify = parsed_args.controller_ssl_verify
+        startup_args.source_dir = parsed_args.source_dir
+        startup_args.action_dir = parsed_args.action_dir
 
     validate_actions(startup_args)
 
@@ -107,7 +109,7 @@ async def run(parsed_args: argparse.Namespace) -> None:
     tasks, ruleset_queues = spawn_sources(
         startup_args.rulesets,
         startup_args.variables,
-        [parsed_args.source_dir],
+        [startup_args.source_dir],
         parsed_args.shutdown_delay,
     )
 
@@ -128,6 +130,7 @@ async def run(parsed_args: argparse.Namespace) -> None:
         parsed_args,
         startup_args.project_data_file,
         file_monitor,
+        [startup_args.action_dir],
     )
 
     await event_log.put(dict(type="Exit"))
--- a/ansible_rulebook/collection.py
+++ b/ansible_rulebook/collection.py
@@ -41,6 +41,9 @@ EDA_PLAYBOOKS_PATHS = [".", "playbooks"]
 
 EDA_YAML_EXTENSIONS = [".yml", ".yaml"]
 
+EDA_ACTION_PATHS = [
+    f"{EDA_PATH_PREFIX}/plugins/rule_action",
+]
 logger = logging.getLogger(__name__)
 
 
@@ -136,6 +139,24 @@ def load_rulebook(collection, rulebook):
             "collection", f"Loading rulebook from {location}"
         )
         return yaml.safe_load(f.read())
+
+
+def has_action(collection, action):
+    return has_object(
+        collection,
+        action,
+        EDA_ACTION_PATHS,
+        ".py",
+    )
+
+
+def find_action(collection, action):
+    return find_object(
+        collection,
+        action,
+        EDA_ACTION_PATHS,
+        ".py",
+    )
 
 
 def has_source(collection, source):
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -15,6 +15,8 @@
 import asyncio
 import gc
 import logging
+import os
+import runpy
 import uuid
 from pprint import pformat
 from types import MappingProxyType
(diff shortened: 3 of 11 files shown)


ansible_rulebook/rule_set_runner.py
```diff
@@ -435,6 +447,46 @@ async def _call_action(
             except BaseException as e:
                 logger.error(e)
                 raise
+        elif action_plugin := self.find_action(action):
```
mkanoor, line 450: @benthomasson are we going to have special names for the custom actions so they don't conflict with our builtin action names. If the name conflict exists we will only run the builtin and not run the custom actions.

mkanoor: @benthomasson Since the code is the same for the 2 find_actions call can we move it into a separate function
```
         def run_action_plugin(self,....):
                action_plugin = self.find_action(action) or find_action(*split_collection_name(action))
                if not action_plugin:
                   return
                    
          try:
                await action_plugin["main"](
                    event_log=self.event_log,
                    inventory=inventory,
                    hosts=hosts,
                    variables=variables,
                    project_data_file=self.project_data_file,
                    source_ruleset_name=ruleset,
                    source_ruleset_uuid=ruleset_uuid,
                    source_rule_name=rule,
                    source_rule_uuid=rule_uuid,
                    rule_run_at=rule_run_at,
                    **action_args,
                )
            except Exception as e:
                logger.error("Error calling action %s, err %s", action, str(e))
            except BaseException as e:
                logger.error(e)
                raise
```

ansible_rulebook/collection.py
```diff
@@ -134,6 +137,24 @@ def load_rulebook(collection, rulebook):
         return yaml.safe_load(f.read())
 
 
+def has_action(collection, action):
```
mkanoor, line 140: @benthomasson if we can take in an additional parameter additional_dirs, here which is the directory names we can search for the files in the collection location and an additional directory

```
    def has_action(collection, action, additional_dirs = []):
       dirs_to_search = additional_dirs
       dirs_to_search.append(EDA_ACTION_PATHS)
       return has_object(
        collection,
        action,
        dirs_to_search,
        ".py",
    )
           
```

ansible_rulebook/rule_set_runner.py
```diff
@@ -435,6 +447,46 @@ async def _call_action(
...
+                    hosts=hosts,
+                    variables=variables,
+                    project_data_file=self.project_data_file,
+                    source_ruleset_name=ruleset,
+                    source_ruleset_uuid=ruleset_uuid,
+                    source_rule_name=rule,
+                    source_rule_uuid=rule_uuid,
+                    rule_run_at=rule_run_at,
+                    **action_args,
+                )
```
mkanoor, line 484: @benthomasson do we have to send the result of the action_plugin to the AAP server, after it has finished execution. Will there be a post_event/set_fact at the end of this action. Would it make sense to extract out some of the logic from builtin which does the reporting for us.

ansible_rulebook/rule_set_runner.py
mkanoor: @benthomasson doing this search with every action call might not make sense, we could load all the actions in when we read the ruleset. Which also allows us to ensure that all the actions are valid before we start accepting the events and running the rule engine. if the actions are missing we can do an early termination to indicate that actions are missing.

Alex-Izquierdo:
What is it the story/ticket related?

ansible_rulebook/rule_set_runner.py
benthomasson: > @benthomasson are we going to have special names for the custom actions so they don't conflict with our builtin action names. If the name conflict exists we will only run the builtin and not run the custom actions.

We should support FQCN for actions.   Names without collections prefixes are either builtin or loaded from a local directory.  The precedence should be builtin and then local directory for non-FQCN.

ansible_rulebook/rule_set_runner.py
benthomasson: We probably should send the result.  Moving the reporting code out of builtin into a different module is a good idea.

benthomasson:
AAP-13408

benthomasson:
Note to self:  this needs a way to dynamically extend the rulebook schema.

Alex-Izquierdo:
We could establish that each action has to be hosted in a folder and that it should contain `action_name.py` and `action_name_schema.json` files

ansible_rulebook/app.py
```diff
@@ -92,6 +92,8 @@ async def run(parsed_args: argparse.Namespace) -> None:
         startup_args.controller_url = parsed_args.controller_url
         startup_args.controller_token = parsed_args.controller_token
         startup_args.controller_ssl_verify = parsed_args.controller_ssl_verify
+        startup_args.source_dir = parsed_args.source_dir
```
mkanoor, line 95: @benthomasson Can these go into the [settings](https://github.com/ansible/ansible-rulebook/blob/eff4dbb0fc1110e0aa8907906ec2b46d5e6453d7/ansible_rulebook/conf.py#L27) and it can be accessed from anywhere. We have been moving most of the passed in args into the settings.

ansible_rulebook/rule_set_runner.py
```diff
@@ -481,6 +515,43 @@ async def _call_action(
...
+                control, action_args, variables_copy = self._build_control(
+                    action,
+                    action_args,
+                    rules_engine_result,
+                    variables,
+                    metadata,
+                    inventory,
+                    hosts,
+                )
+                await action_plugin.main(metadata, control, **action_args)
```
mkanoor, line 548: @benthomasson Should we pass the metadata, control which are data classes as dictionaries into the action plugin so they can be tested outside of ansible rulebook otherwise they would have to know about the Metadata,Control data class definitions and if we change it it will effect the plugin. Also for the feedback should we be passing in a function that can they call to send the feedback to the eda-server. We should have a very loose coupling between action plugin and ansible-rulebook so each can be tested separately.

PR #533, opened 2023-06-08
Add inventory loading in worker mode

benthomasson:
This is useful in a use case where EDA is running isolated with the inventory provided locally.


Alex-Izquierdo:
Should we not add tests for the inventory here? https://github.com/ansible/ansible-rulebook/blob/main/tests/test_websocket.py

PR #534, opened 2023-06-08
[AAP-12909] rule_run_at was not specified when action gets exception

mkanoor:
If we get a network outage for some reason and the action fails we were not passing in the rule_run_at attribute to the aap server which would reject the action.
This PR addresses the issue.
https://issues.redhat.com/browse/AAP-12909


hsong-rh (approved):

ansible_rulebook/rule_set_runner.py
```diff
@@ -451,6 +451,7 @@ async def _call_action(
                     playbook_name=action_args.get("name"),
                     status="failed",
                     run_at=run_at(),
+                    rule_run_at=rule_run_at,
                     reason=dict(error=str(error)),
```
hsong-rh, line 455: @mkanoor Sever side accepts `message`, not `reason` when parsing the event.

mkanoor: @hsong-rh Please re-review I have changed reason to message so it can be successfully parsed on the server side. On the server side we need a database change to store the message and also change the API end point 
/api/eda/v1/audit-rules/<id>/ to provide the message to the caller especially the UI so we can display the reason why an action failed.

mkanoor: I have created https://issues.redhat.com/browse/AAP-13096 to track the server side changes for the next version

hsong-rh (approved):

Alex-Izquierdo (approved):
LGTM

PR #538, opened 2023-06-13
[AAP-13209] Skip connecting to controller if no run_job_template action

mkanoor:
At startup once we validate the rulebook for actions if there is an action called run_job_template, we will validate the connection to the controller.

https://issues.redhat.com/browse/AAP-13209


Alex-Izquierdo (approved):
lgtm

PR #539, opened 2023-06-14
AAP-10738 DROOLS-7475 Proposed feature for ['key'] accessor

tarilabs:
work-in-progress as i'm volunteering to work on this e2e, so this is a quite-complete draft for the ansible-rulebook side; i'll keep posted as i will progress on drools-ansible-rulebook-integration side.

EDIT: demonstrate working locally and added e2e test as requested

**JIRA**: https://issues.redhat.com/browse/DROOLS-7475

**referenced Pull Requests**: 

* https://github.com/ansible/ansible-rulebook/pull/539
* https://github.com/ansible/drools_jpy/pull/50
* https://github.com/kiegroup/drools-ansible-rulebook-integration/pull/56
* https://github.com/kiegroup/drools/pull/5333
* https://issues.redhat.com/browse/AAP-10738
* https://github.com/kiegroup/drools-ansible-rulebook-integration/pull/65
* https://github.com/ansible/drools_jpy/pull/51

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## [Unreleased]
 
 ### Added
+- rulebook and Drools squared accessor syntax (AAP-10738, DROOLS-7475)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 ## [Unreleased]
 
 ### Added
-- rulebook and Drools squared accessor syntax (AAP-10738, DROOLS-7475)
+- rulebook and Drools bracket notation syntax
 
 ### Fixed
 
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -48,6 +48,56 @@ The data type is of great importance for the rules engine. The following types a
 * floats (dot notation and scientific notation)
 * null
 
+Navigate structured data
+************************
+
+You can navigate strutured event, fact, var data objects using either dot notation or bracket notation:
+
+    .. code-block:: yaml
+
+      rules:
+        - name: Using dot notation
+          condition: event.something.nested == true
+          action:
+            debug:
+        - name: Analogous, but using bracket notation
+          condition: event.something["nested"] == true
+          action:
+            debug:
+
+Both of the above examples checks for the same value (attribute "nested" inside of "something") to be equal to `true`.
+
+Bracket notation might be preferable to dot notation when the structured data contains a key using symbols
+or other special characters:
+
+    .. code-block:: yaml
+
+      name: Looking for specific metadata
+      condition: event.resource.metadata.labels["app.kubernetes.io/name"] == "hello-pvdf"
+      action:
+        debug:
+
+You can find more information about dot notation and bracket notation also in the Ansible playbook `manual <https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#referencing-key-value-dictionary-variables>`_.
+
+You can access list in strutured event, fact, var data objects using bracket notation too.
+The first item in a list is item 0, the second item is item 1.
+Like Python, you can access the `n`-to-last item in the list by supplying a negative index.
+For example:
+
+    .. code-block:: yaml
+
+      rules:
+        - name: Looking for the first item in the list
+          condition: event.letters[0] == "a"
+          action:
+            debug:
+        - name: Looking for the last item in the list
+          condition: event.letters[-1] == "z"
+          action:
+            debug:
+
+You can find more information the bracket notation for list also in the Ansible playbook `manual <https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#referencing-list-variables>`_.
+
 Supported Operators
 *******************
 
@@ -796,6 +846,8 @@ Check if an item does not exist in a list based on a test
 | The value is based on the operator used, if the operator is regex then the value is a pattern.
 | If the operator is one of >,>=,<,<= then the value is either an integer or a float
 
+You can find more information for the *select* condition also in the Ansible playbook `manual <https://docs.ansible.com/ansible/latest/playbook_guide/complex_data_manipulation.html#loops-and-list-comprehensions>`_.
+
 Checking if an object exists in a list based on a test
 ------------------------------------------------------
 
@@ -833,6 +885,8 @@ Checking if an object does not exist in a list based on a test
 | The value is based on the operator used, if the operator is regex then the value is a pattern.
 | If the operator is one of >, >=, <, <= then the value is either an integer or a float.
 | If the operator is in or not in then the value is list of integer, float or string.
+
+You can find more information for the *selectattr* condition also in the Ansible playbook `manual <https://docs.ansible.com/ansible/latest/playbook_guide/complex_data_manipulation.html#loops-and-list-comprehensions>`_.
 
 
 FAQ
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ install_requires =
 	janus
 	ansible-runner
 	websockets
-	drools_jpy == 0.3.5
+	drools_jpy == 0.3.6
 
 [options.packages.find]
 include =
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
 	janus
 	ansible-runner
 	websockets
-	drools_jpy == 0.3.4
+	drools_jpy == 0.3.5
 
 [options.packages.find]
 include = 
--- a/tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml
+++ b/tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml
@@ -112,6 +112,15 @@
                     - 15
                     - 8121.99
                     - 1111
+          - id: "Testcase #12"
+            asd:
+              x:
+              - 0
+              -
+                - 0
+                - 0
+                - a:
+                    b: 3.1415
 
 
   rules:
@@ -213,3 +222,9 @@
       action:
         debug:
           msg: "Output for testcase #11"
+
+    - name: selectattr and squared accessor interaction
+      condition: event.asd["x"][1][2] is selectattr("a.b", "==", 3.1415)
+      action:
+        debug:
+          msg: "Output for testcase #12"
--- a/ansible_rulebook/condition_parser.py
+++ b/ansible_rulebook/condition_parser.py
@@ -32,6 +32,7 @@ from pyparsing import (
     infix_notation,
     one_of,
     pyparsing_common,
+    originalTextFor,
 )
 
 from ansible_rulebook.exception import (
@@ -101,7 +102,18 @@ valid_prefix = (
     | Keyword("fact")
 )
 varname = (
-    Combine(valid_prefix + ZeroOrMore("." + ident))
+    Combine(valid_prefix + ZeroOrMore(
+        ("." + ident)
+        |
+        (
+            ("[")
+            + (
+                originalTextFor(QuotedString('"'))
+                | originalTextFor(QuotedString("'"))
+            )
+            + ("]")
+        )
+    ))
     .copy()
     .add_parse_action(lambda toks: Identifier(toks[0]))
 )
--- a/ansible_rulebook/condition_parser.py
+++ b/ansible_rulebook/condition_parser.py
@@ -31,8 +31,8 @@ from pyparsing import (
     delimitedList,
     infix_notation,
     one_of,
-    pyparsing_common,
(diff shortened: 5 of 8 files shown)


Alex-Izquierdo:
I would add e2e test cases, specially to ensure the new accessor works combined with filters like selectattr, e.g: https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml

tarilabs:
> I would add e2e test cases, specially to ensure the new accessor works combined with filters like selectattr, e.g: https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml

Done with 36ecfd6

Verified locally with:
Screenshot 2023-06-23 at 13 29 19

Naturally because of these new e2e test will now _require_ first the merge of:
- https://github.com/ansible/drools_jpy/pull/50

tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml
```diff
@@ -213,3 +222,9 @@
       action:
         debug:
           msg: "Output for testcase #11"
+
+    - name: selectattr and squared accessor interaction
+      condition: event.asd["x"][1][2] is selectattr("a.b", "==", 3.1415)
```
Alex-Izquierdo, line 227: Can we use the bracket accessor inside the selectattr? this is the thing that I am most concerned.

tarilabs: I'm a bit confused: should it?

- in the ansible-rulebook doc, it only [mentions "a valid key"](https://ansible.readthedocs.io/projects/rulebook/en/stable/conditions.html#:~:text=The%20key%20is%20a%20valid%20key%20name%20in%20the%20object), so I would say actually I would _NOT_ expect that
- in the jinjia doc, there is never example of an expression (searching for "selectattr" in [this page](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.selectattr)) only of a "valid key", so I would say actually I would _NOT_ expect that

So I am putting forward this test showing the dot accessor still worked per previous behaviour (ie _before_ this PR), even if I can't find a reference that actually claims it is expected to be working.

👉  Is there a semantic definition of `selectattr` beyond the linked docs above I can reference to, please?

I'm happy to investigate if it's possible to make the syntax work as part of the string literal of the first argument of the selectattr, but I will keep the format of
```
id.something["else"]
```
etc.
In other words, like per ansible rulebook, you cannot start with a `[`.

Alex-Izquierdo: Sorry, I'm not sure If I'm misunderstanding something. From my understanding, if we support the bracket accessor users would try to do something like:
```
event.alerts is selectattr(labels["api.k8s.v1/namespace"], "==", "production")
```
Or probably 
```
event.alerts is selectattr('labels["api.k8s.v1/namespace"]', "==", "production")
```

I expected to support that kind of usecases. Am I wrong?

tarilabs: I do see your point, and thank you indeed for raising it in the first place!

My understanding is however as I highlighted in https://github.com/ansible/ansible-rulebook/pull/539#discussion_r1239772983

In other words, just by looking at the documentation of ansible-rulebook and jinjia (see links I provided above), I would _not_ have expected even the simple dot-accessor to work as some existing tests are actually checking, [example here](https://github.com/ansible/ansible-rulebook/blob/0f5e4d4291691ec23c4c67d2751da7c9293b9716/tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml#L119).

So, in this PR, I just kept existing behaviour nevertheless.

I'm open to discuss extending the `id.something["else"][0][-1]['x]` to other use-cases, but I would prefer if we ground the discussion based on some documentation/reference for the semantic to respect?

Alex-Izquierdo: As you wish. If it's a matter of a proper documentation, well, I admit that `The key is a valid key name in the object.` is not clear enough. We made that reading from the beginning and I'm just applying the same criteria here. 

If you consider that it is out of the scope, up to you. 

In that case can I ask for an update the of these docs for a better explanation to set properly the user expectation? Anyway this PR should include documentation to indicate the support for the new accessor. 

Also, it should include it in the changelog. https://github.com/ansible/ansible-rulebook/blob/main/CHANGELOG.md

tarilabs: Sure; I propose this way: with regards to the impact/relation on `selectattr` specifically, I will investigate how that works ("empirically") in _play_book; since we have this doubt, it's worthy to investigate it at once.

Regardless, will anyway update doc and changelog note.

What do you think?

setup.cfg
```diff
@@ -32,7 +32,7 @@ install_requires =
 	janus
 	ansible-runner
 	websockets
-	drools_jpy == 0.3.4
+	drools_jpy == 0.3.5
```
mkanoor, line 35: @tarilabs the drools_jpy 0.3.5 is available now.

tarilabs: Awesome!🚀💪

tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml
tarilabs: > Sure; I propose this way: with regards to the impact/relation on `selectattr` specifically, I will investigate how that works ("empirically") in _play_book; since we have this doubt, it's worthy to investigate it at once.

I have taken some time to investigate as agreed ☝️ 

## official manual and source code

I believe `selectatrr` is "a bit underspecified" in the Ansible playbook [manual](https://docs.ansible.com/ansible/latest/playbook_guide/complex_data_manipulation.html#:~:text=of%20the%20condition.-,selectattr,-/rejectattr%3A%20very%20similar); that is, the behaviour is not described beyond referencing it to Jinja.

If I go to Jinja [manual](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.selectattr:~:text=applying%20a%20test%20to%20the-,specified%20attribute,-of%20each%20object%2C%20and%20only), it is "a bit underspecified" there in the ultimate sense nowhere in the docs I seem to find how the 1st string argument is supposed to behave (just a key? or a "dot accessor expression"?).

If I look at jinja source code I need to perform at least 3 hops before finding any mention if the 1st argument is directly a key or a "dot accessor expression" or else (from [here](https://github.com/pallets/jinja/blob/7b48764688dce68e99341bec59b1eedf0c5c3ba5/src/jinja2/filters.py#L1589) to [here](https://github.com/pallets/jinja/blob/7b48764688dce68e99341bec59b1eedf0c5c3ba5/src/jinja2/filters.py#L1757) to [here](https://github.com/pallets/jinja/blob/7b48764688dce68e99341bec59b1eedf0c5c3ba5/src/jinja2/filters.py#L1723) to [here](https://github.com/pallets/jinja/blob/7b48764688dce68e99341bec59b1eedf0c5c3ba5/src/jinja2/filters.py#L62-L63) )    
and appears to be actually manipulated by splitting on the `.`.

## empirically testing _play_book

I then checked empirically using Ansible _play_book, here: https://github.com/tarilabs/demo20230626-ansible-playground#readme

So on the playbook front I would conclude, while a bit underspecified maybe, I believe most people expect it to be using as direcly as a key attribute (in "full") as that anyway is what is reported by the official manual, and actually most 3rd party tutorial I could google online as I was learning more about this feature of Ansible myself 😅 

In other words, most manual/tutorial/examples I could find are in the form of:

```
selectattr("color", "==", "red") 
```

and I could _**not**_ find any manual/tutorial/examples in the form of:

```
selectattr("navigating.deep.inside", "==", "red") 
```

ie: no squared accessor in the 1st parameter of selectattr, but splitting on `.` "brutally"

## current behaviour of drools-ansible

In drools-ansible-rulebook-integration, the `extractMapAttribute` implements currently per jinja2/actual behaviour:
https://github.com/kiegroup/drools-ansible-rulebook-integration/blob/711ad015893c0cc1f469a793df886b5415e8d7e9/drools-ansible-rulebook-integration-api/src/main/java/org/drools/ansible/rulebook/integration/api/domain/constraints/SelectAttrConstraint.java#L50

ie: no squared accessor in the 1st parameter of `selectattr`, but splitting on `.` "brutally" like playbook/jinja does.

---

So I believe on a behaviour point-of-view for rulebook as discussed here, we have two options:

## conclusions

1. **Option1**. leave current rulebook behaviour as-is, which mimics ansible playbook and jinja2 behaviour. In other words in `selectattr` use key (or at best a dot-expression). In other words, leave "underspecified" also in rulebook.

2. **Option2**. change rulebook behaviour for `selectattr` first argument to _extend_ to support squared accessors, etc. In other words diverge from ansible playbook and jinjia2 selectattr function behaviour.

If it was for me, I'd leave it at option1 without any further changes. If anything, I could make a documentation link in rulebook when speaking of selectattr, to point to Ansible _playbook_ documentation for the same function.

👉  @Alex-Izquierdo do you agree? I remain available with you and team if a quick call is preferred.

tarilabs:
@Alex-Izquierdo :
> In that case can I ask for an update the of these docs for a better explanation to set properly the user expectation? Anyway this PR should include documentation to indicate the support for the new accessor.
> Also, it should include it in the changelog. https://github.com/ansible/ansible-rulebook/blob/main/CHANGELOG.md

done with 68cc376

tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml
Alex-Izquierdo: Well, on one hand I really believe users will expect or try to use the new accessor there if it is supported, that was my original concern. 
For sure option one is the most reasonable and we are following the "emulation" of ansible for all these kind of things, so I'm not sure if we will want to implement it even if it could improve the functionality and UX. 

IMO, we can go ahead with option one and we can discuss it in later sync calls. 

Thank you again for this exhaustive and excellent contribution @tarilabs

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ## [Unreleased]
 
 ### Added
+- rulebook and Drools squared accessor syntax (AAP-10738, DROOLS-7475)
```
Alex-Izquierdo, line 7: Minor: as this is user faced I think it should not contain internal references, just the description of the feature. This is how we are doing it until now.

Alex-Izquierdo (approved):
Minor comment regarding the changelog, otherwise LGTM

CHANGELOG.md
tarilabs: done with 081421a 👍

tests/e2e/files/rulebooks/operators/test_selectattr_operator.yml
tarilabs: glad to contribute 🚀 👍

Alex-Izquierdo (approved):

docs/conditions.rst
```diff
@@ -48,6 +48,56 @@ The data type is of great importance for the rules engine. The following types a
...
+        - name: Using dot notation
+          condition: event.something.nested == true
+          action:
+            debug:
+        - name: Analogous, but using bracket notation
+          condition: event.something["nested"] == true
+          action:
+            debug:
+
+Both of the above examples checks for the same value (attribute "nested" inside of "something") to be equal to `true`.
```
mkanoor, line 68: @tarilabs What are the invalid charachters now, can we use [ ] in key names. Previously we had anything other than alphanumeric and underscore is invalid. Would we allow non english characters e.g.
```
---
- name: Simple Test
  hosts: all
  execution_strategy: parallel
  default_events_ttl: 46 day
  sources:
    - ansible.eda.generic:
        loop_count: 1
        timestamp: true
        create_index: i
        payload:
          - level: warning
            人: 的
  rules:
    - name: "Print the event"
      condition: event.人 is search("的", ignorecase=true) 
      action:
        print_event:
          pretty: True 
```

tarilabs:
required bump of drools_jpy from 51e04f9 tested, in full, locally:

Screenshot 2023-07-26 at 14 45 43

benthomasson:
I'd like to get these tests passing before approving and merging.

tarilabs:
@benthomasson 

> I'd like to get these tests passing before approving and merging.

As mentioned in exchanges, the tests are failing because:

```
ERROR: Could not find a version that satisfies the requirement drools-jpy==0.3.6 (from ansible-rulebook) 
```

it needs the drools_jpy being published.

Alex-Izquierdo (approved):

mkanoor (approved):

Issue #414, opened 2023-03-02
RFE: Add launch_job_workflow


rooftopcellist:
### Description

Feature request to add a new `run_workflow_template` action.  

Currently we have a [`run_job_template` action](https://ansible-rulebook.readthedocs.io/en/stable/actions.html#run-job-template); in a similar fashion, when conditions are met, this would launch a workflow in the specified AutomationController or AWX instance.  

### Additional Info

* ansible-rulebook version: v0.10.1

PR #542, opened 2023-06-21
Stop gap to add Workflow Job Template to Job Template Runner

l3acon:
I needed this functionality for a Summit demo and am submitting because it seems like it's something we should have. I realize there are a number of ways this could be done. If  separate classes are desired that seems like a reasonable way forward and https://github.com/ansible/ansible-rulebook/pull/537 needs to be figured out. However, this is relatively few changes.

Requested in https://github.com/ansible/ansible-rulebook/issues/414.

### DIFF
--- a/ansible_rulebook/builtin.py
+++ b/ansible_rulebook/builtin.py
@@ -733,6 +733,7 @@ async def run_job_template(
     ruleset: str,
     name: str,
     organization: str,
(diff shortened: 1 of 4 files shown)


mkanoor:
@l3acon Please review #543

l3acon:
Closing in favor of Mkanoor's PR that is looking better.

PR #543, opened 2023-06-21
feat: [AAP-9829] added run_workflow_template action

### LINKED ISSUE #414: RFE: Add launch_job_workflow

mkanoor:
Added support for run_workflow_template action
This is similar to run_job_template.

Fixes #414 

https://issues.redhat.com/browse/AAP-9829


l3acon:
Thanks Mkanoor, looks good to me!

Is it worth mentioning this addresses https://github.com/ansible/ansible-rulebook/issues/414?

tests/test_controller.py
```diff
@@ -36,6 +55,7 @@ def vcr_cassette_dir(request):
     return str(Path(f"{__file__}").parent / "cassettes")
 
 
+@pytest.mark.skip(reason="need to update data with new vcr")
```
Alex-Izquierdo, line 58: If you are replacing these tests, why not delete them instead of skipping them?

ansible_rulebook/job_template_runner.py
```diff
@@ -89,65 +90,118 @@ def _sslcontext(self) -> Union[bool, ssl.SSLContext]:
...
-    ) -> dict:
-        jt_id = await self._get_job_template_id(name, organization)
-        url = urljoin(self.host, f"{self.JOB_TEMPLATE_SLUG}/{jt_id}/launch/")
-
+    async def _launch(self, job_params: dict, url: str) -> dict:
+        body = None
         try:
             async with self._session.post(
-                url, json=job_params, ssl=self._sslcontext
+                url,
```
Alex-Izquierdo, line 193: You can call directly to json() method, and no need to raise the exception later. The print of the error is redundant, because the exception contains the body. If we want to distinguish the error we can capture the different exceptions. 

I suggest:

```
        try:
            async with self._session.post(
                url,
                json=job_params,
                ssl=self._sslcontext,
                raise_for_status=True,
            ) as post_response:
                body = await post_response.json()
                return body
        except aiohttp.ClientResponseError as e:
            logger.error("Response error: %s", str(e))
            raise ControllerApiException(str(e))   
        except aiohttp.ClientError as e:
            logger.error("Error connecting to controller %s", str(e))
            raise ControllerApiException(str(e))
```

tests/test_controller.py
```diff
@@ -22,9 +22,28 @@
...
+    JOB_TEMPLATE_POST_RESPONSE,
+    NO_JOB_TEMPLATE_PAGE1_RESPONSE,
+    ORGANIZATION_NAME,
+    UNIFIED_JOB_TEMPLATE_PAGE1_RESPONSE,
+    UNIFIED_JOB_TEMPLATE_PAGE1_SLUG,
+    UNIFIED_JOB_TEMPLATE_PAGE2_RESPONSE,
+    UNIFIED_JOB_TEMPLATE_PAGE2_SLUG,
+)
+
+os.environ["EDA_JOB_TEMPLATE_REFRESH_DELAY"] = "0.05"
```
Alex-Izquierdo, line 45: This is not a good practice because you are changing the environment for the whole process and eventually generate unexpected side effects. It would be preferably a fixture. Ideally use that fixture for all these tests that need it using "userfixtures"

def sometest():

ansible_rulebook/job_template_runner.py
mkanoor: @Alex-Izquierdo The issue is the body contains good error information like 

ERROR - Error {'extra_vars': ['Variables ansible_eda are not allowed on launch. Check the Prompt on Launch setting on the Workflow Job Template to include Extra Variables.']}

When you use the raise_for_status it doesn't parse the body since the body could be huge so we can't get the details.
Otherwise the error message would look like

ERROR - Error connecting to controller 400, message='Bad Request', url=URL('https://mycontroller/api/v2/workflow_job_templates/183/launch/')

In our case we know we are doing a POST and the response body is pretty small. so we have to do the raise_for_status=False and once we have the body we can call the raise_for_status. https://github.com/aio-libs/aiohttp/issues/3248

tests/test_controller.py
mkanoor: @Alex-Izquierdo I had this delayed setting on a test by test basis, the problem is in how the Job_template_runner instance is created. There is a single instance for the entire run and it gets loaded here. https://github.com/ansible/ansible-rulebook/blob/03b5fe4ed064f4c9b06631b5215e2f3286600101/tests/test_controller.py#L26 the env var is accessed once in the constructor. https://github.com/ansible/ansible-rulebook/blob/03b5fe4ed064f4c9b06631b5215e2f3286600101/ansible_rulebook/job_template_runner.py#L46 I am not sure how we would overwrite it. Either we would have to check for this env_var for every get/post call or add an argument to pass in the refresh_delay

ansible_rulebook/job_template_runner.py
Alex-Izquierdo: You are right, I assumed wrongly the body it was in exception.

tests/test_controller.py
Alex-Izquierdo: Why not modify the attribute after instantiaton as you do with other attributes in the tests? https://github.com/ansible/ansible-rulebook/pull/543/files#diff-f7dbf2e52de217bf88ebb0364a5529e8f303b44ed043c4c3e4a1a1d681588ea4R116
In fact, you don't need to do it for each test because you are modifying the global instance. Or even better, you could create a fixture for it: 

```
@pytest.fixture
def mocked_job_template_runner():
    from ansible_rulebook.job_template_runner import job_template_runner
    job_template_runner.host = "https://example.com"
    job_template_runner.token = "DUMMY"
    job_template_runner.refresh_delay = 0.5
    return job_template_runner
```

tests/test_controller.py
mkanoor: Removed

Alex-Izquierdo (approved):
lgtm

ttuffin (dismissed):

ansible_rulebook/job_template_runner.py
```diff
@@ -27,13 +27,14 @@
...
+    WorkflowJobTemplateNotFoundException,
 )
 
 logger = logging.getLogger(__name__)
 
 
 class JobTemplateRunner:
-    JOB_TEMPLATE_SLUG = "/api/v2/job_templates"
+    UNIFIED_TEMPLATE_SLUG = "/api/v2/unified_job_templates/"
     CONFIG_SLUG = "/api/v2/config"
```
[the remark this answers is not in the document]
Alex-Izquierdo: Added by myself in https://github.com/ansible/ansible-rulebook/pull/543/commits/eae94f20735ba29e35a694acf75db8c8b9add257

Alex-Izquierdo (dismissed):

benthomasson (approved):

Alex-Izquierdo (approved):

PR #545, opened 2023-06-22
test: tool to convert rulebook yml to ast format

mkanoor:
ansible-rulebook internally uses code to convert a rulebook from yml format to a Abstract Syntax Tree (ast) format. This AST is then passed into the rule engine. The parsing is done using pyparsing in ansible-rulebook. When writing tests for drools_jpy we need the AST format, since the code is hidden inside ansible-rulebook its not easy to access it. This tool exposes that ast conversion feature so we can convert one rulebook at a time to AST format for testing. If the rulebook has references to vars and env vars they can be passed into the converter since it would substitute the variables in the AST.

usage:

python3 tools/convert_to_ast.py [-h] -r RULEBOOK [-e VARS] [-o OUTPUT] [-E ENV_VARS]

### DIFF
--- a/tools/convert_to_ast.py
+++ b/tools/convert_to_ast.py
@@ -1,0 +1,83 @@
+import argparse
+import json
+import os
+import sys
+from pathlib import Path
+from typing import List
+
+import yaml
+
+from ansible_rulebook.json_generator import visit_ruleset
+from ansible_rulebook.rules_parser import parse_rule_sets
+
+
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "-r",
+        "--rulebook",
+        help="The rulebook file or rulebook from a collection",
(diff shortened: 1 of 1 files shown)


tools/convert_to_ast.py
```diff
@@ -0,0 +1,83 @@
...
+        required=True,
+    )
+    parser.add_argument(
+        "-e",
+        "--vars",
+        help="Variables file",
+    )
+    parser.add_argument(
+        "-a",
+        "--ast",
```
Alex-Izquierdo, line 30: This is typically `-o` `--output`

Alex-Izquierdo (approved):
Minor comment, LGTM

tools/convert_to_ast.py
mkanoor: @Alex-Izquierdo Updated

Alex-Izquierdo (approved):

PR #551, opened 2023-06-29
feat: support inventory as a single file or directory

mkanoor:
We don't load the inventory data in ansible-rulebook its passed into the ansible-runner as is.

Based on work done in PR #523 

https://github.com/ansible/ansible-rulebook/issues/344
https://issues.redhat.com/browse/AAP-7661


ansible_rulebook/builtin.py
```diff
@@ -555,6 +558,11 @@ def cancel_callback():
                     verbosity=verbosity,
                     event_handler=event_callback,
                     cancel_callback=cancel_callback,
+                    inventory=os.path.join(
+                        private_data_dir,
+                        "inventory",
+                        os.path.basename(inventory),
```
Alex-Izquierdo, line 564: Is there a reason to use a nested folder instead of directly "inventory" dir?

Alex-Izquierdo (approved):
Inventory is value that remains through the live cycle of the process. Have you considered to moving this kind of things to a shared place, like https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/conf.py#L18 ?

That would allow us to reduce the number of params passed through functions. 

A couple of questions, but LGTM

ansible_rulebook/builtin.py
mkanoor: @Alex-Izquierdo I am following the directory structure that they recommend for ansible-runner input
https://ansible.readthedocs.io/projects/runner/en/stable/intro/#runner-input-directory-hierarchy

Alex-Izquierdo: Ok,

benthomasson (approved):

Issue #554, opened 2023-07-19
run_job_template action returns 401 when 443 port is present in CONTROLLER_URL


Alex-Izquierdo:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/ansible-rulebook/issues) for duplicates.
- [X] I understand that ansible-rulebook is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When CONTROLLER_URL has an https scheme and it contains the default port (443) AWX returns a redirect without the port.[ A bug related with aiohttp library](https://github.com/aio-libs/aiohttp/issues/7381) causes the deletion of the authorization header ending in a 401 response. 

Regardless if the port is considered or not part of the host, this error can be avoided avoiding unnecessary redirects from AWX. This would not happen if the port is not the standard one (in that case AWX would return the port in the redirect and aiohttp client would resend the authorization header)

### Environment

ansible-rulebook v1.0.0

### Steps to reproduce

```ansible-rulebook -r somerulebook.yml --controller-url https://some-awx:443 --controller-token some-valid-token --controller-ssl-verify no
```

### Actual results

```
2023-07-18 19:23:22,870 - ansible_rulebook.job_template_runner - ERROR - Error connecting to controller 401, message='Unauthorized', url=URL('https://some-awx:443/api/v2/config/')
2023-07-18 19:23:22,872 - ansible_rulebook.cli - ERROR - Terminating 401, message='Unauthorized', url=URL('https://containerized.ansible.com/api/v2/config/')
```

### Expected results

The rulebook should work because it was provided a valid token.

PR #555, opened 2023-07-19
fix: wrong 401 from AWX with 443 port - issue #554 

Alex-Izquierdo:
Avoid unnecessary redirects from AWX. 
Fixes https://github.com/ansible/ansible-rulebook/issues/554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 
 ### Fixed
+- Fixes a wrong 401 response from AWX when 443 is present in an https CONTROLLER_URL (<https://github.com/ansible/ansible-rulebook/issues/554>)
 
 ### Removed
 - Remove official support for Python 3.8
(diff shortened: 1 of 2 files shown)


hsong-rh (approved):

mkanoor (approved):
Looks good maybe we should rename the PR title to append trailing slash to all URLs for the controller to save on the  redirect that is issued by the Django app if a URL doesn't have a trailing slash. It saves an extra trip to the server

mkanoor:
@Alex-Izquierdo Would you be fixing the vcr data with this PR? In another #543 I have removed the VCR test data

Alex-Izquierdo:
I forgot the existence of that PR. Honestly, I would prefer to ship this fix in that PR. @mkanoor

Alex-Izquierdo:
Closed in favor of https://github.com/ansible/ansible-rulebook/pull/543 that includes the fix in https://github.com/ansible/ansible-rulebook/pull/543/commits/eae94f20735ba29e35a694acf75db8c8b9add257

PR #557, opened 2023-07-24
Bump version: 1.0.0 → 1.0.1

Alex-Izquierdo:


Alex-Izquierdo:
Only changes with meaning for the final user must be included in the changelog @ttuffin

ttuffin (approved):
@Alex-Izquierdo but is the addition of a new tool not meaningful to a user?

Alex-Izquierdo:
@ttuffin There is anything for the final user, just for engineering. That's why the changelog wasn't updated at that point.

Alex-Izquierdo:
@ttuffin Note as well that tools or pre-commit are not shipped in the distribution.

PR #562, opened 2023-08-03
fix: wait for the feedback task to end

mkanoor:
[AAP-13874](https://issues.redhat.com/browse/AAP-13874)

ansible-rulebook uses a websocket channel to send feedback events for reporting purposes. The feedback task runs asynchronously and might get shutdown/cancelled before it can post all of the events. This PR ensures that the feedback task gets sufficient time to post all of the events back to the server


ttuffin (dismissed):

Alex-Izquierdo (dismissed):
lgtm

ansible_rulebook/websocket.py
```diff
@@ -104,22 +104,26 @@ async def send_event_log_to_websocket(
 
             while True:
                 event = await event_log.get()
+                logger.debug(f"Event received, {event}")
                 await websocket.send(json.dumps(event))
 
-                if event == dict(type="Shutdown"):
+                if event == dict(type="Exit"):
```
hsong-rh, line 110: @mkanoor why do we change type from Shutdown to Exit? Server side has a list of acceptable types. We need add `Exit` into it.
https://github.com/ansible/eda-server/blob/main/src/aap_eda/wsapi/consumers.py#L34

mkanoor: @hsong-rh That is an internal message asking the feedback task to end, it will not be sent to the server. I have fixed it to not send the Exit message.

hsong-rh (approved):

PR #563, opened 2023-08-03
fix: send session_stats if heartbeat enabled after a rule is triggered

mkanoor:
[AAP-12987](https://issues.redhat.com/browse/AAP-12987)
Previously we were only sending the session stats on timed intervals and not when a rule gets triggered. This was causing the lag as reported in the ticket. With this PR we are now sending the session stats when a rule fires in addition to the timed intervals.


Dostonbek1 (approved):
tested; works as expected.

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #565, opened 2023-08-06
feat: support firing of multiple rules

mkanoor:
https://issues.redhat.com/browse/AAP-9755
https://issues.redhat.com/browse/AAP-13131

Drools has always supported firing of multiple rules when an event or a fact comes in. Durable rules only supported this behavior for facts but not for events. This PR adds a feature to a ruleset to indicate if it should match_multiple rules. The default is false

https://github.com/ansible/ansible-rulebook/assets/6452699/20daa5aa-75c7-4c63-93f0-01678ba4c986

### DIFF
--- a/ansible_rulebook/rule_types.py
+++ b/ansible_rulebook/rule_types.py
@@ -78,6 +78,7 @@ class RuleSet(NamedTuple):
     gather_facts: bool
     uuid: Optional[str] = None
     default_events_ttl: Optional[str] = None
+    match_multiple_rules: bool = False
 
 
 class ActionContext(NamedTuple):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## [Unreleased]
 
 ### Added
+- support for firing multiple rules
 
 ### Fixed
 
--- a/ansible_rulebook/json_generator.py
+++ b/ansible_rulebook/json_generator.py
@@ -309,6 +309,9 @@ def visit_ruleset(ruleset: RuleSet, variables: Dict):
     if ruleset.default_events_ttl:
         data["default_events_ttl"] = ruleset.default_events_ttl
 
+    if ruleset.match_multiple_rules:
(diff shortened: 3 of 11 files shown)


Alex-Izquierdo:
This PR doesn't require an upgrade of drools-jpy? I believe it depends of some PR to be merged in the drools-integration project.

Alex-Izquierdo (changes requested):
Release notes update missing.

mkanoor:
@Alex-Izquierdo this was added in drools-jpy via https://issues.redhat.com/browse/AAP-13076, the current drools_jpy already has this change, we don't need a new drools-jpy

Alex-Izquierdo (dismissed):
I really think it should have some e2e test as well, regardless of that LGTM

ansible_rulebook/rule_types.py
```diff
@@ -78,6 +78,7 @@ class RuleSet(NamedTuple):
     gather_facts: bool
     uuid: Optional[str] = None
     default_events_ttl: Optional[str] = None
+    match_multiple_rules: Optional[bool] = False
```
Alex-Izquierdo, line 81: Minor: type hint is wrong. If the value is not nullable, it should not be optional.
```suggestion
    match_multiple_rules: bool = False
```

ttuffin:
> I really think it should have some e2e test as well, regardless of that LGTM

+1. It would be great if any new features for rulebook will have coverage added to the e2e tests.

mkanoor:
@Alex-Izquierdo @ttuffin I have added the e2e tests

benthomasson:
@mkanoor this exposes a bug in the Rule Audit view assumptions.   This page is rule-focused and requires hunting for the event that caused the rule to fire instead of being event centric and showing what rules and actions fired from that event.

benthomasson:
We should hold off on merging this until we have a meeting to discuss the Rule Audit view.

benthomasson (approved):

PR #566, opened 2023-08-08
Update docs for run_job_template action

cloin:
Docs update

It is not clear that in order to use event facts and vars within a job template that "Prompt on launch" must be enabled for Variables and Limit on the job template configuration page. Alternatively, `ansible_eda` can be added to a survey to pass event info to the job template.

This is an incredibly common question when testing EDA.


ttuffin (approved):

docs/actions.rst
```diff
@@ -103,6 +103,9 @@ Run a job template.
 
 .. note::
     ``--controller-url`` and ``--controller-token`` cmd options must be provided to use this action
+    
+    In order to access event information under the ``ansible_eda`` namespace, be sure to check the box for "Prompt on launch" for the Variables field within the job template. Alternatively, a survey can be created that includes the variable ``ansible_eda``. Similarly, if you plan to limit host execution based on event information, enable "Prompt on launch" for the Limit field within the job template.
+
```
mkanoor, line 108: @cloin If a Job Template doesn't have the "Prompt on Launch" should we send the extra vars to the Job Template. It's possible that the Job Template is self sufficient and doesn't need the extra vars or the ansible_eda data. We have fixed this for workflows, in case of workflows if the "Prompt on Launch" is not enabled the Workflow launch fails if we blindly send the extra vars. The run_workflow_template action was merged recently

mkanoor (approved):

PR #568, opened 2023-08-09
Refactor messages into types

benthomasson:
Messages should have well-defined types so that senders and receivers can agree on the data that should be in the message.    Refactor the messages into dataclasses.


Alex-Izquierdo:
Have you considered to use pydantic? I wonder their builtin capabilites for serialization and validation might be useful for us.

benthomasson:
Yes, pydantic is even better than dataclasses.

PR #573, opened 2023-08-17
Connecting to controller if run_workflow_template too

akira6592:
In ansible-rulebook 1.0.1 (#538), the connection to the controller was checked only when using the `run_job_template` action.

In ansible-rulebook 1.0.2 (#543), `run_workflow_template` action was added.

This PR allows checking the connection to the controller even when using the `run_workflow_template` action.

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -57,6 +57,7 @@ class NullQueue:
 
 logger = logging.getLogger(__name__)
 INVENTORY_ACTIONS = ("run_playbook", "run_module")
+CONTROLLER_ACTIONS = ("run_job_template", "run_workflow_template")
 
 
 # FIXME(cutwater): Replace parsed_args with clear interface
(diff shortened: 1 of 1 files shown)


ansible_rulebook/app.py
```diff
@@ -234,7 +234,10 @@ def validate_actions(startup_args: StartupArgs) -> None:
     for ruleset in startup_args.rulesets:
         for rule in ruleset.rules:
             for action in rule.actions:
-                if action.action == "run_job_template":
+                if action.action in [
```
mkanoor, line 237: @akira6592 Can you add a CONTROLLER_ACTIONS constant similar to this https://github.com/ansible/ansible-rulebook/blob/0da53a65379f0ae1128227c4419f9576de83c1c9/ansible_rulebook/app.py#L59 and use it in both the places.

akira6592: @mkanoor Thank you for suggestion. I fixed it.

mkanoor (approved):

hsong-rh (approved):

PR #575, opened 2023-08-25
AAP-15000: Changes to container image to reduce overall size

ttuffin:
Closes [AAP-15000](https://issues.redhat.com/browse/AAP-15000).

The main changes are switching to Centos stream image and installing ansible-core instead of the full ansible package.
Image size is reduced from ~2gb to ~500mb.
Tests in the eda-qa test suite are passing with this image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,22 @@
-FROM registry.access.redhat.com/ubi8/python-39
+FROM quay.io/centos/centos:stream9-development
 
 ARG USER_ID=${USER_ID:-1001}
+ARG APP_DIR=${APP_DIR:-/app}
 ARG DEVEL_COLLECTION_LIBRARY=0
 ARG DEVEL_COLLECTION_REPO=git+https://github.com/ansible/event-driven-ansible.git
-WORKDIR $HOME
 
 USER 0
-RUN dnf install -y java-17-openjdk-devel
+RUN useradd -u $USER_ID -d $APP_DIR appuser
+WORKDIR $APP_DIR
+COPY . $WORKDIR
+RUN chown -R $USER_ID $APP_DIR
+RUN dnf install -y java-17-openjdk-devel python3-pip
+
+USER $USER_ID
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+ENV PATH="${PATH}:$APP_DIR/.local/bin"
 RUN pip install -U pip \
-    && pip install ansible \
+    && pip install ansible-core \
     ansible-runner \
     jmespath \
     asyncio \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,14 +1,21 @@
-FROM registry.access.redhat.com/ubi8/python-39
+FROM quay.io/centos/centos:stream9-development
```
Alex-Izquierdo, line 1: Why not use the regular `quay.io/centos/centos:stream9` base image?

Dockerfile
```diff
@@ -1,14 +1,21 @@
...
+COPY . $WORKDIR
+RUN chown -R $USER_ID $APP_DIR
+RUN dnf install -y java-17-openjdk-devel python3-pip
+
+USER $USER_ID
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+ENV PATH="${PATH}:$APP_DIR/.local/bin"
 RUN pip install -U pip \
-    && pip install ansible \
+    && pip install ansible-core \
```
Alex-Izquierdo, line 18: IMO I prefer to use `ansible` package to have a full ansible experience.

Dockerfile
ttuffin: Using the base image instead of the development image doubles the size from ~500mb to ~1gb.

Dockerfile
ttuffin: I agree the `ansible` package is usually preferable, but in this case it adds an extra ~500mb due to all of the collections that get installed alongside it.

Dockerfile
Alex-Izquierdo: weird, I would thought that a "development" image would have always more things than a regular one.

Dockerfile
```diff
@@ -1,14 +1,21 @@
...
 ARG USER_ID=${USER_ID:-1001}
+ARG APP_DIR=${APP_DIR:-/app}
 ARG DEVEL_COLLECTION_LIBRARY=0
-WORKDIR $HOME
 
 USER 0
-RUN dnf install -y java-17-openjdk-devel
+RUN useradd -u $USER_ID -d $APP_DIR appuser
+WORKDIR $APP_DIR
+COPY . $WORKDIR
```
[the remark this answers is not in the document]
Alex-Izquierdo: Oh I didn't notice that we are not using a requirements file for the app dependencies. Create a new file for it would need more changes out of the scope of your PR and I want to migrate the project to poetry at some point. So we can do this improvement later.

Alex-Izquierdo (commented):
LGTM. One way to get more confidence could be run the e2e test suite against this image.

ttuffin:
> LGTM. One way to get more confidence could be run the e2e test suite against this image.

Tests are passing :+1:

Alex-Izquierdo:
Lets put on hold this PR until implement a CI workflow to test the image.

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

Dostonbek1 (approved):

PR #576, opened 2023-08-30
chore: add commitlint

mkanoor:
The commit messages have to conform to the conventional commit standard.
https://www.conventionalcommits.org/en/v1.0.0/

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -53,6 +53,7 @@ Before you submit a pull request, check that it meets these guidelines:
    your new functionality into a function with a docstring, and add the
    feature to the list in README.rst.
 3. The pull request should work for Python 3.9
+4. Commit messages must conform to `conventional commit <https://www.conventionalcommits.org/en/v1.0.0/>`__
 
 
 Submit Feedback
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -14,3 +14,5 @@ coverage
 
 # For building DEs
 ansible-builder
+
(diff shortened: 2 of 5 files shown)


docs/contributing.rst
```diff
@@ -53,6 +53,7 @@ Before you submit a pull request, check that it meets these guidelines:
    your new functionality into a function with a docstring, and add the
    feature to the list in README.rst.
 3. The pull request should work for Python 3.9
+4. Commit messages should conform to `conventional commit <https://www.conventionalcommits.org/en/v1.0.0/>`__
```
Alex-Izquierdo, line 56: We are adding the check in the PR so it is actually mandatory. 
```suggestion
4. Commit messages must conform to `conventional commit <https://www.conventionalcommits.org/en/v1.0.0/>`__
```

requirements_dev.txt
```diff
@@ -14,3 +14,5 @@ coverage
 
 # For building DEs
 ansible-builder
+
+pre-commit
```
Alex-Izquierdo, line 18: If we add pre-commit to the requirements, [this section of the docs](https://github.com/ansible/ansible-rulebook/blob/main/docs/development_environment.rst#git-pre-commit-hooks-optional) needs to be updated as well because we no longer need to install it manually.

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

benthomasson (approved):

PR #579, opened 2023-09-12
chore: add .DS_Store to .gitignore

mkanoor:
On Mac we were not ignoring the .DS_Store

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,7 @@ ENV/
 
 # awx provision
 tests/e2e/utils/awx/artifacts
+
+# Mac
+.DS_Store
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (approved):

Alex-Izquierdo:
It is fine, but FYI, you can configure a global exception in your local environment (I assume you want to exclude .ds_store files in all your projects)
```
git config --global core.excludesFile [path-to-global-ignore.file]
```

Same for the file `.git/info/exclude` if you want to do it only for a specific project.

ptoscano (approved):

PR #587, opened 2023-09-26
chore: ruleset.name is required parameter

mkanoor:
Every ruleset should have a name so we can post events/facts between them.

### DIFF
--- a/ansible_rulebook/schema/ruleset_schema.json
+++ b/ansible_rulebook/schema/ruleset_schema.json
@@ -56,6 +56,7 @@
                 }
             },
             "required": [
+                "name",
                 "hosts",
                 "sources",
                 "rules"


Alex-Izquierdo (approved):

ttuffin (approved):

Alex-Izquierdo:
I just noticed that the release notes should be updated because this is would be a breaking change.

mkanoor:
@Alex-Izquierdo It was always a required parameter, we just didn't have the schema reflecting that.
We have tests https://github.com/ansible/ansible-rulebook/blob/f862c1a0e7f046e47da4d2faf00862e1bc92e94e/tests/test_rules.py#L159
It was also one of the requirements for set_facts and post_events since they need a ruleset name to use.

PR #588, opened 2023-09-27
test: e2e test for run_module

mkanoor:
https://issues.redhat.com/browse/AAP-14735
Added e2e test for run_module

### DIFF
--- a/tests/e2e/test_run_module_output.py
+++ b/tests/e2e/test_run_module_output.py
@@ -1,0 +1,108 @@
+"""
+Module with tests for websockets
+"""
+import asyncio
+import logging
+from functools import partial
+
+import dpath
+import pytest
+import websockets.server as ws_server
+
+from . import utils
+
+LOGGER = logging.getLogger(__name__)
+DEFAULT_TIMEOUT = 15
+
+
+@pytest.mark.e2e
+@pytest.mark.asyncio
+async def test_run_module_output():
+    """
+    Verify that ansible-rulebook can handle output of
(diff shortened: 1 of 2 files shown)


mkanoor:
E2E test for PR https://github.com/ansible/ansible-rulebook/pull/564

tests/e2e/test_run_module_output.py
```diff
@@ -0,0 +1,108 @@
...
+            assert stats["ruleSetName"] == "29 run module"
+            assert stats["numberOfRules"] == 2
+            assert stats["numberOfDisabledRules"] == 0
+            assert data["activation_id"] == proc_id
+
+    assert stats["rulesTriggered"] == 2
+    assert stats["eventsProcessed"] == 6
+    assert stats["eventsMatched"] == 2
+
+    assert session_stats_counter >= 2
```
Alex-Izquierdo, line 107: why this is not exact?

mkanoor: @Alex-Izquierdo because the number of session stats sent will depend on how long the process runs.

* There is a heartbeat at the beginning
* There is a heartbeat every 2 seconds as configured when the ansible-rulebook is called
* There is a heartbeat when every rule fires.
* There is a heartbeat at the end when the rulebook is about to end

We don't know for how long the process is going to run so its not possible to get the exact count.

Alex-Izquierdo (approved):

ttuffin (approved):

PR #595, opened 2023-10-05
ci: update CI pipeline image build tests

ttuffin:
- Move build-and-test-image to actions
- Run image test on PR and push to main
- Run tests against amd64 arch (arm64 CI takes way too long (30+ minutes) due to poor emulator performance)


.github/workflows/ci.yml
```diff
@@ -79,3 +79,14 @@ jobs:
           flags: "unittests-${{ matrix.python-version }}"
           name: codecov-umbrella
           verbose: true
+
+  build-and-test-image:
+    if: github.repository == 'ansible/ansible-rulebook'
```
Alex-Izquierdo, line 84: I think forks might want to run this as well, any contribution can break the image.

Alex-Izquierdo: yes, but under that logic we would prevent to run everything in forks. The purpose is be able to run the workflows in the fork prior the PR. IMO, only the workflows that depends on the repo (eg due to secrets) should be disabled. 
If some fork doesn't want to run a specific workflow they can disable it there.

Alex-Izquierdo (approved):

.github/workflows/ci.yml
ttuffin: Fixed.

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #596, opened 2023-10-05
ci: add more configuration for commitlint

Alex-Izquierdo:
- Disable subject-case: we might use acronyms or jira tickets that will contain uppercase letters. 
- Fix wrong configuration for body length
- Add a dependency to pre-commit
- Extends docs for commitlint with pre-commit

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -4,7 +4,6 @@ module.exports = {
     extends: ['@commitlint/config-conventional'],
     rules: {
         'body-max-line-length': [2, 'always', Infinity],
-        'subject-case': [0, 'always', 'lower-case'],
         'body-case': [0, 'always', 'lower-case'],
     },
 };
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -4,6 +4,5 @@ module.exports = {
     extends: ['@commitlint/config-conventional'],
     rules: {
         'body-max-line-length': [2, 'always', Infinity],
-        'body-case': [0, 'always', 'lower-case'],
     },
 };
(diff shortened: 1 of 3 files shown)


ttuffin (dismissed):

mkanoor (approved):

mkanoor:
@Alex-Izquierdo I think for the subject I think we should still use lower case. Since when they build the release notes the non title case for the first character makes sense. You can have AAP-XXXX in the subject but it should be worded differently
```
echo "fix: allow for TESTS" | commitlint
```
```
echo "fix: resolved issue in AAP-94588" | commitlint
```

Its one that they don't want the ticket to start with Uppercase letters
```
echo "fix: AAP-94588" | commitlint
```
This will fail with subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]

commitlint.config.js
```diff
@@ -3,6 +3,7 @@
 module.exports = {
     extends: ['@commitlint/config-conventional'],
     rules: {
-        'body-max-line-length': [2, 'always', '*'],
+        'body-max-line-length': [2, 'always', Infinity],
+        'body-case': [0, 'always', 'lower-case'],
```
mkanoor, line 7: @Alex-Izquierdo Not sure if this makes sense, the body is just a paragraph and it will have collection of sentences. Why do we have to make sentences in lower case?

mkanoor (approved):

commitlint.config.js
Alex-Izquierdo: I disabled the rule because according to the rule we are enforcing the lowercase in the body which looked to me excessive.

Alex-Izquierdo: removed

ttuffin (approved):

PR #604, opened 2023-10-25
fix: [AAP-12264] output job url at run completion

jshimkus-rh:


Alex-Izquierdo (dismissed):

ansible_rulebook/action/run_job_template.py
```diff
@@ -66,6 +66,9 @@ async def __call__(self):
         )
         await self._job_start_event()
         await self._run()
+        url = self._controller_job_url()
+        if url:
+            print(f"job results url: {url}")
```
mkanoor, line 71: @jshimkus-rh It might make more sense to 
do a 
logger.info(a_log)
here 
https://github.com/ansible/ansible-rulebook/blob/c6830d0c33069ba1a17814da10f54f32db2e82f1/ansible_rulebook/action/run_job_template.py#L119

That's equivalent to the data being sent up to the server for reporting purposes.

ansible_rulebook/action/run_workflow_template.py
```diff
@@ -66,6 +66,9 @@ async def __call__(self):
         )
         await self._job_start_event()
         await self._run()
+        url = self._controller_job_url()
+        if url:
+            print(f"job results url: {url}")
```
mkanoor, line 71: @jshimkus-rh It might make more sense to 
do a 
logger.info(a_log)
here 
https://github.com/ansible/ansible-rulebook/blob/c6830d0c33069ba1a17814da10f54f32db2e82f1/ansible_rulebook/action/run_workflow_template.py#L125

It reports on errors as well as success and is the same data that is sent to the server for logging purpose.

jshimkus-rh: @mkanoor @Alex-Izquierdo 
Using logger was my initial intent. Alejandro wanted the url unconditionally.  Using --verbose produces a lot of other output that otherwise makes locating just the url more difficult. 

We could do both and conditionally skip the print if --verbose is set but that requires tracking any format changes from logger or accepting two different output formats (ugh). 

Anyhow, there was a discussion at dev sync about standardizing output in general, similarly to ansible-playbook but that requires some investigation. I would expect this change to be affected by that so something simple for now seems appropriate.

mkanoor: @jshimkus-rh We also need to track errors, if the controller rejects the request we need to log it as an error that is currently missing. We report this to the backend but we don't have anything when running the standalone CLI.

jshimkus-rh: Not certain I'm following you.  Nothing has changed with previously existing logging so if there's a rejection at that level it should still be getting logged.  Perhaps you mean that outputting the url should be conditional on the rejection?  That'd make sense.  

Or perhaps you mean that the preexisting code is missing a log message; that'd be orthogonal to this change.

mkanoor: The existing code is missing error messages, they don't get logged to the screen they are directly sent to the eda-server.

jshimkus-rh: It seems to me (based on the interpretation of the above that only logs messages should be generated)  that you and @Alex-Izquierdo need to find an agreement on what you can both live with.

I will re-mention that last Tuesday's dev sync had a brief discussion  on what upstream output should possibly be generated. While that is still an open question the presumption is that it would involve terminal output.

Alex-Izquierdo: IMO, certain information should be printed to the stdout by default because it aligns with user expectations. I understand that the log level (indicated by the -v flag) is intended to increase the amount of data for debugging purposes. However, as a user, I expect to see the results of some action in my rulebook without having to specify it explicitly, as this information belongs to the application's main purpose rather than just for debugging. This approach is also consistent with other actions like playbooks. If we display the result of the playbook using runner, we should also display the result of the playbook when using AWX, or at  least, provide the URL for the job result, as we currently do.

That being said, I don't intend to block this approach. If the consensus is different, that's fine to me.

Alex-Izquierdo: I agree that only the url is confusing, it should have more context about the rule/action it has triggered it, I mention it in a previous comment. 
About the silent mode, I feel it makes more sense to implement a `-q/--quiet` flag. It depends on what is the main usage of the ansible-rulebook, if a backend for eda-server or a standalone CLI program. To me is different driving decisions on the requirements of eda-server than build a cli program that is also used by eda-server.

Alex-Izquierdo: I'm not sure if we already have a well defined requirement on this matter. I am not against to use the logger if is the agreement, but we should clarify this point.

jshimkus-rh: My understanding of the jira issue is the desire to output the url when **manually** running ansible-rulebook, thus context would be implicit to the user.  But perhaps I have misunderstood.

Regardless, let's figure out what the correct approach should be.

benthomasson (dismissed):
Until we have a UX review for the text UI we should send this to the logger.

ansible_rulebook/action/run_job_template.py
```diff
@@ -113,6 +113,12 @@ async def _post_process(self) -> None:
             "url": self._controller_job_url(),
             "matching_events": self.helper.get_events(),
         }
+        a_log["job_results_url"] = (
```
mkanoor, line 116: @jshimkus-rh Wouldn't the a_log["url"] suffice here, it already has the job URL, we could just log the a_log or pick the a_log["url"] if we done have an error.

jshimkus-rh: Yes, it would.  I overlooked it.

mkanoor (approved):

PR #606, opened 2023-10-27
feat: support true or false for verify ssl options

Alex-Izquierdo:


jshimkus-rh (approved):

msmagnanijr (approved):

ansible_rulebook/job_template_runner.py
```diff
@@ -84,9 +84,9 @@ def _auth_headers(self) -> dict:
     @cached_property
     def _sslcontext(self) -> Union[bool, ssl.SSLContext]:
         if self.host.startswith("https"):
-            if self.verify_ssl.lower() == "yes":
+            if self.verify_ssl.lower() in ["yes", "true"]:
```
mkanoor, line 87: @Alex-Izquierdo should it mimic the https://docs.python.org/3.9/distutils/apiref.html#distutils.util.strtobool which is being deprecated in 3.12, we could implement and function and put it in our utils.

mkanoor (approved):

ttuffin (approved):

PR #616, opened 2023-11-06
feat: added pg_notify

mkanoor:
Added an action to send notifications via postgres LISTEN/NOTIFY

This feature requires an updated ansible.eda collection with the pg_listener source plugin

Using this branch you can create an image with the new collection and push it to your private quay.io repo

* podman login quay.io
* podman build --no-cache --build-arg DEVEL_COLLECTION_LIBRARY=1 --build-arg DEVEL_COLLECTION_REPO=git+https://github.com/mkanoor/event-driven-ansible.git,pg_listener -t quay.io/<your_userid>/ansible-rulebook:latest .
* podman push quay.io/<your_user_id>/ansible-rulebook:latest

This will ensure that your image has the latest collection which can use the pg_listener source plugin in a rulebook.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN useradd -u $USER_ID -d $APP_DIR appuser
 WORKDIR $APP_DIR
 COPY . $WORKDIR
 RUN chown -R $USER_ID $APP_DIR
-RUN dnf install -y java-17-openjdk-devel python3-pip
+RUN dnf install -y java-17-openjdk-devel python3-pip postgresql-devel gcc python3-devel
 
 RUN bash -c "if [ $DEVEL_COLLECTION_LIBRARY -ne 0 ]; then \
     dnf install -y git; fi"
--- a/ansible_rulebook/action/pg_notify.py
+++ b/ansible_rulebook/action/pg_notify.py
@@ -1,0 +1,111 @@
+#  Copyright 2023 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import json
+import logging
+import uuid
+
+import xxhash
+from psycopg import AsyncClientCursor, AsyncConnection, OperationalError
+
+from .control import Control
+from .helper import FAILED_STATUS, Helper
+from .metadata import Metadata
+
+logger = logging.getLogger(__name__)
+
+MAX_MESSAGE_LENGTH = 7 * 1024
+MESSAGE_CHUNKED_UUID = "_message_chunked_uuid"
+MESSAGE_CHUNK_COUNT = "_message_chunk_count"
+MESSAGE_CHUNK_SEQUENCE = "_message_chunk_sequence"
+MESSAGE_CHUNK = "_chunk"
+MESSAGE_LENGTH = "_message_length"
+MESSAGE_XX_HASH = "_message_xx_hash"
+
+
+class PGNotify:
+    """The PGNotify action sends an event to a PG Pub Sub Channel
+    Needs
+    dsn https://www.postgresql.org/docs/current/libpq-connect.html
+    #LIBPQ-CONNSTRING-KEYWORD-VALUE
+    channel the channel name to send the notifies
+    event
+    """
+
+    def __init__(self, metadata: Metadata, control: Control, **action_args):
+        self.helper = Helper(metadata, control, "pg_notify")
+        self.action_args = action_args
+
+    async def __call__(self):
+        if not self.action_args["event"]:
+            return
+
+        try:
+            async with await AsyncConnection.connect(
+                conninfo=self.action_args["dsn"],
+                autocommit=True,
+            ) as conn:
+                async with AsyncClientCursor(connection=conn) as cursor:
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN useradd -u $USER_ID -d $APP_DIR appuser
 WORKDIR $APP_DIR
 COPY . $WORKDIR
 RUN chown -R $USER_ID $APP_DIR
-RUN dnf install -y java-17-openjdk-devel python3-pip
+RUN dnf install -y java-17-openjdk-devel python3-pip git
```
benthomasson, line 13: Why do we need git?

mkanoor: @benthomasson Its needed to build an image in development mode while we wait for the collection to be result. We can build an image by setting the branch to use 
ARG DEVEL_COLLECTION_REPO=git+https://github.com/ansible/event-driven-ansible.git

mkanoor: e.g.
```
podman build --no-cache --build-arg DEVEL_COLLECTION_LIBRARY=1 --build-arg DEVEL_COLLECTION_REPO=git+https://github.com/mkanoor/event-driven-ansible.git,pg_listener -t quay.io/mkanoor/ansible-rulebook:latest .
```

Alex-Izquierdo: This would be addressed in https://github.com/ansible/ansible-rulebook/pull/637

Alex-Izquierdo, line 13: https://github.com/ansible/ansible-rulebook/pull/637 is already merged, we don't need this.

Alex-Izquierdo (dismissed):
LGTM. As a suggestion for an enhancement, the chunk logic can be encapsulated in a separate function, and it can have its own unit tests to ensure that the chunk logic works as expected.

ansible_rulebook/action/pg_notify.py
```diff
@@ -0,0 +1,108 @@
...
+            ) as conn:
+                async with AsyncClientCursor(connection=conn) as cursor:
+                    if self.action_args.get("remove_meta", False):
+                        event = self.action_args["event"].copy()
+                        if "meta" in event:
+                            event.pop("meta")
+                    else:
+                        event = self.action_args["event"]
+
+                    payload = json.dumps(event)
```
Alex-Izquierdo, line 64: Should we check if the event has real content to avoid a potential empty transaction?

Alex-Izquierdo (approved):
LGTM

Alex-Izquierdo (approved):

setup.cfg
```diff
@@ -32,7 +32,9 @@ install_requires =
 	ansible-runner
 	websockets
 	drools_jpy == 0.3.8
-    watchdog
+	watchdog
+	psycopg[c]
```
benthomasson, line 36: I am a bit concerned that we are going to run into issues with lots of dependencies in ansible-rulebook.     This dependency is fine, but any more actions with dependencies should go into action plugins.

benthomasson: We already have problems with ansible.eda dependencies on different platforms.   We might need platform specific dependencies.

benthomasson: psycopg[c] requires a compiler for installation.

Alex-Izquierdo: The dockerfile already includes the compiler and for downstream the change is also already being coordinated.

ansible_rulebook/action/pg_notify.py
```diff
@@ -0,0 +1,109 @@
...
+        logger.debug("Message length exceeds, will chunk")
+        message_uuid = str(uuid.uuid4())
+        number_of_chunks = int(message_length / MAX_MESSAGE_LENGTH) + 1
+        chunked = {
+            MESSAGE_CHUNKED_UUID: message_uuid,
+            MESSAGE_CHUNK_COUNT: number_of_chunks,
+            MESSAGE_LENGTH: message_length,
+            MESSAGE_XX_HASH: xx_hash,
+        }
+        logger.debug(f"Chunk info {message_uuid}")
```
benthomasson, line 99: Use the old style % string substitution in logging because it avoids building the strings if that level of logging is turned off.   https://docs.python.org/3/howto/logging.html#optimization

benthomasson (changes requested):

setup.cfg
Alex-Izquierdo: As it is now everything is in the image, but there is a parallel effort to improve the dockerfile with multistage builds:
https://github.com/ansible/ansible-rulebook/pull/639
So, after this change, that PR can be update to just use the compiler in the build stage and not ship it in the final one.

ansible_rulebook/action/pg_notify.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, metadata: Metadata, control: Control, **action_args):
+        self.helper = Helper(metadata, control, "pg_notify")
+        self.action_args = action_args
+
+    async def __call__(self):
+        if not self.action_args["event"]:
+            return
+
+        try:
+            async with await AsyncConnection.connect(
```
benthomasson, line 55: It might be beneficial to cache the connection between actions if they have the same connection info.  This could be stored as a dictionary on the module or class.

benthomasson: We should measure the performance with and without this caching to see if there is a performance improvement.

ansible_rulebook/action/pg_notify.py
mkanoor: Updated

ansible_rulebook/action/pg_notify.py
mkanoor: There is a separate package called psycopg[pool] which has an AsyncConnectionPool https://www.psycopg.org/psycopg3/docs/advanced/pool.html#connection-pools
We can test using the pooled connection in a different PR
```
async with AsyncConnectionPool(...) as pool:
    async with pool.connection() as conn:
        await conn.execute("SELECT something FROM somewhere ...")

        with conn.cursor() as cur:
            await cur.execute("SELECT something else...")
```

benthomasson (approved):

PR #620, opened 2023-11-22
feat: support for kafka_notify

mkanoor:
Added a new action called kafka_notify to contrast with pg_notify This action would allow an event to be posted to a Kafka topic for other consumers.

### DIFF
--- a/ansible_rulebook/action/kafka_notify.py
+++ b/ansible_rulebook/action/kafka_notify.py
@@ -1,0 +1,63 @@
+#  Copyright 2023 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


ansible_rulebook/action/kafka_notify.py
```diff
@@ -0,0 +1,42 @@
...
+        self.helper = Helper(metadata, control, "kafka_notify")
+        self.action_args = action_args
+
+    async def __call__(self):
+        producer = KafkaProducer(**self.action_args["connection"])
+        producer.send(
+            self.action_args["topic"],
+            json.dumps(self.action_args["event"]).encode("utf-8"),
+        )
+        producer.close()
```
Alex-Izquierdo, line 40: `send` op can fail. We should ensure that the connection is closed. It can be in a try/finally block or also you can implement it as a context manager:
https://docs.python.org/3/library/contextlib.html#contextlib.closing

Alex-Izquierdo (commented):
Changelog and docs should be updated as well, don't? Specially the "connection" parameters. I think also we want to support authentication and ssl as we noticed with the kafka source plugin.

PR #625, opened 2023-12-01
feat: implement banner output

jshimkus-rh:
Adds banner-type output to ansible-rulebook in the manner previously demonstrated.

There's an open question as to how we want to handle outputting  events in rule_set_runner.py.
The previous code disabled args.print_events if verbosity was 2+ as the logging would output the event.
We shouldn't have two different forms of output for the same data and I'd opt for only the args.print_event form in all cases for consistency unless there's specific, great enough value in using the logger when verbosity is 2+. 
At present, this code forcibly enables args.print_events when verbosity is 2+ so both forms are output.


ansible_rulebook/terminal.py
```diff
@@ -0,0 +1,191 @@
...
+                    cls.__singleton = super().__call__(*args, **kwargs)
+        return cls.__singleton
+
+
+###############################################################################
+class Singleton(metaclass=_Singleton):
+    pass
+
+
+###############################################################################
```
Alex-Izquierdo, line 46: Why all these lines?

ansible_rulebook/terminal.py
```diff
@@ -0,0 +1,191 @@
...
+class DisplayBannerIncompleteError(DisplayBannerError):
+    pass
+
+
+###############################################################################
+class Display(Singleton):
+    ###########################################################################
+    @classmethod
+    def instance(cls, level: int = None):
```
Alex-Izquierdo, line 60: If type is int, default should not be None.

ansible_rulebook/terminal.py
```diff
@@ -0,0 +1,191 @@
...
+import pprint
+import re
+import threading
+import typing
+
+logger = logging.getLogger(__name__)
+
+
+###############################################################################
+class _Singleton(type):
```
Alex-Izquierdo, line 27: If we are going to provide a singleton metaclass would not be better to put it in a different place?

ansible_rulebook/action/post_event.py
```diff
@@ -34,8 +36,12 @@ class PostEvent:
     def __init__(self, metadata: Metadata, control: Control, **action_args):
         self.helper = Helper(metadata, control, "post_event")
         self.action_args = action_args
+        self.display = terminal.Display()
 
     async def __call__(self):
+        self.display.banner(
```
Alex-Izquierdo, line 42: Why display a banner on the post_event action?

ansible_rulebook/terminal.py
```diff
@@ -0,0 +1,191 @@
...
+    ###########################################################################
+    def _format_banner(self, banner: str) -> str:
+        if len(banner) > 0:
+            banner = self._prefix_time(banner)
+        banner = self._rule_embed(banner)
+        return banner
+
+    ###########################################################################
+    def _prefix_time(self, content: str) -> str:
+        return f"{datetime.datetime.now().isoformat(' ')} {content}"
```
Alex-Izquierdo, line 167: It would be convenient to use a timezone-aware object: 
https://docs.astral.sh/ruff/rules/call-datetime-now-without-tzinfo/

ansible_rulebook/terminal.py
```diff
@@ -0,0 +1,191 @@
...
+    pass
+
+
+###############################################################################
+class DisplayBannerIncompleteError(DisplayBannerError):
+    pass
+
+
+###############################################################################
+class Display(Singleton):
```
Alex-Izquierdo, line 57: I would say that this object deserves tests.

ansible_rulebook/terminal.py
```diff
@@ -0,0 +1,191 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+###############################################################################
+class _Singleton(type):
+    def __init__(cls, name, bases, dct, **kwargs) -> None:
+        super().__init__(name, bases, dct, **kwargs)
+        cls.__singleton = None
+        cls.__lock = threading.RLock()
```
Alex-Izquierdo, line 31: What is the need for a lock?

I think this might be a potential issue. Ansible-rulebook is essentially an asyncio-based application. Asyncio is designed for cooperative multitasking and doesn't mix well with traditional thread synchronization primitives like locks. When you use a blocking lock in an asyncio application, you risk blocking the entire event loop, which can lead to performance issues or deadlocks.

Instead, I propose considering `asyncio.lock()` https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event if locking is really necessary. Alternatively, we could reconsider the whole singleton pattern in favor of dependency injection or using shared instances. These alternatives fit more naturally with asyncio's paradigms.

ansible_rulebook/terminal.py
jshimkus-rh: They provide easily (human) recognized delineation of methods, classes, etc.
I expected "controversy." 😄 But I'm not wedded to them, was more curious as to what people might think.

ansible_rulebook/terminal.py
jshimkus-rh: With only a single consumer of it there's no specific need.  When/if there is more than one consumer that would make sense.

ansible_rulebook/terminal.py
Alex-Izquierdo: I mean the typehint is incorrect. If the type is int, the default value should be int. Otherwise it should be typing.optional[int]

ansible_rulebook/terminal.py
Alex-Izquierdo: Sure, this is a style concern. ¯\_(ツ)_/¯
IMO is noise and is readable and clear enough without it and aligns with the rest of the code which doesn't have such a marks. 
I don't have my life in it either

ansible_rulebook/terminal.py
jshimkus-rh: These are terminal-level outputs not log entries.  Using a timezone would produce output with timestamps inconsistent with other ansible-rulebook output terminal output.

ansible_rulebook/terminal.py
jshimkus-rh: The lock is only present to synchronize the creation of the singleton and is held only long enough to do so.  Once the singleton is created the lock is never again utilized.  It should not be an issue.
But using asyncio.lock() should also suffice.

jshimkus-rh: Using asyncio, with its imposed requirements on methods, doesn't play well with the metaclassing (object hierarchy creation) and use of the descendant class.  Given the limited nature of the lock use I don't see a need for changes here.

ansible_rulebook/rule_set_runner.py
```diff
@@ -203,12 +205,18 @@ async def _drain_source_queue(self):
...
+                # If print_events is specified adjust the level to the
+                # display's current level to guarantee output.
                 if self.parsed_args and self.parsed_args.print_events:
-                    PrettyPrinter(indent=4).pprint(data)
+                    level = self.display.level
 
-                logger.debug(
-                    "Ruleset: %s, received event: %s ", self.name, str(data)
+                self.display.banner(
+                    "rule set runner", data, pretty=True, level=level
```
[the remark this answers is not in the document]
jshimkus-rh: Want to keep the banner headline generic, but changed to "received event".
Added ruleset name to body of the output.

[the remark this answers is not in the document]
Alex-Izquierdo: Yeah, quite better now IMO. 
```
** 2023-12-15 18:34:27.729557 [received event] ********************************************************************************************************************************************************************
Ruleset: Ruleset 1
Event:
{'meta': {'received_at': '2023-12-15T17:34:27.728838Z',
          'source': {'name': 'ansible.eda.generic',
                     'type': 'ansible.eda.generic'},
          'uuid': '68b4e67f-104b-4515-aaa2-d5c67b2db087'},
 'payload': {'meta': {'hosts': ['localhost', 'localhost2']}, 'name': 'alex'}}

```

Alex-Izquierdo (approved):

benthomasson:
I like this.  I think we should add a configuration option to send the logging to another stream besides standard out so we can see the output of this alone.

benthomasson (approved):
LGTM

PR #628, opened 2023-12-07
feat: add websocket authentication using jwt token

bzwei:
The cli expects to receive websocket-access-token, websocket-refresh-token, and websocket-token-address.

It does not send the authentication header if above arguments are not provided, so it works with the old eda-server that does not authenticate incomming websocket connecitons.

Fixes AAP-17776: ansible-rulebook uses token for authentication

### DIFF
--- a/ansible_rulebook/cli.py
+++ b/ansible_rulebook/cli.py
@@ -97,16 +97,32 @@ def get_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument(
         "-W",
-        "--websocket-address",
         "--websocket-url",
+        "--websocket-address",
         help="Connect the event log to a websocket",
+        default=os.environ.get("EDA_WEBSOCKET_URL", ""),
     )
     parser.add_argument(
         "--websocket-ssl-verify",
         help="How to verify SSL when connecting to the "
         "websocket: (yes|true) | (no|false) | <path to a CA bundle>, "
         "default to yes for wss connection.",
-        default="yes",
+        default=os.environ.get("EDA_WEBSOCKET_SSL_VERIFY", "yes"),
+    )
+    parser.add_argument(
+        "--websocket-access-token",
+        help="Token used to autheticate the websocket connection.",
+        default=os.environ.get("EDA_WEBSOCKET_ACCESS_TOKEN", ""),
+    )
+    parser.add_argument(
+        "--websocket-refresh-token",
+        help="Token used to renew a websocket access token.",
+        default=os.environ.get("EDA_WEBSOCKET_REFRESH_TOKEN", ""),
+    )
+    parser.add_argument(
+        "--websocket-token-url",
+        help="Url to renew websocket access token.",
+        default=os.environ.get("EDA_WEBSOCKET_TOKEN_URL", ""),
     )
     parser.add_argument("--id", help="Identifier")
     parser.add_argument(
@@ -215,10 +231,8 @@ def get_version() -> str:
 
 
 def validate_args(args: argparse.Namespace) -> None:
-    if args.worker and (not args.id or not args.websocket_address):
-        raise ValueError(
-            "Worker mode needs an id and websocket address specfied"
-        )
+    if args.worker and (not args.id or not args.websocket_url):
+        raise ValueError("Worker mode needs an id and websocket url specfied")
     if not args.worker and not args.rulebook:
         raise ValueError("Rulebook must be specified in non worker mode")
 
@@ -255,6 +269,11 @@ def update_settings(args: argparse.Namespace) -> None:
         settings.default_execution_strategy = args.execution_strategy
 
     settings.print_events = args.print_events
+    settings.websocket_url = args.websocket_url
+    settings.websocket_ssl_verify = args.websocket_ssl_verify
+    settings.websocket_token_url = args.websocket_token_url
+    settings.websocket_access_token = args.websocket_access_token
+    settings.websocket_refresh_token = args.websocket_refresh_token
 
 
 def main(args: List[str] = None) -> int:
--- a/ansible_rulebook/token.py
+++ b/ansible_rulebook/token.py
@@ -1,0 +1,50 @@
+#  Copyright 2024 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+import ssl
+import typing as tp
+
+import aiohttp
+
+from ansible_rulebook.conf import settings
+from ansible_rulebook.exception import TokenNotFound
+
+logger = logging.getLogger(__name__)
+
+
+async def renew_token() -> str:
+    logger.info("Renew websocket token from %s", settings.websocket_token_url)
+    async with aiohttp.ClientSession() as session:
+        async with session.post(
+            settings.websocket_token_url,
+            data={"refresh": settings.websocket_refresh_token},
+            ssl_context=_sslcontext(),
+        ) as resp:
+            data = await resp.json()
+            if "access" not in data:
+                logger.error(f"Failed to renew token. Error: {str(data)}")
+                raise TokenNotFound("Response does not contain access token")
+            return data["access"]
+
+
+def _sslcontext() -> tp.Optional[ssl.SSLContext]:
+    if settings.websocket_token_url.startswith("https"):
+        ssl_verify = settings.websocket_ssl_verify.lower()
+        if ssl_verify in ["yes", "true"]:
+            return ssl.create_default_context()
+        if ssl_verify in ["no", "false"]:
+            return ssl._create_unverified_context()
+        return ssl.create_default_context(cafile=ssl_verify)
+    return None
--- a/ansible_rulebook/websocket.py
+++ b/ansible_rulebook/websocket.py
@@ -12,126 +12,216 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import asyncio
 import base64
 import json
 import logging
 import os
+import random
 import ssl
 import tempfile
 import typing as tp
-from asyncio.exceptions import CancelledError
+from dataclasses import dataclass, field
 
 import websockets
 import yaml
+from websockets.client import WebSocketClientProtocol
 
 from ansible_rulebook import rules_parser as rules_parser
 from ansible_rulebook.common import StartupArgs
+from ansible_rulebook.conf import settings
+from ansible_rulebook.exception import ShutdownException
+from ansible_rulebook.token import renew_token
 
 logger = logging.getLogger(__name__)
 
 
-async def request_workload(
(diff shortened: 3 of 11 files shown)


ansible_rulebook/websocket.py
```diff
@@ -12,123 +12,254 @@
...
+async def request_workload(
+    activation_id: str,
+    websocket_address: str,
+    websocket_ssl_verify: str,
+    websocket_token_address: str,
+    websocket_access_token: str,
+    websocket_refresh_token: str,
+) -> StartupArgs:
+    return await _connect_websocket(
+        websocket_address=websocket_address,
```
mkanoor, line 154: @bzwei would it make sense to move all these to settings https://github.com/ansible/ansible-rulebook/blob/8743f53a1a7bf3eb24dfbc4cbe03c1cd1fad66da/ansible_rulebook/conf.py#L23 and use it from there we don't have to keep passing it around to every function, they are part of the settings and constant once the process starts, the only thing changing would be the access_token.

ansible_rulebook/cli.py
```diff
@@ -105,6 +105,19 @@ def get_parser() -> argparse.ArgumentParser:
...
+        help="Token used to autheticate the websocket connection.",
+    )
+    parser.add_argument(
+        "--websocket-refresh-token",
+        help="Token used to renew a websocket access token.",
+    )
+    parser.add_argument(
+        "--websocket-token-address",
+        "--websocket-token-url",
+        help="Url to renew websocket access token.",
```
mkanoor, line 119: @bzwei can we use environment variables here, would provide an easy way to test in the current setup https://github.com/ansible/ansible-rulebook/blob/8743f53a1a7bf3eb24dfbc4cbe03c1cd1fad66da/ansible_rulebook/cli.py#L125

ansible_rulebook/websocket.py
mkanoor: @bzwei We were planning on getting rid of the StartupArgs and just use settings every where and start using dynaconf.

ansible_rulebook/websocket.py
```diff
@@ -12,123 +12,254 @@
...
 from ansible_rulebook.common import StartupArgs
+from ansible_rulebook.exception import ShutdownException
 
 logger = logging.getLogger(__name__)
 
 
-async def request_workload(
-    activation_id: str, websocket_address: str, websocket_ssl_verify: str
-) -> StartupArgs:
+async def _renew_token(
```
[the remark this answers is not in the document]
mkanoor: If this goes into a separate function we can run an async task that can sleep and fetch the token just before it expires and updates the access_token in settings.

ansible_rulebook/websocket.py
bzwei: I just amend the existing cli arguments for websocket. Sure I can try to move all websocket options to settings.

ansible_rulebook/websocket.py
```diff
@@ -12,126 +12,208 @@
...
-                        os.close(project_data_fh)
-                        logger.debug("wrote %s", response.project_data_file)
-                if data.get("type") == "Rulebook":
-                    response.rulesets = rules_parser.parse_rule_sets(
-                        yaml.safe_load(base64.b64decode(data.get("data")))
+        except asyncio.CancelledError:  # pragma: no cover
+            raise
+        except ShutdownException:
+            break
+        except Exception as e:
```
mkanoor, line 75: @bzwei can we catch websockets.exceptions.InvalidStatusCode here instead of catching Exception

ansible_rulebook/websocket.py
```diff
@@ -12,126 +12,208 @@
...
+            if (
+                isinstance(e, websockets.exceptions.InvalidStatusCode)
```
mkanoor, line 77: @bzwei https://websockets.readthedocs.io/en/stable/reference/exceptions.html indicates this is legacy is there a newer alternative

ansible_rulebook/auth.py
```diff
@@ -0,0 +1,51 @@
+#  Copyright 2023 Red Hat, Inc.
```
mkanoor, line 1: @bzwei another suggestion for filename would be tokens or token helpers. auth has different meaning people would be expecting to see authentication classes or methods in that file.

ansible_rulebook/auth.py
```diff
@@ -0,0 +1,51 @@
...
+    async with aiohttp.ClientSession() as session:
+        async with session.post(
+            settings.websocket_token_address,
+            data={"refresh": settings.websocket_refresh_token},
+            ssl_context=_sslcontext(),
+        ) as resp:
+            data = await resp.json()
+            if "access" not in data:
+                logger.error(f"Failed to renew token. Error: {str(data)}")
+                return ""
```
mkanoor, line 39: @bzwei We should throw a well known exception here if we can't get the refresh token for any reason, e.g. the eda-server could be down for maintenance.

ansible_rulebook/cli.py
```diff
@@ -97,13 +97,30 @@ def get_parser() -> argparse.ArgumentParser:
...
+    )
+    parser.add_argument(
+        "--websocket-access-token",
+        help="Token used to autheticate the websocket connection.",
+        default=os.environ.get("EDA_WEBSOCKET_ACCESS_TOKEN", ""),
+    )
+    parser.add_argument(
+        "--websocket-refresh-token",
+        help="Token used to renew a websocket access token.",
+        default=os.environ.get("EDA_WEBSOCKET_REFRESS_TOKEN", ""),
```
mkanoor, line 117: @bzwei Typo is this REFRESH instead of REFRESS

ansible_rulebook/auth.py
mkanoor: @bzwei I think this file is making an assumption that the token refresh follows the simplejwt model and end points, if we change the backend auth method on the eda-server to something else the ansible-rulebook would have to change. Should we have a field that defines which Auth refresh token method is in play. Simple JWT, OAuth? So maybe this code could adjust itself.

ansible_rulebook/cli.py
```diff
@@ -97,13 +97,30 @@ def get_parser() -> argparse.ArgumentParser:
...
+    parser.add_argument(
+        "--websocket-refresh-token",
+        help="Token used to renew a websocket access token.",
+        default=os.environ.get("EDA_WEBSOCKET_REFRESH_TOKEN", ""),
+    )
+    parser.add_argument(
+        "--websocket-token-address",
+        "--websocket-token-url",
+        help="Url to renew websocket access token.",
+        default=os.environ.get("EDA_WEBSOCKET_TOKEN_ADDRESS", ""),
```
mkanoor, line 123: @bzwei should the be EDA_WEBSOCKET_TOKEN_URL, if it is using https do we need a ssl_verify option for that address?

ansible_rulebook/websocket.py
```diff
@@ -12,126 +12,213 @@
...
+            ) as websocket:
+                result = await handler(websocket, **kwargs)
+                if not retry:
+                    break
+        except asyncio.CancelledError:  # pragma: no cover
+            raise
+        except ShutdownException:
+            break
+        except websockets.exceptions.WebSocketException as e:
+            status403_lagcy = (
```
mkanoor, line 76: @bzwei is this status403_legacy instead of status403_lagcy?

ansible_rulebook/auth.py
bzwei: @mkanoor Yes this implementation makes the assumption of simple jwt or alike. The CLI takes corresponding arguments. If we switches to another authentication method, the CLI arguments may be different, and the parameters needed for fetching/renewing tokens can be very different too.

ansible_rulebook/websocket.py
bzwei: In my test I only see InvalidStatusCode. I added code to handle InvalidStatus but could not test.

ansible_rulebook/cli.py
bzwei: Here wss and https share the same ssl_verify option. It can't be the case that one requires ssl verification but the other does not, can it?

mkanoor: @bzwei Its possible that the resource server and the auth server are different.

ansible_rulebook/token.py
```diff
@@ -0,0 +1,52 @@
...
+
+import aiohttp
+
+from ansible_rulebook.conf import settings
+from ansible_rulebook.exception import TokenNotFound
+
+logger = logging.getLogger(__name__)
+
+
+async def renew_token() -> str:
```
mkanoor, line 27: @bzwei can we please add some tests for this file?

ansible_rulebook/websocket.py
```diff
@@ -12,126 +12,216 @@
...
+        else:
+            # Connection succeeded - reset backoff delay
+            backoff_delay = BACKOFF_MIN
+            refresh = True
+
+    return result
+
+
+async def request_workload(activation_id: str) -> StartupArgs:
+    return await _connect_websocket(
```
mkanoor, line 128: @bzwei when the passed in token is bad will the program end or would it try the refresh token?

bzwei: In this case it will try refreshing once, and then re-throw the exception if it doesn't help.

ansible_rulebook/token.py
```diff
@@ -0,0 +1,52 @@
+#  Copyright 2023 Red Hat, Inc.
```
mkanoor, line 1: @bzwei should we make this 2024

mkanoor:
@bzwei the PR and commit title are not following the [conventional commit ](https://www.conventionalcommits.org/en/v1.0.0/)style. Most probably the pre commit hooks have not been installed in your dev env.

ansible_rulebook/cli.py
```diff
@@ -279,6 +296,14 @@ def main(args: List[str] = None) -> int:
     if args.execution_strategy:
         settings.default_execution_strategy = args.execution_strategy
 
+    settings.websocket_address = args.websocket_address
+    settings.websocket_ssl_verify = args.websocket_ssl_verify
+    settings.websocket_token_address = args.websocket_token_address
```
mkanoor, line 301: @bzwei Should we make this the URL I know the the previous one incorrectly uses address when its a url like ```ws://host.containers.internal:8001/api/eda/ws/ansible-rulebook
```

File: ansible_rulebook/websocket.py [hunk 15]
```diff
@@ -12,126 +12,216 @@
...
+class EventLogQueue:
+    queue: asyncio.Queue = field(default=None)
+    event: dict = field(default=None)
+
+
+async def send_event_log_to_websocket(event_log: asyncio.Queue):
+    logs = EventLogQueue()
+    logs.queue = event_log
+
+    return await _connect_websocket(
```
mkanoor, line 188: @bzwei previously we were using [async for](https://github.com/ansible/ansible-rulebook/blob/c72a42087508e2aa03c0a183cf140dcc2b5aa093/ansible_rulebook/websocket.py#L94) to put the connection in a loop so when it got interrupted the for loop will create a new connection. In the new design how is the loop recreated.

mkanoor:
@bzwei Since a lot of e2e tests depend on the audit events being correctly reported we should run these changes thru the [eda-qa](https://github.com/ansible/eda-qa) e2e tests to ensure we are not dropping any events

ansible_rulebook/websocket.py
bzwei: There is a `while True:` loop inside `_connect_websocket` function

bzwei:
> @bzwei Since a lot of e2e tests depend on the audit events being correctly reported we should run these changes thru the [eda-qa](https://github.com/ansible/eda-qa) e2e tests to ensure we are not dropping any events

Ran e2e tests in eda-qa. Auditing works fine.

mkanoor (approved):

hsong-rh (approved):
LGTM

PR #632, opened 2023-12-13
chore: include 3.12 as supported python version

Alex-Izquierdo:
Add python 3.12 as supported version. 
jira: https://issues.redhat.com/browse/AAP-13387


Alex-Izquierdo:
python 3.12 is not 100% confirmed yet as a requirement. On hold.

PR #635, opened 2024-01-03
Point to correct stream9 image tag

sc68cal:
`stream9-development` was last updated Oct 9 2022, while the`stream9` tag is being actively updated 

https://quay.io/repository/centos/centos?tab=tags

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/centos/centos:stream9-development
+FROM quay.io/centos/centos:stream9
 
 ARG USER_ID=${USER_ID:-1001}
 ARG APP_DIR=${APP_DIR:-/app}


Alex-Izquierdo:
@ttuffin If I remember correctly you used centos9-development because the image was ligther. Is that correct? If it is not maintained we should consider to move to regular stream9 or probably better to stream9-minimal.

ttuffin:
@Alex-Izquierdo thats correct. Using stream9 instead of stream9-development adds an extra ~600mb to the image size. I didn't realise that tag is no longer being updated though. Lets see how much reduction we get in #639.

ptoscano:
This was fixed long ago with commit c0dcb5828663eb21cded950b9a19d87c45cf805e. Hence, I'm closing this PR, as it is not needed anymore.

@sc68cal: sorry for the long delay here, and thanks again for your work!

PR #636, opened 2024-01-03
fix: [AAP-16740] display set facts and post events 

jshimkus-rh:


ansible_rulebook/action/run_playbook.py
```diff
@@ -208,7 +217,19 @@ async def _post_process(self):
         post_events = self.action_args.get("post_events", False)
 
         if rc == 0 and (set_facts or post_events):
-            logger.debug("set_facts")
+            # Default to output events at debug level.
+            level = logging.DEBUG
+
+            # If print_events is specified adjust the level to the display's
+            # current level to guarantee output.
+            if self.print_events:
```
mkanoor, line 225: @jshimkus-rh Instead of passing the print events thru can we just use a new settings field called print_events and just use it here.

jshimkus-rh: @mkanoor We could, as long as folks are okay with introducing a runtime modification to the settings based on specifying `--print-events`.  My searching indicates it would be the first time that would be done.

Because the instance is module-level it wouldn't pose a problem w/ multiple copies.  Though we might prefer making it a singleton for clarity.

mkanoor: @jshimkus-rh It's a command line option and we are planning on using the settings for many of these startup options. For the websocket URL and auth we are using the settings. It's set once at startup from the command line and it remains the same throughout the program. There aren't multiple tasks when we are parsing the command line so its a pretty safe option to just set it once.

mkanoor: @jshimkus-rh The goal of the --print-events command line option is to display the event structure so the user can look at the structure and write conditions. This was for external events, the post_event/set_fact/retract_fact are generating internal events from the output of job template or workflow. These were not getting displayed when --print-events was passed in on the command line. There is also a downside to displaying these events if they have any sensitive information so this is meant to be used when developing the conditions for the rulebook. This --print-event is different from the print_event action which prints the matching event that triggered the rule.

mkanoor:
@jshimkus-rh The individual commits are following the conventional commit guidelines but the PR title is not. Should the PR title also follow the conventional commit guidelines.

ansible_rulebook/action/run_playbook.py
jshimkus-rh: @mkanoor the post_event/set_fact/retract_fact were only modified to adhere to the coded common instantiation process; they didn't actually output anything.  As we've now changed to using settings for defining when to print events  they are back to their previous forms.

mkanoor (approved):

Alex-Izquierdo (approved):

PR #638, opened 2024-01-03
Include gcc package in container

sc68cal:
Many python packages will compile native extensions and require GCC

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN useradd -u $USER_ID -d $APP_DIR appuser
 WORKDIR $APP_DIR
 COPY . $WORKDIR
 RUN chown -R $USER_ID $APP_DIR
-RUN dnf install -y java-17-openjdk-devel python3-pip
+RUN dnf install -y java-17-openjdk-devel python3-pip gcc
 
 USER $USER_ID
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN useradd -u $USER_ID -d $APP_DIR appuser
 WORKDIR $APP_DIR
 COPY . $WORKDIR
 RUN chown -R $USER_ID $APP_DIR
-RUN dnf install -y java-17-openjdk-devel python3-pip gcc
+RUN dnf install -y java-17-openjdk-devel python3-pip gcc python-devel
 
 USER $USER_ID
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk


Alex-Izquierdo:
If those packages are not required by this base image, we should avoid increasing the size of the image. Any other custom image that uses this one as a base can install any necessary components.

sc68cal:
How do we handle the implicit coupling between those images and this base image? The issue is that this image uses a custom `$USER_ID` and authors of containers based off that have to switch between `USER 0` to be able to do `dnf install` and then know that they have to switch back to `$USER_ID` so that their install doesn't get buggered up with path and permission issues.

ttuffin:
@sc68cal what is the size of the image build after adding these packages?

Alex-Izquierdo:
> How do we handle the implicit coupling between those images and this base image? The issue is that this image uses a custom `$USER_ID` and authors of containers based off that have to switch between `USER 0` to be able to do `dnf install` and then know that they have to switch back to `$USER_ID` so that their install doesn't get buggered up with path and permission issues.

I don't think this a big deal, It is a common a good practice to use a non-root user as default and I think it is always expected to inspect the original source if you are going to alter/extend the original image. On the other hand, we should not add packages that are not strictly required only "just-in-case". Who extends the image is responsible to know what it is needed and how to do it.

sc68cal:
> @sc68cal what is the size of the image build after adding these packages?

With GCC and the Python development headers the image is 896MB on aarch64. The standard image, built from 5d4ebf1 is 710MB

sc68cal:
@Alex-Izquierdo @ttuffin 

I don't want to open an issue for this since it's a lot of mandatory fields for a feature request, but since we've broached the subject of custom images layered overtop of this image, can we consider setting `APP_DIR` to something different? Could we install `ansible-rulebook` and the related parts that make up ansible-rulebook either in `/usr/local/` or `/bin`? Because if I'm looking to use this image as my base layer for my rulebooks I would have expected `/app` to be available for my use, instead of where `ansible-rulebook` is installed.

Alex-Izquierdo:
Hi @sc68cal The [ansible-rulebook image](https://quay.io/repository/ansible/ansible-rulebook) is intended to be used as a base decision environment for [eda-server](https://github.com/ansible/eda-server) as well as a standalone CLI tool (in addition to our CI workflows) so, while it is not its main purpose, the extensibility of the image is in the scope. 

I think your request is reasonable. Look at the work of https://github.com/ansible/ansible-rulebook/pull/639 done by @dhaustein 
I think with the current stuff related with APP_DIR should be enough. Feel free to suggest there changes.

Alex-Izquierdo:
This change was finally added in https://github.com/ansible/ansible-rulebook/pull/616

PR #639, opened 2024-01-04
feat: reduce image size further

dhaustein:
Reduces the size of the resulting image to around 400 MB. 
- Uses multi-stage build process. 
- Python dependencies are packaged into one venv. 
- Java JRE is created custom, only keeping the modules Drools needs. 
- Both the venv and the JRE are copied to the second stage of the build based on stream9-minimal image to reduce the size even further.

Comparison with the original image:
```
$  buildah build --no-cache --tag localhost/ansible-rulebook:slim -f Dockerfile

$  podman images --format "table {{.Repository}}:{{.Tag}} {{.Size}}" | rg ansible-rulebook
31128877f949  localhost/ansible-rulebook:slim                  323 MB
5cbebc6a4a9b  localhost/ansible-rulebook:latest                702 MB
```

_Note: When built with podman, the image size will be ~100 MB larger for some reason (despite buildah being used internally in podman-build), even with compression forced on._

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ARG USER_ID=${USER_ID:-1001}
 ARG APP_DIR=${APP_DIR:-/app}
 
-FROM quay.io/centos/centos:stream9-development as build
+FROM quay.io/centos/centos:stream9 as build
 
 ARG USER_ID
 ARG APP_DIR
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,20 @@
-FROM quay.io/centos/centos:stream9-development
-
 ARG USER_ID=${USER_ID:-1001}
 ARG APP_DIR=${APP_DIR:-/app}
+
+FROM quay.io/centos/centos:stream9-development as build
+
+ARG USER_ID
+ARG APP_DIR
 ARG DEVEL_COLLECTION_LIBRARY=0
 ARG DEVEL_COLLECTION_REPO=git+https://github.com/ansible/event-driven-ansible.git
 
-USER 0
-RUN useradd -u $USER_ID -d $APP_DIR appuser
-WORKDIR $APP_DIR
-COPY . $WORKDIR
-RUN chown -R $USER_ID $APP_DIR
-RUN dnf install -y java-17-openjdk-devel python3-pip
-
-USER $USER_ID
+RUN dnf install -y java-17-openjdk-devel java-17-openjdk-jmods binutils python3-pip
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk
-ENV PATH="${PATH}:$APP_DIR/.local/bin"
-RUN pip install -U pip \
+
+ENV PATH="$APP_DIR/venv/bin:${PATH}"
+
+RUN python3 -m venv $APP_DIR/venv \
+    && pip install -U pip \
     && pip install ansible-core \
     ansible-runner \
     jmespath \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,21 +1,20 @@
-FROM quay.io/centos/centos:stream9-development
-
 ARG USER_ID=${USER_ID:-1001}
 ARG APP_DIR=${APP_DIR:-/app}
+
+FROM quay.io/centos/centos:stream9-development as build
```
Alex-Izquierdo, line 4: Can you look to use stream9-minimal? see https://github.com/ansible/ansible-rulebook/pull/635

dhaustein: @Alex-Izquierdo I will try the regular, non minimal, version for the build stage. Minimal will be missing a lot of packages needed in the build phase.

dhaustein:
**TODO**
- fix E2E tests failing on CI (https://github.com/ansible/ansible-rulebook/actions/runs/7411894851/job/20167469002?pr=639)
- use regular stream9 image instead of -development in the build stage (https://github.com/ansible/ansible-rulebook/pull/639#discussion_r1441952632)
- add comments

dhaustein:
@Alex-Izquierdo I remember you wanted to use python dependencies defined in the requirements file(s) instead of hard-coding them in the Dockerfile. I will want to move onto that in a future PR. Didn't want to mix that into this already big change.

Dockerfile
```diff
@@ -29,6 +28,41 @@ RUN pip install -U pip \
 RUN bash -c "if [ $DEVEL_COLLECTION_LIBRARY -ne 0 ]; then \
     ansible-galaxy collection install ${DEVEL_COLLECTION_REPO} --force; fi"
 
+WORKDIR $APP_DIR
+COPY . $WORKDIR
```
Alex-Izquierdo, line 32: Can you use "$APP_DIR" instead? I don't remember the details but I remember issues with $WORKDIR variable which is not honored in some places.

dhaustein: @Alex-Izquierdo I can but these two lines were there originally so I left them unchanged, only moved them. Should I change them then?

Alex-Izquierdo: Oh, that's true. I thought it was something new. Ok, if nobody has complained until now I think it is fine to leave them. :)

Dockerfile
dhaustein: https://github.com/ansible/ansible-rulebook/pull/639/commits/9ab6ad4022996a2ece0b06e2069ba5ba7cf10680

mkanoor:
@dhaustein I think we should also ensure that all the source plugins that we ship in [ansible.eda](https://github.com/ansible/event-driven-ansible/tree/main/extensions/eda/plugins/event_source) collection work with the slim image and all dependencies are satisfied.

dhaustein:
> @dhaustein I think we should also ensure that all the source plugins that we ship in [ansible.eda](https://github.com/ansible/event-driven-ansible/tree/main/extensions/eda/plugins/event_source) collection work with the slim image and all dependencies are satisfied.

@mkanoor I don't understand what you mean 100% but do you simply mean that I should test the new image with the EDA test suite from https://github.com/ansible/eda-qa/ ?

mkanoor:
> > @dhaustein I think we should also ensure that all the source plugins that we ship in [ansible.eda](https://github.com/ansible/event-driven-ansible/tree/main/extensions/eda/plugins/event_source) collection work with the slim image and all dependencies are satisfied.
> 
> @mkanoor I don't understand what you mean 100% but do you simply mean that I should test the new image with the EDA test suite from https://github.com/ansible/eda-qa/ ?

@dhaustein @ttuffin Currently there are 15 source plugins that we support in ansible.eda collection, I am not sure if QE has tests that cover each and every source plugin that we have. Each of these plugins have dependencies on other python packages, which is defined in the collection [requirements](https://github.com/ansible/event-driven-ansible/blob/main/requirements.txt). Our DE that we built should support these 15 source plugins, if there is a package that is missing then the rulebook using our official ansible.eda collection source will fail. At the least I am thinking that we should have some basic tests for each of these source plugins. To test the sources we would need access to other cloud providers and I am not sure if that integration test is happening.

dhaustein:
@mkanoor Testing it with every plugin is a rather tall order and I am not sure the EDA test suite covers everything. But I will see what I can do.

ttuffin:
> @mkanoor Testing it with every plugin is a rather tall order and I am not sure the EDA test suite covers everything. But I will see what I can do.

@mkanoor @dhaustein we have [integration tests](https://github.com/ansible/event-driven-ansible/tree/main/tests/integration) in the collection repository for this purpose, and a [workflow](https://github.com/ansible/event-driven-ansible/blob/main/.github/workflows/integration-tests.yml#L36) that installs the dependencies and runs the tests. That said, we only have coverage for four of the source plugins so we definitely need to look at increasing the coverage there.

ttuffin:
@dhaustein first and foremost it is a base DE + CLI, and since it's used in our CI pipelines we want to keep it as small as possible. Allowing for extensibility comes second to that, imo.

Alex-Izquierdo:
Hi @dhaustein the image is the base DE, which implicitly serves as an execution environment for the CLI. This makes it ideal for CI use, particularly for the eda-server. It's not intended for development purposes but rather for general usage. I believe that focusing on size, a reasonable requirement for a production-ready image, doesn't conflict with the goal of extensibility. Users can always build their own images if they have more complex or specific needs.

Regarding the tests for our plugins in the collection I don't think that the API test suite is the right place for it. All the plugins there should have integration tests with ansible-rulebook and the source repo is probably the ideal place for it, eventually we could use them to test this image as @mkanoor suggested. I don't think he is proposing a blocker for this PR, I think he just wants to make sure that the collection integrates well with the image or the CLI. We could just do a manual check for now, but I think it is an interesting milestone for later iterations. I believe we already have a ticket in our internal Jira for it.

PR #641, opened 2024-01-08
fix: add activation_instance_id alone with activation_id

bzwei:
Both keys will co-exist in the message body sent to eda-server for backward compatibility reason. activation_id will be removed in the future.

Fixes AAP-18855: ansible-rulebook reports activation instance ID as activation ID

### DIFF
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -494,6 +494,7 @@ class RuleSetRunner:
                     action=action,
                     action_uuid=str(uuid.uuid4()),
                     activation_id=settings.identifier,
+                    activation_instance_id=settings.identifier,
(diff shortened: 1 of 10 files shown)


mkanoor:
@bzwei In the backend there might be warning messages because the attribute is not defined here https://github.com/ansible/eda-server/blob/01746bec37107c64bd0e11a58adb6600015e6951/src/aap_eda/wsapi/messages.py#L31

bzwei:
> @bzwei In the backend there might be warning messages because the attribute is not defined here https://github.com/ansible/eda-server/blob/01746bec37107c64bd0e11a58adb6600015e6951/src/aap_eda/wsapi/messages.py#L31

I tested there is no warning.

mkanoor (approved):

hsong-rh (approved):

PR #644, opened 2024-01-16
fix: handle unexpected error raised from websocket server

bzwei:
Enhance error handling for websocket communication. Differenciate error at connection or at sending or receiving. Will not retry if an unexpected error (code 1011) is raised by the server.

fix AAP-19535: Ansible-rulebook keeps on retrying requesting workload through websocket


ansible_rulebook/websocket.py
```diff
@@ -137,20 +157,26 @@ async def _handle_request_workload(
...
+            json.dumps(
+                dict(
+                    type="Worker",
+                    activation_id=activation_instance_id,
+                    activation_instance_id=activation_instance_id,
+                )
             )
         )
-    )
+    except Exception as e:
```
mkanoor, line 170: @bzwei Can we be explicit about which exception we are catching, also we need a log line if we are catching Exception

ansible_rulebook/websocket.py
```diff
@@ -206,7 +232,11 @@ async def _handle_send_event_log(
 
     if logs.event:
         logger.info("Resending last event...")
-        await websocket.send(json.dumps(logs.event))
+        json_str = json.dumps(logs.event)
+        try:
+            await websocket.send(json_str)
+        except Exception as e:
```
mkanoor, line 238: @bzwei Can we not use https://websockets.readthedocs.io/en/stable/reference/exceptions.html#websockets.exceptions.ConnectionClosed

mkanoor: Catching and suppressing Exception is never a good idea.

ansible_rulebook/websocket.py
```diff
@@ -58,68 +85,55 @@ async def _connect_websocket(
...
-                # Increase delay with truncated exponential backoff.
-                backoff_delay = backoff_delay * BACKOFF_FACTOR
-                backoff_delay = min(backoff_delay, BACKOFF_MAX)
-                continue
-        else:
-            # Connection succeeded - reset backoff delay
-            backoff_delay = BACKOFF_MIN
-            refresh = True
+        except websockets.exceptions.InvalidStatusCode as e:
+            if refresh and e.status_code == 403:
```
mkanoor, line 106: @bzwei Can the 2 403's logic be put into a function, they are running the same code

ansible_rulebook/websocket.py
```diff
@@ -58,68 +85,55 @@ async def _connect_websocket(
...
+            else:
+                raise  # abort
+        except websockets.exceptions.ConnectionClosedError as e:
+            if e.code == 1011:
+                # unexpected error raised from server
+                raise  # abort
+            else:
+                backoff_delay = await _wait_before_retry(backoff_delay)
+        except websockets.exceptions.ConnectionClosedOK:
+            backoff_delay = await _wait_before_retry(backoff_delay)
```
mkanoor, line 136: @bzwei Shouldn't this be only called if there handler wants to have a retry.

bzwei: There is no problem to retry for both getting workload or sending event log. I removed the retry parameter. It is unlikely to meet connection close situation while getting the workload, but even if it happens we should be able to retry automatically.

ansible_rulebook/websocket.py
```diff
@@ -55,79 +85,53 @@ async def _connect_websocket(
...
-            backoff_delay = BACKOFF_MIN
-            refresh = True
-
-    return result
+        except websockets.exceptions.InvalidStatusCode as e:
+            if refresh_token and e.status_code == 403:
+                await _update_authorization_header(extra_headers)
+                # Only attempt to refresh token once. If a new token cannot
+                # establish the connection, something else must have caused 403
+                refresh_token = False
```
mkanoor, line 108: @bzwei Will this cause the refresh token to be updated only once, since ansible-rulebook is a long running process maybe running for days/weeks if they expire the token every 24 hours will be able to refresh the token every day. or will this block it and will allow only 1 refresh

bzwei: Thanks for pointing out. I wanted to fix this part, but forgot. I will update.

bzwei: updated

ansible_rulebook/websocket.py
mkanoor: @bzwei Since the server controls when it is going to close the connection, if it closes the connection after it has given the worker payload will this cause a refetch of the payload.

mkanoor: For reference this what the old code did when the server closed the connection
https://github.com/ansible/ansible-rulebook/blob/eac79808254997b47859d6422fd4fb4a0ffdaf2e/ansible_rulebook/websocket.py#L87

mkanoor (approved):

hsong-rh (approved):
LGTM

PR #653, opened 2024-01-22
chore: use psycopg[binary] in development

mkanoor:
On the MacOS the dev environment was causing the postgresql to be installed if we use the psycopg[c].
This PR makes the psycopg[binary] in development and uses psycopg[c] in production.


.github/actions/setup/action.yml
```diff
@@ -22,7 +22,7 @@ runs:
 
     - name: Install package
       shell: bash
-      run: python -m pip install .
+      run: python -m pip install .[development]
```
Alex-Izquierdo, line 25: I think here we want also production, don't?

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

hsong-rh (approved):

.github/actions/setup/action.yml
mkanoor: 👍

PR #659, opened 2024-01-30
docs: add warning about webhook host usage in podman

dhaustein:
- added a small warning section to Callback plugins about when running ansible-rulebook in Podman
  - Ref: https://issues.redhat.com/browse/AAP-18943

Discussed this with the Ansible docks team (Jameira) , downstream we have no good place to put this. This was the only place we could think of.


dhaustein:
Will be included in a downstream KB instead.

Issue #510, opened 2023-05-18
The run_job_template action is not passing the host limit correctly


AdamMack2007:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/ansible-rulebook/issues) for duplicates.
- [X] I understand that ansible-rulebook is open source software provided for free and that I might not receive a timely response.

### Bug Summary

When a condition is triggered, the run_job_template action is not successfully passing the limit variable to Automation Controller.

The log output from EDA looks correct, as seen in the 'Actual Results' log output, but Controller shows the limit as 'Limit: {{ event.payload.meta.hosts }}'

### Environment

__version__ = '0.13.0'
  Executable location = /opt/app-root/bin/ansible-rulebook
  Drools_jpy version = 0.3.1
  Java home = /usr/lib/jvm/java-17-openjdk
  Java version = 17.0.7
  Python version = 3.9.16 (main, Dec 21 2022, 10:57:18) [GCC 8.5.0 20210514 (Red Hat 8.5.0-17)]

### Steps to reproduce

Testing via curl

curl -X POST http://192.168.6.109:5000/endpoint -d '{"message": "BGP neighbor down","meta": {"hosts": "leaf3"}}' -H 'Content-Type: application/json'

### Actual results

EDA log output

```
2023-05-18 17:03:06,051 - ansible_rulebook.engine - INFO - ruleset define: {"name": "Example", "hosts": ["all"], "sources": [{"EventSource": {"name": "ansible.eda.webhook", "source_name": "ansible.eda.webhook", "source_args": {"host": "0.0.0.0", "port": 5000}, "source_filters": []}}], "rules": [{"Rule": {"name": "Get device info if BGP neighbor down", "condition": {"AllCondition": [{"EqualsExpression": {"lhs": {"Event": "payload.message"}, "rhs": {"String": "BGP neighbor down"}}}]}, "actions": [{"Action": {"action": "run_job_template", "action_args": {"name": "Parsers", "job_args": {"limit": "{{ event.payload.meta.hosts }}"}, "organization": "MackNet"}}}], "enabled": true}}]}

2023-05-18 17:03:06,070 - ansible_rulebook.engine - INFO - load source

2023-05-18 17:03:07,047 - ansible_rulebook.engine - INFO - load source filters

[...]

PR #660, opened 2024-02-07
docs: add 'inserting hosts' section in docs

Alex-Izquierdo:


docs/host_limit.rst
```diff
@@ -17,3 +17,26 @@ Event data example:
...
+
+Example:
+
+.. code-block:: yaml
+
+    sources:
+        - ansible.eda.webhook:
+            port: 4444
+            host: 0.0.0.0
+        filters:
```
ttuffin, line 40:
```suggestion
          filters:
```

ttuffin (dismissed):

docs/host_limit.rst
```diff
@@ -17,3 +17,26 @@ Event data example:
...
+The plugins may not send metadata like for example the `ansible.eda.webhook` plugin which
+sends the arbitrary payload under the `payload` key.
+
+To accommodate this, the EDA collection provides the `insert_hosts_to_meta` filter,
+allowing any plugin to customize the value of `event.meta.hosts` based on the contents
+of a specific key in the event.
+
+Example:
+
+.. code-block:: yaml
```
mkanoor, line 34: @Alex-Izquierdo Can we also add this into the webhook source plugin notes in the collection

Alex-Izquierdo: Sure we can. But what we need is definitely address the whole documentation for source plugins.

bzwei (approved):

docs/host_limit.rst
Alex-Izquierdo: are you fine to merge this?

mkanoor:
@Alex-Izquierdo we need to clarify another piece here
1. the insert_meta_filter will silently ignore if the key is missing which could be bad. For example in https://github.com/ansible/ansible-rulebook/issues/510 the crux of the issue was host_path had an incorrect key, note that event is not needed.
```
     filters:
        - ansible.eda.insert_hosts_to_meta:
            host_path: **event.payload.event.fields.source**
```

The host_path should have been just **payload.event.fields.source** but the [filter](https://github.com/ansible/event-driven-ansible/blob/fbf92e8de074565bec91cca3039907c987c5ce7a/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py#L48) ignores this error 
Which caused the customer to think it's broken and they tried different ways to get the hosts into the job_args and started using the limit.

So we need to document that event prefix is needed for conditions and action args but not for source args and source filters.
We could have handled it differently by adding an extra argument to the filter say 
 ```raise_exception: True``` 
and have it terminate the rulebook if it was bad once the customer was content with the hosts being in there they could have turned it off with 
 ```raise_exception: False```.
 If the customer is putting the host_path in the filter they need to be assured that it is there in every payload or if they feel some payloads will have it and others won't they can use the raise_exception accordingly.

2. It would also have helped if we had put a logger.warn in the source filter so it was apparent from the beginning that there is something wrong.

mkanoor (changes requested):
Can we add some more notes on how the filter is suppressing key errors

Alex-Izquierdo:
> @Alex-Izquierdo we need to clarify another piece here
> 
> 1. The insert_hosts_to_meta filter will silently suppress any issues  if the key is missing which could be bad. For example in [The run_job_template action is not passing the host limit correctly #510](https://github.com/ansible/ansible-rulebook/issues/510) the crux of the issue was host_path had an incorrect key, note that event is not needed.
> 
> ```
>      filters:
>         - ansible.eda.insert_hosts_to_meta:
>             host_path: event.payload.event.fields.source
> ```
> 
> The host_path should have been just **payload.event.fields.source** but the [filter](https://github.com/ansible/event-driven-ansible/blob/fbf92e8de074565bec91cca3039907c987c5ce7a/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py#L48) ignores this error Which caused the customer to think it's broken and they tried different ways to get the hosts into the job_args and started using the limit.
> 
> So we need to document that event prefix is needed for conditions and action args but not for source args and source filters. We could have handled it differently by adding an extra argument to the filter say `raise_exception: True` and have it terminate the rulebook if it was bad, once the customer was content that the hosts is there in the payload they could have turned it off with `raise_exception: False`. If the customer is putting the host_path in the filter they need to be assured that it is there in every payload or if they feel some payloads will have it and others won't they can use the raise_exception accordingly.
> 
> 2. It would also have helped if we had put a logger.warn in the source filter so it was apparent from the beginning that there is something wrong.

I think we have to clarify the scope of this PR and the documentation of ansible-rulebook. What you say is not general, but specific to the insert_hosts_to_meta. This section is just to indicate the use case of the plugin regarding host limits, not to document the plugin itself. 

Unless we decide that ansible-rulebook docs portal is going to store the documentation for the plugins of the EDA collection, such a documentation should be in the documentation for the plugins (wherever it is going to be) and not here otherwise it will create more confusion. Documentation that by the way, right now only exists in the comments of the plugin, we already have a task to create that documentation. 

So I can remark here that it should not include the "event" prefix until we write the documentation of the plugins and/or update the plugin to handle better the issue that you have described, but we should not put here the whole documentation for that plugin. 

We definitely should prioritize the documentation of the plugins.

docs/filters.rst
```diff
@@ -45,6 +48,17 @@ e.g.
             exclude_keys: ['*_url', '_links', 'base', 'sender', 'owner', 'user']
         - dashes_to_underscores:
 
+
```
bzwei, line 51: You may want to push this example down after the next paragraph which explains what the first example is doing.

bzwei:
I created  https://github.com/ansible/event-driven-ansible/pull/197 to add the option to raise an exception from the insert_hosts_to_meta filter.

Alex-Izquierdo:
> I created [ansible/event-driven-ansible#197](https://github.com/ansible/event-driven-ansible/pull/197) to add the option to raise an exception from the insert_hosts_to_meta filter.

@bzwei great! If we merge it first, I can update this.

docs/host_limit.rst
```diff
@@ -17,3 +17,32 @@ Event data example:
...
+The plugins may not send metadata like for example the `ansible.eda.webhook` plugin which
+sends the arbitrary payload under the `payload` key.
+
+To accommodate this, the EDA collection provides the `insert_hosts_to_meta` filter,
+allowing any plugin to customize the value of `event.meta.hosts` based on the contents
+of a specific key in the event.
+
+.. note::
+
+    The `host_path` in filter `insert_hosts_to_meta` don't expect the `event` root key unlike conditions
```
mkanoor, line 34: @Alex-Izquierdo 
this might be better
```
The host_path argument in the filter `insert_hosts_to_meta` does not need the `event.` prefix like conditions or action arguments.
```

docs/host_limit.rst
```diff
@@ -17,3 +17,32 @@ Event data example:
...
+
+
+Example:
+
+.. code-block:: yaml
+
+    sources:
+        - ansible.eda.webhook:
+            port: 4444
+            host: 0.0.0.0
```
mkanoor, line 45: @Alex-Izquierdo The newer webhook plugin has been defaulting the host to 0.0.0.0 so we can skip it here.

Alex-Izquierdo: Removed

docs/host_limit.rst
Alex-Izquierdo: changed

mkanoor (approved):

bzwei (approved):

PR #664, opened 2024-02-19
fix: catch ContentTypeError

bzwei:
AAP-20301: renew_token function in ansible-rulebook assumes response is always JSON


Alex-Izquierdo (approved):

ansible_rulebook/token.py
```diff
@@ -32,7 +32,12 @@ async def renew_token() -> str:
             data={"refresh": settings.websocket_refresh_token},
             ssl_context=_sslcontext(),
         ) as resp:
-            data = await resp.json()
+            try:
+                data = await resp.json()
+            except aiohttp.client_exceptions.ContentTypeError as e:
+                logger.error(f"failed to renew token. Error: {e}")
+                msg = "Token URL does not return expected format"
```
mkanoor, line 39: @bzwei I think the msg could be
```
 msg = "Refresh Token URL did not return expected format"
```

mkanoor (approved):

PR #665, opened 2024-02-22
Automated release

benthomasson:
This PR creates an automated release process using github actions and an ansible rulebook.  

This can also be run locally using this command:

```
ansible-playbook tools/ansible/release.yml -i localhost -e github_token={YOUR_GH_TOKEN}
```


mkanoor:
@benthomasson Would this get published to PyPi at the end?

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  stage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    permissions:
+      packages: write
+      contents: write
+    steps:
+      - name: Checkout dab
```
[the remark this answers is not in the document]
benthomasson: Thanks I missed that.

benthomasson:
> @benthomasson Would this get published to PyPi at the end?

Yes, we probably should.

mkanoor:
@benthomasson Would switching from semver to calver have issues if there are users using versions in their requirements.txt files or Dockerfiles?

benthomasson:
> @benthomasson Would switching from semver to calver have issues if there are users using versions in their requirements.txt files or Dockerfiles?

The calver will be greater than semver, so going that way won't be a problem.

Dostonbek1 (approved):

mkanoor:
@benthomasson Should the release document that we are switching to CalVer?

Alex-Izquierdo:
The main goal of this automation is the automatic publication of the package in pypi. That step would be even more complicated with this pipeline, because with the current manual procedure we can generate the artifacts and publish them in pypi with the same command but now we should download the artifacts generated by this pipeline and publish them manually. I recommend to add that step before merge.

benthomasson:
> The main goal of this automation is the automatic publication of the package in pypi. That step would be even more complicated with this pipeline, because with the current manual procedure we can generate the artifacts and publish them in pypi with the same command but now we should download the artifacts generated by this pipeline and publish them manually. I recommend to add that step before merge.

We should not be building any released binary locally.  It is very easy to slip in development files into a release build without realizing it if it is built locally.

Alex-Izquierdo:
> We should not be building any released binary locally. It is very easy to slip in development files into a release build without realizing it if it is built locally.

The local build/publish is already handled by make: https://github.com/ansible/ansible-rulebook/blob/main/Makefile#L80

My point is that the publication on pypi is the main goal of an automated release process. Users usually won't consume the artifacts from github, they will run `pip install ansible-rulebook` as it is described in the documentation. 

With the current process you only have to run make. Now, with this pipeline as it is now, I should have to go to github, download the artifacts generated and publish them in pypi, which is a more tedious process than `make release`

That's why I'm requesting to automate also the publication on pypi.

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+
+env:
+  LC_ALL: "C.UTF-8" # prevent ERROR: Ansible could not initialize the preferred locale: unsupported locale setting
+
+on:
+  workflow_dispatch:
+
+jobs:
+  stage:
+    runs-on: ubuntu-latest
```
Alex-Izquierdo, line 12: Can you add `if: github.repository == 'ansible/ansible-rulebook'` to make clear this workflow is not intended to be executed by forks?

Alex-Izquierdo (commented):
As it is now, the release could be done accidentally with the code of any branch. I think it would be good to ensure that only can be executed from the expected branches.

Alex-Izquierdo:
I think calver doesn't meet our requirements. We have to maintain multiple versions right now are two: 
main and https://github.com/ansible/ansible-rulebook/tree/stable/1.0

Eventually we will have to maintain more versions. The calver itself doesn't tell you what is the branch. Everytime you need to check the releases for a given supported version you would have to inspect the tags per branch.

.github/workflows/release.yml
benthomasson: It doesn't work on forks because it doesn't have permissions to push to this repo.

benthomasson:
> I think calver doesn't meet our requirements.

This is a valid concern.  I am debating this myself.  With semver we need to be able to bump the patch number every week to do automated weekly releases.   That might make the patch number meaningless.

benthomasson:
I am closing this in favor of #667

PR #672, opened 2024-04-05
feat: support string interpolation of encrypted variables

mkanoor:
https://issues.redhat.com/browse/AAP-22487

We can now support
```
"{{ some_encrypted_variable }}"
```
and also
```
"Some {{ some_encrypted_variable }}"
```


mkanoor:
> Since we do decryption earlier than the substitution, we may not need to check wether the string is encrypt again in https://github.com/mkanoor/ansible-rulebook/blob/213545c6ee53bd3b452cde8ad162d4d629c23f4c/ansible_rulebook/util.py#L69

@bzwei These relates to variables that don't go thru Jinja substitution because the vaulted strings are directly embedded in the rulebook like [here](https://github.com/ansible/ansible-rulebook/blob/0e417bfbd49ef7bb69fe057202a8ebf274085490/tests/e2e/files/rulebooks/test_vaulted_rulebook.yml#L24)

bzwei (approved):

PR #673, opened 2024-04-09
fix: check if variables are decryptable outside of substitution

mkanoor:
We were relying on checking if a string is decryptable by using substitution which causes issues when the data can contain data that will only be available when events are processed

Added a separate function called decryptable that can be called to check if the passed in 

* strings in extra vars if vaulted can be decrypted
* strings in action args if vaulted can be decrypted

source args are used at startup so that failure is seen right away

https://issues.redhat.com/browse/AAP-22487

This is part of the story where we want to catch decryption errors at startup so we don't have to wait for the rulebook to start executing and report decryption errors when running actions. https://issues.redhat.com/browse/AAP-19611

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -32,7 +32,7 @@ from ansible_rulebook.conf import settings
 from ansible_rulebook.engine import run_rulesets, start_source
 from ansible_rulebook.job_template_runner import job_template_runner
 from ansible_rulebook.rule_types import RuleSet, RuleSetQueue
-from ansible_rulebook.util import substitute_variables
+from ansible_rulebook.util import decryptable
 from ansible_rulebook.validators import Validate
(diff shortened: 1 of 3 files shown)


ansible_rulebook/app.py
```diff
@@ -282,7 +280,7 @@ def validate_actions(startup_args: StartupArgs) -> None:
                 if startup_args.check_vault:
                     for _key, value in action.action_args.items():
                         # attempt to substitute, raise an error on failure
-                        substitute_variables(value, {})
+                        decryptable(value)
```
bzwei, line 283: since decryptable can take a dict, do we still need this for loop?

bzwei (approved):

fincamd (approved):

ansible_rulebook/app.py
mkanoor: Changed

PR #679, opened 2024-04-30
feat: add terse but informative task cancellation message

gebhardtr:
Related: https://issues.redhat.com/browse/AAP-22056

Make it obvious that tasks are being cancelled when a source plugin shuts down (cleanup) or an error occurs while running the ruleset.

### DIFF
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -156,9 +156,15 @@ class RuleSetRunner:
                 timeout=self.shutdown.delay,
             )
 
-        for task in self.active_actions:
-            logger.debug("Cancelling active task %s", task.get_name())
-            task.cancel()
+        if self.active_actions:
+            logger.info(
+                "Cancelling %d active tasks, ruleset %s cleanup",
+                len(self.active_actions),
+                self.name,
+            )
+            for task in self.active_actions:
+                logger.debug("Cancelling active task %s", task.get_name())
+                task.cancel()
         if self.shutdown:
             await self.event_log.put(
                 dict(


benthomasson (approved):

ansible_rulebook/rule_set_runner.py
```diff
@@ -128,10 +128,10 @@ async def run_ruleset(self):
             tasks.append(self.source_loop_task)
             await asyncio.wait([self.action_loop_task])
         except asyncio.CancelledError:
-            logger.debug("Cancelled error caught in run_ruleset")
+            logger.error("Cancelled error caught in run_ruleset")
             for task in tasks:
                 if not task.done():
-                    logger.debug("Cancelling (2) task %s", task.get_name())
+                    logger.info("Cancelling (2) task %s", task.get_name())
```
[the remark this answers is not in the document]
gebhardtr: @mkanoor great point -- I'm always torn on how to distinguish errors from their handler side effects. I think your idea is good and it's consistent.

I assume it won't result in a lot of extra log messages because async cancellations should be rare, is that true?

ansible_rulebook/rule_set_runner.py
```diff
@@ -157,7 +157,7 @@ async def _cleanup(self):
             )
 
         for task in self.active_actions:
-            logger.debug("Cancelling active task %s", task.get_name())
+            logger.info("Cancelling active task %s", task.get_name())
```
gebhardtr, line 160: @benthomasson and @mkanoor, [related comment](https://github.com/ansible/ansible-rulebook/pull/679/files#r1585503286): should we also reclassify all the log messages to `info` for the cleanup tasks?

mkanoor: At one point we used to log all these messages at info level it was too much information so we ended up removing them. Should a summary message be more appropriate saying how many actions got canceled

ansible_rulebook/rule_set_runner.py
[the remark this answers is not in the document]
gebhardtr: Omitting this change for now because it exceeds the original scope.

ansible_rulebook/rule_set_runner.py
gebhardtr: I left the loop message at `debug` and added a new `info` message with only the count of tasks to be cancelled.

mkanoor (approved):

ttuffin (approved):

ansible_rulebook/rule_set_runner.py
benthomasson: Start-up and clean-up messages are a finite amount of messages and are probably useful for end users.     I like that we log that we are pulling images in job history for instance.   AWX doesn't do that and users wonder why it is taking so long to start a job.

PR #683, opened 2024-05-16
fix: ensure inventory files are copied

Alex-Izquierdo:


Alex-Izquierdo:
After discussion with @mkanoor we agreed to fix only the internal reference to the inventory path when it is a directory and keep the current behavior of copy only the inventory file if it is a file. 

Closed in favor of https://github.com/ansible/ansible-rulebook/pull/684

PR #694, opened 2024-06-09
fix: websocket logging

mkanoor:
Set the log level of websocket library to be ERROR since at DEBUG level it logs Authorization Headers leaking tokens

### DIFF
--- a/ansible_rulebook/websocket.py
+++ b/ansible_rulebook/websocket.py
@@ -34,6 +34,7 @@ from ansible_rulebook.token import renew_token
 from ansible_rulebook.vault import Vault, has_vaulted_str
 
 logger = logging.getLogger(__name__)
+logging.getLogger("websockets").setLevel(logging.ERROR)
 
 
 BACKOFF_MIN = 1.92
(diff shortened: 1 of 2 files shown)


ansible_rulebook/websocket.py
```diff
@@ -34,6 +34,7 @@
 from ansible_rulebook.vault import Vault, has_vaulted_str
 
 logger = logging.getLogger(__name__)
+logging.getLogger("websockets").setLevel(logging.ERROR)
```
Alex-Izquierdo, line 37: The problem is the "DEBUG" level, right? Should we honor the `-vv` and set the INFO level instead of forcing always the ERROR one?

mkanoor: @Alex-Izquierdo That's when it leaks from the websocket library logging every HTTP byte. We never want that level of tracing from the websocket layer we only need errors from it.

Alex-Izquierdo (approved):

benthomasson (approved):

PR #697, opened 2024-06-14
fix: switch slugs based on presence of path in controller url

mkanoor:
ansible-rulebook needs to be able to connect to 2.4 controller as well as the contoller in version 2.5 via the gateway

This is accomplished by the paths in the URL
If we need to talk to the 2.4 controller the URL is
* https://my_old_controller/
 If we need to talk to the 2.5 contoller via Gateway the URL
* https://my_gateway/api/controller/

We append the slugs based on the presence of path

If the path is missing we append
  * /api/v2/config
  * /api/v2/unified_job_templates/

If the path is defined in the URL we append the following slug
  * /v2/config
  * /v2/unified_job_templates/

https://issues.redhat.com/browse/AAP-25509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## [Unreleased]
 ### Changed
 ### Added
+- Support for switching slugs to connect to controller via gateway
 ### Fixed
 - Fix log level for websocket
 
--- a/ansible_rulebook/action/run_job_template.py
+++ b/ansible_rulebook/action/run_job_template.py
@@ -125,6 +125,7 @@ class RunJobTemplate:
             a_log["reason"] = {"error": self.controller_job["error"]}
         else:
             logger.info(f"job results url: {a_log['url']}")
+            logger.info(f"job id: {a_log['job_id']}")
 
         await self.helper.send_status(a_log)
         set_facts = self.action_args.get("set_facts", False)
--- a/ansible_rulebook/action/run_workflow_template.py
+++ b/ansible_rulebook/action/run_workflow_template.py
@@ -131,6 +131,7 @@ class RunWorkflowTemplate:
             a_log["reason"] = {"error": self.controller_job["error"]}
         else:
             logger.info(f"job results url: {a_log['url']}")
+            logger.info(f"job id: {a_log['job_id']}")
 
         await self.helper.send_status(a_log)
         set_facts = self.action_args.get("set_facts", False)
--- a/ansible_rulebook/cli.py
+++ b/ansible_rulebook/cli.py
@@ -141,7 +141,9 @@ def get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--controller-url",
         help="Controller API base url, e.g. https://host1:8080 can also be "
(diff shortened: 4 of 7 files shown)


Alex-Izquierdo:
Hi @mkanoor 
Hi Madhu, I see a couple of problems
1. AFAIK controller/awx still can work in standalone way and ansible-rulebook should support it. I don't think it is "legacy", it's just an additional way. 

2. There is no rule for the base path for controller 2.4/standalone. It can be served in the root or in any other path. Maybe you are misunderstanding the tests included in https://github.com/ansible/ansible-rulebook/pull/658 when I fixed the support for custom base paths. 

So, I think the right solution is to receive a param to let it know which is the mode to work(standalone or gateway), or introduce a breaking change where "controller_url" would be "controller_api_url" and it is the caller the responsible to send the right value and ansible-rulebook always uses the same slugs.

Alex-Izquierdo (commented):
Hi @mkanoor
Hi Madhu, I see a couple of problems

1 AFAIK controller/awx still can work in standalone way and ansible-rulebook should support it. I don't think it is "legacy", it's just an additional way.

2 There is no rule for the base path for controller 2.4/standalone. It can be served in the root or in any other path. Maybe you are misunderstanding the tests included in https://github.com/ansible/ansible-rulebook/pull/658 when I fixed the support for custom base paths.

So, I think the right solution is to receive a param to let it know which is the mode to work(standalone or gateway), or introduce a breaking change where "controller_url" would be "controller_api_url" and it is the caller the responsible to send the right value and ansible-rulebook always uses the same slugs.

mkanoor:
@Alex-Izquierdo The controller_urls main purpose was an API end point and if the user gives us an api path we should be honoring it. The other job related URL seems like it was just a guess that only works when we are connecting to the Controller directly. When there are other services in between or if there is a different internal/external.
From the eda_server side we have access to the [JobID ](https://github.com/ansible/eda-server/blob/a347c4a3fc174766003569df1a37e3aa3340f688/src/aap_eda/wsapi/messages.py#L44)and ansible-rulebook is sending the [Job ID](https://github.com/ansible/ansible-rulebook/blob/3b15bcc401e6dedd1f30f0e3b3da2e294f9ae3eb/ansible_rulebook/action/run_job_template.py#L117)
So we might be able to do the URL creation of the server side.

Alex-Izquierdo:
So, what is the plan? 
Are we going to remove the url guessing and just return the job id? In addition to update the documentation to explain that controller_url must contain the api path so we can just use the same slug in all the cases?

mkanoor:
@Alex-Izquierdo The jobid is already being sent to the server. For ansible-rulebook that URL has always been an API end point. So we still need the /api to be attached for legacy mode to stay backward compatible for the newer style we expect the callers to pass in the path with api embedded in it. The controller URL I think we should be able to fix when it comes in using some env var we were told that exists.

mkanoor:
@Alex-Izquierdo Docs updated

Alex-Izquierdo (approved):

bzwei (approved):

PR #707, opened 2024-08-27
Add jUnit test reporter and upload results

zkayyali812:
Outputs junit test results when running unit tests to an xml file.
Upload this file for integration into The Platform Readiness Dashboard


mkanoor:
@zkayyali812 What is junit? Isn't that something from the Java world?

jneedle (dismissed):

.github/workflows/ci.yml
```diff
@@ -68,6 +68,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: All tests
+        env:
+          PDE_ORG_RESULTS_AGGREGATOR_UPLOAD_USER: ${{ vars.PDE_ORG_RESULTS_AGGREGATOR_UPLOAD_USER }}
```
Alex-Izquierdo, line 72: I'm not sure what will be the behavior of Git Hub in forks here, if it will fail because the variables or secrets don't exist or just it will set an empty value. 

Can you create this same PR from your fork? I think that is the simplest way to test if the change will work with forks.

zkayyali812: Yes, lets try it out :)

zkayyali812: PR here - https://github.com/ansible/ansible-rulebook/pull/709

zkayyali812: After modifying my branch check to check if the variables are empty or not, this works.

Alex-Izquierdo (approved):

jneedle (approved):

PR #711, opened 2024-08-31
feat: support receiving file contents and env vars from EDA Server

mkanoor:
AAP Credential Types support File contents via file injectors.
https://docs.ansible.com/automation-controller/latest/html/userguide/credential_types.html

We create temporary files with the data sent from Server The source plugin can access the temporary file names to access the data.

Supports single file as well as multiple files.

To access the filenames in the source rulebook you can use the following extra vars

- eda.filename or
- eda.filename.<<key_name1>>
- eda.filename.<<key_name2>>

e.g.
```
- name: Use payload file to check events
  hosts: all
  sources:
    - ansible.eda.generic:
        payload_file: "{{ eda.filename.test_payload_file }}"
```

The env variables are treated like extra_vars and added to the current processes environment

https://issues.redhat.com/browse/AAP-25519

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -32,7 +32,11 @@ from ansible_rulebook.conf import settings
 from ansible_rulebook.engine import run_rulesets, start_source
 from ansible_rulebook.job_template_runner import job_template_runner
 from ansible_rulebook.rule_types import RuleSet, RuleSetQueue
-from ansible_rulebook.util import decryptable
+from ansible_rulebook.util import (
+    decryptable,
+    decrypted_context,
+    substitute_variables,
+)
 from ansible_rulebook.validators import Validate
 from ansible_rulebook.vault import has_vaulted_str
 from ansible_rulebook.websocket import (
@@ -74,6 +78,12 @@ async def run(parsed_args: argparse.Namespace) -> None:
             raise WebSocketExchangeException(
                 "Error communicating with web socket server"
             )
+        context = decrypted_context(startup_args.variables)
+        startup_args.env_vars = substitute_variables(
+            startup_args.env_vars, context
+        )
+        for k, v in startup_args.env_vars.items():
+            os.environ[k] = str(v)
     else:
         startup_args = StartupArgs()
         startup_args.variables = load_vars(parsed_args)
--- a/ansible_rulebook/websocket.py
+++ b/ansible_rulebook/websocket.py
@@ -180,6 +180,8 @@ async def _handle_request_workload(
 
     project_data_fh = None
     response = StartupArgs()
+    non_fq_key = False
+    file_template_vars = {}
     while True:
         msg = await websocket.recv()
         data = json.loads(msg)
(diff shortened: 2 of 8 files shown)


Alex-Izquierdo (commented):
Questions:
1. Should the CHANGELOG be updated?
2. Does make sense to support `non_fq_key`? I feel it would be simpler if `filename` is `filenames` and it is always a dict with N nested keys. 

Concerns: 
I'm concerned about the amount of assumptions that we make from the server side that would end up in unhandled exceptions:
* `template_key` can be None, 
* `data.get("data")` can be None
* what if `len(keys) == 1 and template_key != "template"` ?

Alex-Izquierdo (dismissed):

ansible_rulebook/websocket.py
```diff
@@ -215,6 +231,14 @@ async def _handle_request_workload(
             response.controller_ssl_verify = data.get("ssl_verify")
             response.controller_username = data.get("username", "")
             response.controller_password = data.get("password", "")
+
+    if non_fq_key and "filename" in file_template_vars:
+        response.variables["eda"] = {
```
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo We have to follow the guidelines set by AAP for this feature. https://docs.ansible.com/automation-controller/latest/html/userguide/credential_types.html

ansible_rulebook/websocket.py
```diff
@@ -199,6 +201,20 @@ async def _handle_request_workload(
...
                 logger.debug("wrote %s", response.project_data_file)
+        if data.get("type") == "FileContents":
+            template_key = data.get("template_key")
+            raw_data = base64.b64decode(data.get("data"))
+            keys = template_key.split(".")
+            if len(keys) == 1 and template_key == "template":
+                key = "filename"
+                non_fq_key = True
+            else:
+                key = keys[1]
```
jshimkus-rh, line 212: Does this presuppose there is, at most, two fields separated by '.'?  If the template key was `one.two.three` this would selected `two` discarding `three`.

mkanoor: @jshimkus-rh Its a quirk in the default keyname when there is only 1 key
```
{
  "file": {
      "template": "[mycloud]\ntoken={{ api_token }}"
  },
  "env": {
      "THIRD_PARTY_CLOUD_API_TOKEN": "{{ api_token }}"
  },
  "extra_vars": {
      "some_extra_var": "{{ username }}:{{ password }}"
  }
}```
From this document https://docs.ansible.com/automation-controller/latest/html/userguide/credential_types.html
It supports single template with a value of filename as well as template with different file names
In the example below the filename key would be cafile/keyfile/certfile since there is more than one.
```
file:
  template.cafile: '{{cafile}}'
  template.keyfile: '{{keyfile}}'
  template.certfile: '{{certfile}}'
```

Comment (jshimkus-rh) [reply ansible_rulebook/websocket.py:212]: This is the sort of thing that drives me nuts.  If controller is going to define the format for this sort of thing it should be providing an api that decomposes the appropriate portion of a payload containing this info.  For specialization by other software (e.g., EDA) that api should be extensible for the specialization or at least include a way of providing what to look for in that specialization.

It doesn't matter how trivial the implementation of that api might be the functionality represented by the definition should be a complete solution.

File: ansible_rulebook/app.py [hunk 3]
```diff
@@ -74,6 +78,12 @@ async def run(parsed_args: argparse.Namespace) -> None:
             raise WebSocketExchangeException(
                 "Error communicating with web socket server"
             )
+        context = decrypted_context(startup_args.variables)
+        startup_args.env_vars = substitute_variables(
+            startup_args.env_vars, context
+        )
+        for k, v in startup_args.env_vars.items():
+            os.environ[k] = str(v)
```
bzwei, line 86: @mkanoor Not only should we set it as an env, we should set it as an extra_var too.

zkayyali812 (approved):

ansible_rulebook/app.py
mkanoor: Done

ansible_rulebook/websocket.py
```diff
@@ -209,12 +226,26 @@ async def _handle_request_workload(
...
             response.controller_ssl_verify = data.get("ssl_verify")
             response.controller_username = data.get("username", "")
             response.controller_password = data.get("password", "")
+
+    if non_fq_key and "filename" in file_template_vars:
+        response.variables["eda"] = {
+            "filename": file_template_vars["filename"]
+        }
+    else:
+        response.variables["eda"] = {"filename": file_template_vars}
```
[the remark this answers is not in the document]
mkanoor: @bzwei These issues have to trapped in the eda-server side.
We need a JIRA to prevent users from adding a variable called eda and ansible.eda. Currently the eda-server allows any field name.
The same can be said about mixing template and template.key_file use case

hsong-rh (approved):

bzwei (approved):

Alex-Izquierdo (approved):

PR #713, opened 2024-09-02
chore: add support for python 3.12

Alex-Izquierdo:
Adds support for python 3.12. 
Also updates a test with a race condition in 3.12

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,6 +16,7 @@ classifiers =
 	Programming Language :: Python :: 3.9
 	Programming Language :: Python :: 3.10
 	Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
 
 [options]
 zip_safe = False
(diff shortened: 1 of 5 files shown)


hsong-rh (approved):

setup.cfg
```diff
@@ -16,6 +16,7 @@ classifiers =
 	Programming Language :: Python :: 3.9
 	Programming Language :: Python :: 3.10
 	Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
```
jshimkus-rh, line 19: This line should use leading tab as the previous lines do.

Alex-Izquierdo: I don't know where might come the mistake, but it is actually correctly indented.

Alex-Izquierdo: updated.

Alex-Izquierdo:
@hsong-rh @msmagnanijr please review again, I had to change spaces by tabs

jshimkus-rh (approved):

hsong-rh (approved):

PR #723, opened 2024-11-14
chore: pin minor version of aiohttp

Alex-Izquierdo:
Yesterday, 3.11 version of `aiohttp` was released with [breaking changes ](https://github.com/aio-libs/aiohttp/releases/tag/v3.11.0 ) which breaks `aioresponses` library used for our mocks and as consequence some of our tests are failing. 
Ref: https://github.com/pnuckowski/aioresponses/pull/262

Regardless of whether this is fixed in aioresponses, aiohttp can introduce breaking changes in minor versions, so I pin it to the latest tested version as this library is critical for the stability in some of our features.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ include_package_data = True
 packages = find:
 python_requires = >=3.9
 install_requires =
-	aiohttp >=3,<4
+	aiohttp >=3,<3.11
 	pyparsing >= 3.0,<4
 	jsonschema >=4,<5
 	jinja2 >=3,<4


Alex-Izquierdo:
> We are going to ship a 3.11.1 to restore back-compat for [pnuckowski/aioresponses#262](https://github.com/pnuckowski/aioresponses/pull/262)
> 
> I want to note that `aioresponses` could break at any patch release of `aiohttp` since third party libs that monkey patch `aiohttp` internals are not something we generally plan for upstream.
> 
> If you can use `aiohttp`'s pytest plugin you should be a lot safer off from breaking changes in test infrastructure.

Thanks for your feedback @bdraco 
We will explore the plugin that you suggest.

hsong-rh (approved):

mkanoor (approved):

Dostonbek1 (approved):

PR #728, opened 2024-12-09
chore: update changelog and prepare next devel version 

Alex-Izquierdo:


zkayyali812 (dismissed):

ttuffin (approved):

zkayyali812:
I think this test needs to be updated to pass the tests  - https://github.com/ansible/ansible-rulebook/blob/main/tests/unit/test_cli.py#L11
Now that we are suffixing `-dev`

Alex-Izquierdo:
> I think this test needs to be updated to pass the tests - https://github.com/ansible/ansible-rulebook/blob/main/tests/unit/test_cli.py#L11 Now that we are suffixing `-dev`

Good catch, @zkayyali812. Test updated

PR #733, opened 2025-02-24
fix: decryption before every substitution disabled

mkanoor:
Calling ansible-vault as a separate process is a huge overhead and takes upto 2 seconds per variable. Doing this over and over before every substitution slows doen the entire processing. Now we decrypt once at startup and keep the decrypted values in memory and reuse them during substitution.

https://issues.redhat.com/browse/AAP-40490

### DIFF
--- a/ansible_rulebook/app.py
+++ b/ansible_rulebook/app.py
@@ -78,12 +78,6 @@ async def run(parsed_args: argparse.Namespace) -> None:
             raise WebSocketExchangeException(
                 "Error communicating with web socket server"
             )
-        context = decrypted_context(startup_args.variables)
-        startup_args.env_vars = substitute_variables(
-            startup_args.env_vars, context
-        )
-        for k, v in startup_args.env_vars.items():
-            os.environ[k] = str(v)
     else:
         startup_args = StartupArgs()
         startup_args.variables = load_vars(parsed_args)
@@ -107,6 +101,12 @@ async def run(parsed_args: argparse.Namespace) -> None:
 
     validate_actions(startup_args)
     validate_variables(startup_args)
+    startup_args.variables = decrypted_context(startup_args.variables)
+    startup_args.env_vars = substitute_variables(
+        startup_args.env_vars, startup_args.variables
+    )
+    for k, v in startup_args.env_vars.items():
+        os.environ[k] = str(v)
 
     if startup_args.check_controller_connection:
         await validate_controller_params(startup_args)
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo (approved):
LGTM, but one question @mkanoor @bzwei 
Right now we are decrypting the values one per one. 
ansible-vault can decrypt a whole file, why not create a file with all the encrypted values and decrypt them in only one call of the cli?

mkanoor:
@Alex-Izquierdo currently we dont store anything in files locally when doing the decryption we pass in the fields via pexpect one at a time.

Alex-Izquierdo:
> @Alex-Izquierdo currently we dont store anything in files locally when doing the decryption we pass in the fields via pexpect one at a time.

@mkanoor That's my point. Through a file would be even more performant since we call the cli only once. 
We can create the file with the encrypted values and read the decrypted values from the stdout, if the security is your concern.

mkanoor:
@Alex-Izquierdo that might require some re-write and testing. @bzwei did we try this earlier I remember we had some solution using python packages not sure if we decrypted all variables in a single file. Not sure how that works. If there are embedded multi lines in the data. Is there a marker of when the clear text starts and when it ends.

ansible_rulebook/app.py
```diff
@@ -107,6 +101,12 @@ async def run(parsed_args: argparse.Namespace) -> None:
 
     validate_actions(startup_args)
     validate_variables(startup_args)
+    startup_args.variables = decrypted_context(startup_args.variables)
```
bzwei, line 104: Do we have to move this block from the worker condiiton?

mkanoor: @bzwei Yes so that we have the decrypted_context in one place. The default env vars is just an empty dictionary which is not used for CLI's but only used when the data is coming from EDA Server

mkanoor:
> If we want to call the decrypt cli only once, we may assemble all secret variables into a json string and encrypt it at eda-server side and decrypt and parse json at the engine side.

@bzwei The problem with that is "Data in Transit" is not encrypted

bzwei:
> > If we want to call the decrypt cli only once, we may assemble all secret variables into a json string and encrypt it at eda-server side and decrypt and parse json at the engine side.
> 
> @bzwei The problem with that is "Data in Transit" is not encrypted

why not encrypted? I am proposing the server encrypts the json string and sends to client. So data in transit is encrypted. Then client decrypts, parses the decrypted json, and inserts sensitive variables in memory.

Alex-Izquierdo:
@mkanoor @bzwei Anyway, I think we can merge this to improve the thing in the meantime while we are working on a better solution. WDYT?

bzwei (approved):

PR #737, opened 2025-02-28
docs: update the rulebook name

reidliu41:
The below command is use the rulebook name, it should be better to mention:
```
root@ansible-rulebook:/root# ansible-rulebook --rulebook webhook-example.yml  << -i inventory.yml --verbose
```


ttuffin (approved):

reidliu41:
may i know this one can merge?

PR #742, opened 2025-02-28
chore: rebase stable 1.1 onto main

Alex-Izquierdo:
This is a rebase from main with conflict solving to keep the parents inline between stable-1.1 and main after diverge in  https://github.com/ansible/ansible-rulebook/pull/726 and https://github.com/ansible/ansible-rulebook/pull/727  
It includes all the missing commits from main
From now, any later rebase should not have conflicts


Alex-Izquierdo:
it is wrong, working on a solution

PR #757, opened 2025-03-11
ci: upgrade python gh action

Alex-Izquierdo:


zkayyali812 (approved):

ttuffin (approved):
Out of curiosity what is the reason for the upgrade? Are there new features we want to leverage?

Alex-Izquierdo:
> Out of curiosity what is the reason for the upgrade? Are there new features we want to leverage?

Anything particular. Some bugfixes and upgraded pip

PR #762, opened 2025-03-14
ci: add release workflow AAP-41583

Alex-Izquierdo:
Add a workflow to automate the build and publishing of a new release. 
Jira: https://issues.redhat.com/browse/AAP-41583

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,60 @@
+name: Release Workflow
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  check_permissions:
+    runs-on: ubuntu-latest
+    if: github.repository == 'ansible/ansible-rulebook'
+    steps:
+      - name: Check if user is a maintainer
+        env:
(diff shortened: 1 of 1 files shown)


zkayyali812 (dismissed):
Approving, but wondering if we need the `check_permissions` part?
Wouldnt maintainers be the only ones that can trigger this anyway since they would be the only ones that can publish release?

Alex-Izquierdo:
@zkayyali812 If I'm not wrong, the release can be created with write permissions. Are through branch/tag protections how we would be enforcing the maintainers. But the release can be created for an existing tag, so I think it is a good double check.

Alex-Izquierdo:
@zkayyali812 another option could be to protect ALL the tags, not only the version ones, but I think that can be less flexible, we could want a tag for other reasons rather than for releases (for example I created a tag in eda-server to set a point in time of the dispatcher). 
In a second thought I think we can restrict by tag pattern v*, then we don't need the check since we rely only on the tag protection.

zkayyali812 (approved):

PR #763, opened 2025-03-15
refactor: extract create context to public func

reidliu41:
- Since both function very familiar, so move to util.


ttuffin (approved):

bzwei (approved):

ttuffin (approved):
Thanks for rebasing @reidliu41. Just need one more review @ansible/eda-maintainers

ptoscano (approved):
Simple refactoring -- thanks!

PR #764, opened 2025-03-17
test: ensure multiprocess tests don't collision ports

Alex-Izquierdo:
We can have race conditions in some tests when running pytest with xdist (multiprocesses) with errors because the port is already in use. This ensures the port is different by adding the cpu core of the process.

### DIFF
--- a/tests/e2e/test_run_module_output.py
+++ b/tests/e2e/test_run_module_output.py
@@ -1,6 +1,7 @@
 """
 Module with tests for websockets
 """
+
 import asyncio
 import logging
 from functools import partial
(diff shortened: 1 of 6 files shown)


ttuffin (dismissed):

tests/e2e/test_run_module_output.py
```diff
@@ -26,7 +27,7 @@ async def test_run_module_output():
     host = "127.0.0.1"
     endpoint = "/api/ws2"
     proc_id = "42"
-    port = 31416
+    port = utils.get_safe_port()
```
zkayyali812, line 30: Is this on purpose, or are we missing a parameter here?

Alex-Izquierdo: @zkayyali812 Sorry by the confusion. I noticed that my previous approach was wrong so I've changed it. Now is not needed a parameter, since the port is chosen by the helper. 
Please, review again.

zkayyali812 (approved):

ttuffin (approved):

PR #776, opened 2025-04-10
chore: pin ansible-core to 2.16

ttuffin:
In light of upcoming changes in ansible-core, pinning the version to 2.16 while we investigate the impact.

[AAP-43723](https://issues.redhat.com/browse/AAP-43723)

https://github.com/ansible/ansible/pull/84621

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,10 +5,14 @@ ARG APP_DIR=${APP_DIR:-/app}
 ARG SETUPTOOLS_SCM_PRETEND_VERSION
 ARG DEVEL_COLLECTION_LIBRARY=0
 ARG DEVEL_COLLECTION_REPO=git+https://github.com/ansible/event-driven-ansible.git
+ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.16}
+ARG PYTHON_VER="3.11"
+ARG PYTHON_BIN="python${PYTHON_VER}"
+ARG PIP_BIN="pip${PYTHON_VER}"
 
 ENV JAVA_HOME=/usr/lib/jvm/java-17-openjdk
 ENV PATH="$APP_DIR/.local/bin:$PATH"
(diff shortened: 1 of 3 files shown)


minimal-decision-environment.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
     - dpath
     - ansible-rulebook
   ansible_core:
-    package_pip: ansible-core==2.14.4
+    package_pip: ansible-core==2.16
```
ttuffin, line 22: No idea if this file is still used.

Alex-Izquierdo: I think it is just as reference for DE building.

requirements_test.txt
```diff
@@ -3,7 +3,7 @@ docopt
 psutil
 requests
 ansible
-ansible-core
+ansible-core==2.16
```
Alex-Izquierdo, line 6: you can choose the version of ansible depending on the version of python
```
ansible-core==2.16; python_version > "3.11"
ansible-core==2.14; python_version < "3.11"
```

Dockerfile
```diff
@@ -31,7 +31,7 @@ WORKDIR $APP_DIR
 COPY --chown=${USER_ID}:0 . $WORKDIR
 
 RUN pip install -U pip \
-    && pip install ansible-core \
+    && pip install ansible-core==2.16 \
```
Alex-Izquierdo, line 34: Better a build arg with this version as default.

Alex-Izquierdo (approved):

requirements_test.txt
```diff
@@ -3,7 +3,8 @@ docopt
 psutil
 requests
 ansible
-ansible-core
+ansible-core==2.16; python_version > "3.11"
```
bzwei, line 6: will it pick ansible-core2.16 if python_version = 3.11?

ttuffin: good catch. fixed.

Alex-Izquierdo (approved):

zkayyali812 (approved):

PR #778, opened 2025-04-10
fix: upgrade to the latest websockets

mkanoor:
Tehcnical debt, upgrading to the latest version of websockets

https://websockets.readthedocs.io/en/stable/howto/upgrade.html

https://issues.redhat.com/browse/AAP-44207

The new websockets RPM is being tracked in https://issues.redhat.com/browse/AAP-44222

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ install_requires =
 	dpath >= 2.1.4,<3
 	janus >=1,<2
 	ansible-runner >=2,<3
-	websockets >=10.4,<12
+        websockets >=15,<15.1
 	drools_jpy == 0.3.10
 	watchdog >=3,<7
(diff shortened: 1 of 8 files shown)


Alex-Izquierdo (changes requested):
we need extra coordination for this change. 
cc @zkayyali812

setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
 	dpath >= 2.1.4,<3
 	janus >=1,<2
 	ansible-runner >=2,<3
-	websockets >=10.4,<12
+	websockets ==15
```
Alex-Izquierdo: Can we set `websockets >=15,<15.1` to be aligned and protected on downstream builds?

ptoscano (approved):

PR #783, opened 2025-04-23
chore: fix ansible-core version pin

ttuffin:
Update the version pins to use latest minor version. This is to resolve the dependabot alert that popped up after the last PR to pin the version: https://github.com/ansible/ansible-rulebook/security/dependabot/10


requirements_test.txt
```diff
@@ -3,8 +3,8 @@ docopt
 psutil
 requests
 ansible
-ansible-core==2.16; python_version >= "3.11"
-ansible-core==2.15; python_version < "3.11"
+ansible-core>=2.16,<2.17; python_version >= "3.11"
+ansible-core>=2.15,<2.16; python_version < "3.11"
```
ptoscano, line 7: I think that you can simplify them using `~=`, e.g.:
```
ansible-core~=2.16.0; python_version >= "3.11"
```
Reference: https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers

minimal-decision-environment.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
     - dpath
     - ansible-rulebook
   ansible_core:
-    package_pip: ansible-core==2.16
+    package_pip: ansible-core==2.16.14
```
ptoscano, line 22: ditto

Dostonbek1 (approved):

PR #792, opened 2025-05-06
chore: adjust CLI log levels

ptoscano:
Align logger levels in the CLI to match our definitions in https://issues.redhat.com/browse/AAP-41839

Included some minor cosmetic changes to some of the log messages to disambiguate them or make them more clear.


ansible_rulebook/cli.py
```diff
@@ -353,8 +353,8 @@ def main(args: List[str] = None) -> int:
         asyncio.run(app.run(args))
     except KeyboardInterrupt:
         return 0
-    except Exception as err:
-        logger.error("Terminating %s", str(err))
+    except Exception:
```
Alex-Izquierdo, line 356: I think the current code was done in purpose to prevent exposing the stack trace. Correct me If I'm wrong @mkanoor but I think we want the current approach `logger.error("Terminating %s", str(err))`

tests/e2e/test_actions.py
```diff
@@ -222,7 +222,7 @@ def test_run_playbook(update_environment):
     )
 
     assert result.returncode == 0
-    assert not result.stderr
+    assert "Previous run_playbook failed" in result.stderr
```
Alex-Izquierdo, line 225: Why a test that expects rc 0 and empty stderr now it expects has an error message? That does not make sense to me.

ansible_rulebook/app.py
```diff
@@ -89,7 +89,7 @@ async def run(parsed_args: argparse.Namespace) -> None:
 
     if parsed_args.worker and parsed_args.websocket_url and parsed_args.id:
         startup_logging(logger)
-        logger.info("Starting worker mode")
+        logger.debug("Starting worker mode")
```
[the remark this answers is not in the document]
ptoscano: Ah ok, I wasn't entirely sure about it. Switching it back to info, thanks!

ansible_rulebook/cli.py
ptoscano: That's what `logging.exception()` already does:
https://docs.python.org/3/library/logging.html#logging.Logger.exception

tests/e2e/test_actions.py
ptoscano: The test in practice expects 1 retry attempt, see the following assertion later on:
https://github.com/ansible/ansible-rulebook/blob/13af68dea8abbfddb4b792947f8328dad35c934b/tests/e2e/test_actions.py#L236-L239
The message is now printed to stderr because I changed that message from INFO to WARNING:
```diff
diff --git a/ansible_rulebook/action/run_job_template.py b/ansible_rulebook/action/run_job_template.py
index 8415bc3..f7d8825 100644
--- a/ansible_rulebook/action/run_job_template.py
+++ b/ansible_rulebook/action/run_job_template.py
@@ -90,7 +90,7 @@ class RunJobTemplate:
                 if i > 0:
                     if delay > 0:
                         await asyncio.sleep(delay)
-                    logger.info(
+                    logger.warning(
                         "Previous run_job_template failed. Retry %d of %d",
                         i,
                         retries,
```
and log messages with severity WARNING and higher are logged to stderr:
https://github.com/ansible/ansible-rulebook/blob/13af68dea8abbfddb4b792947f8328dad35c934b/ansible_rulebook/cli.py#L261-L274

ansible_rulebook/cli.py
Alex-Izquierdo: exception prints the traceback, which is want I believe we want to avoid. 
```
import logging

# Configure the logger
logging.basicConfig(
    level=logging.ERROR,
    format="%(levelname)s:%(name)s:%(message)s"
)

if __name__ == "__main__":
    try:
        1 / 0
    except ZeroDivisionError as e:
        logger.error(f"{e}, no exception info is shown")

    try:
        1 / 0
    except ZeroDivisionError:
        logger.exception("An exception occurred here.")
```

tests/e2e/test_actions.py
Alex-Izquierdo: ok, thanks for the clarification

ansible_rulebook/cli.py
ptoscano: Right :facepalm: switching it back

ansible_rulebook/action/run_job_template.py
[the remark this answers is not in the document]
B-Whitt: addressed ... https://redhat-internal.slack.com/archives/C053CJ4JFP1/p1747150697842539

B-Whitt (approved):
looks good

Alex-Izquierdo (approved):

PR #793, opened 2025-05-13
dump main on stable1.1

bzwei:
This is a routine backport to stable-1.1 from main. No plan to make a new release to pypi yet.


Alex-Izquierdo (approved):
I have renamed the PR assuming you agree, @bzwei 
We don't bump manually anything in the code since it is automated and even you even you say there is no plan to release yet. 
"Release xxx" as PR title would not make sense and is misleading.

PR #794, opened 2025-05-14
feat: [AAP-45812] Accept ansible tagged dicts in fact cache from ansible-core 2.19

hsong-rh:
https://issues.redhat.com/browse/AAP-45812

The ansible-core 2.19 adds the data tagging features, which change the fact cache of artifacts into a new format with taggings. Ansible rulebook needs to handle this new format.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -3,7 +3,7 @@ docopt
 psutil
 requests
 ansible
-ansible-core~=2.16.0; python_version >= "3.11"
+ansible-core~=2.19.0b4; python_version >= "3.11"
 ansible-core~=2.15.0; python_version < "3.11"
 ansible-runner
 jmespath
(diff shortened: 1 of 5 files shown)


requirements_test.txt
```diff
@@ -3,7 +3,7 @@ docopt
 psutil
 requests
 ansible
-ansible-core~=2.16.0; python_version >= "3.11"
+ansible-core~=2.19.0b4; python_version >= "3.11"
```
Alex-Izquierdo, line 6: we can not merge this. If it is some thing temporal to later link to passed tests, ok. But even in that case could not be the best approach. Instead, we can update the CI pipeline and add a new item in the matrix, so we can install ansible 2.19 in a different run and be able to test both versions of ansible for this change and any other afterwards.

PR #796, opened 2025-05-16
[DON'T MERGE] test ansible-core 2.19

hsong-rh:


Alex-Izquierdo:
We are going to want ansible 2.19 into the matrix test when it is released.

PR #800, opened 2025-06-03
chore: add dependabot config for github actions

ptoscano:


Alex-Izquierdo (changes requested):
I would like to discuss this

Alex-Izquierdo (approved):

PR #810, opened 2025-06-12
fix: gather_facts requires inventory

mkanoor:
gather_facts requires inventory, when running as CLI it always has inventory but when run as Activation the inventory is missing.

Ignore the gather_facts if no inventory is provided and print a warning message.

https://issues.redhat.com/browse/AAP-47070

### DIFF
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -286,7 +286,13 @@ async def run_rulesets(
     rulesets = {}
     for ruleset, _ in ruleset_queues:
         if ruleset.gather_facts and not hosts_facts:
-            hosts_facts = collect_ansible_facts(inventory)
+            if inventory:
+                hosts_facts = collect_ansible_facts(inventory)
+            else:
+                logger.warning(
+                    "Ignoring gather_facts, since it requires inventory"
+                )
+
         ruleset_names.append(ruleset.name)
         rulesets[ruleset.name] = ruleset
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add support for playbook that has set_stats in it
 ### Fixed
 - Fix a bug in the processing of a playbook that has a set_fact in it
+- Fix a bug in gather_facts. It needs inventory, else its ignored
 
 
 ## [1.1.6] - 20205-04-25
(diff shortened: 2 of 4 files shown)


mkanoor:
@Alex-Izquierdo I forgot to include the old files and just added the new one initially.

ptoscano (approved):

ansible_rulebook/engine.py
```diff
@@ -286,7 +286,13 @@ async def run_rulesets(
     rulesets = {}
     for ruleset, _ in ruleset_queues:
         if ruleset.gather_facts and not hosts_facts:
-            hosts_facts = collect_ansible_facts(inventory)
+            if inventory:
+                hosts_facts = collect_ansible_facts(inventory)
+            else:
+                logger.warning(
```
Alex-Izquierdo, line 292: is the warning message appropriate here? A warning message is for scenarios where something is or might be wrong. Here we would be speaking of a expected situation, since we don't want to run gather facts in absence of inventory. 

If we want to inform to the user that gather fact is not executed due to the absence of inventory,  I feel it should be more INFO level, otherwise we would be printing a warning message in practically all the cases where ansible-rulebook runs in AAP creating unnecessary noise.

Alex-Izquierdo (commented):
LGTM, but asked a question about the log level.

ansible_rulebook/engine.py
ptoscano: >  otherwise we would be printing a warning message in practically all the cases where ansible-rulebook runs in AAP creating unnecessary noise.

The warning is printed only if `gather_facts` is `true`: since the default is `false`, it means that the user actually changed that in the rulebook. I think that it is sensible to print the warning about a setup/configuration that it is not correct.

mkanoor: The use case here is a customer takes the rulebook that he runs with a CLI and uses it directly as part of the Activation. gather_facts is an experimental feature that was never fully flushed out and the presence of it currently fails the Activation and the way out for the customer is to remove it. This PR prints out a warning message to let the user know that the rulebook has gather_facts and the behaviour that they were expecting when they ran as CLI is not going to match when they run it as part of Activation. It helps in those situations where a customer says it works when I run it part of the CLI but fails when I run it as part of the Activation.

Alex-Izquierdo: Ok, you are right, thanks for the clarification.

PR #820, opened 2025-07-05
feat: event splitter filter to support array of events

mkanoor:
When we get in an event or get get back data from a filter check if its a list so we can inject them individually into Drools

We have added a new builtin filter called
* eda.builtin.event_splitter

This filter splits an event into an array of events There are 1 required and 3 optional parameters

* splitter_key (Required: string)
* attributes_key_map (Optional: Dictionary)
* extras (Optional: Dictionary)
* raise_error (bool: Optional default false)

e.g.
```
        - eda.builtin.event_splitter:
            splitter_key: bundle.employees
            attributes_key_map:
              employer: bundle.employer
              owner: bundle.owner
            extras:
              city: Bedrock
```

https://issues.redhat.com/browse/AAPRFE-1772
https://issues.redhat.com/browse/AAP-49670

One of the main driving forces behind this feature is the multiple events that get bundled in Promethus and Big Panda

For Prometheus we could have the following sources snippet
```
- ansible.eda.alertmanager:
    host: 0.0.0.0
    port: 8000
    filters:
        - eda.builtin.event_splitter:
            splitter_key: alerts 
            attributes_key_map:
              header: header
              hosts: labels.instance
  ```
This is based on this source [plugin](https://github.com/ansible/event-driven-ansible/blob/5236ec58711b7623e9df4f8215dc832e16100d11/extensions/eda/plugins/event_source/alertmanager.py#L90) logic and allows us to swap out ansible.eda.alertmanager with an EventStream since an EventStream will preserve the filter  

For [Big Panta](https://github.com/wwt/eda_collection/blob/215e22c9f820b8a8538318055d063564ddf02151/extensions/eda/plugins/event_source/bigpanda.py#L101) we could use something like this
```
- bigpanda.......:
    host: 0.0.0.0
    port: 8000
    filters:
        - eda.builtin.event_splitter:
            splitter_key: incident.alerts
            attributes_key_map:
                  id: incident.id
                  active: incident.active
                  severity: incident.severity
                  status: incident.status
                  environments: incident.environments
 ```

### DIFF
--- a/ansible_rulebook/event_filter/event_splitter.py
+++ b/ansible_rulebook/event_filter/event_splitter.py
@@ -1,0 +1,138 @@
+#  Copyright 2025 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+import copy
+import multiprocessing as mp
+from typing import Any, Optional
+
+import dpath
+
+DOCUMENTATION = r"""
+---
+short_description: Split an event payload to multiple events based on a key
+description: |
+  - An event filter that allows users to split an event payload which has
+    several events bundled inside.
+  - For instance, Prometheus and Big Panda send in a collection of events
+    If we know the key name that stores this array of events we can have
+    the event filter slit the payload into individual events.
+    If the splitter key doesn't exist the payload is left intact
+  - Optionally we can include other attributes from the payload into every
+    event
+  - Optionally we can add additional static data into each event
+options:
+  splitter_key:
+    description:
+      - This mandatory field specifies the key in the payload
+      - that has an array (list) of events
+      - You can use dotted notation to specify the key path.
+      - Example splitter_key: topkey1.key2.key3
+    type: str
+  attributes_key_map:
+    description:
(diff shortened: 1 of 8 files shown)


ansible_rulebook/event_filter/event_splitter.py
```diff
@@ -0,0 +1,125 @@
...
+    splitter_key: str,
+    attributes_key_map: Optional[dict[str, Any]] = None,
+    extras: Optional[dict[str, Any]] = None,
+) -> list[dict[str, Any]]:
+    """Split event into an array of events."""
+    logger = mp.get_logger()
+    try:
+        event_array = dpath.get(event, splitter_key, separator=".")
+    except KeyError:
+        logger.warning(
```
Alex-Izquierdo, line 108: Should we have "raise_error" option, like we have in hosts to meta? 
https://github.com/ansible/event-driven-ansible/blob/main/extensions/eda/plugins/event_filter/insert_hosts_to_meta.py#L36

mkanoor: Done

ansible_rulebook/event_filter/event_splitter.py
```diff
@@ -0,0 +1,138 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+import copy
+import multiprocessing as mp
+from typing import Any, Optional
+
+import dpath
+
+DOCUMENTATION = r"""
```
Alex-Izquierdo, line 20: I see a problem with the documentation. Nothing will render this content. Right now we only have support from [galaxy importer ](https://github.com/ansible/galaxy-importer/pull/330) for collections in ansible galaxy and automation hub. 
I think we should add documentation in https://github.com/ansible/ansible-rulebook/blob/main/docs/filters.rst

In the meantime I can create a ticket to explore a way to implement generation in our own portal.

Alex-Izquierdo (approved):

PR #821, opened 2025-07-16
feat: event splitter filter to support array of events (#820)

mkanoor:
When we get in an event or get get back data from a filter check if its a list so we can inject them individually into Drools

We have added a new builtin filter called
* eda.builtin.event_splitter

This filter splits an event into an array of events There are 1 required and 3 optional parameters

* splitter_key (Required: string)
* attributes_key_map (Optional: Dictionary)
* extras (Optional: Dictionary)
* raise_error (bool: Optional default false)

e.g.
```
        - eda.builtin.event_splitter:
            splitter_key: bundle.employees
            attributes_key_map:
              employer: bundle.employer
              owner: bundle.owner
            extras:
              city: Bedrock
```

https://issues.redhat.com/browse/AAP-49750

Backport related to PR #820 
https://issues.redhat.com/browse/AAPRFE-1772
https://issues.redhat.com/browse/AAP-49670

One of the main driving forces behind this feature is the multiple events that get bundled in Promethus and Big Panda

For Prometheus we could have the following sources snippet

```
- ansible.eda.alertmanager:
    host: 0.0.0.0
    port: 8000
    filters:
        - eda.builtin.event_splitter:
            splitter_key: alerts
            attributes_key_map:
              header: header
              hosts: labels.instance
```

```
- bigpanda.......:
    host: 0.0.0.0
    port: 8000
    filters:
        - eda.builtin.event_splitter:
            splitter_key: incident.alerts
            attributes_key_map:
                  id: incident.id
                  active: incident.active
                  severity: incident.severity
                  status: incident.status
                  environments: incident.environments
 ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 ## [Unreleased]
 ### Changed
 ### Added
+- Add support for eda.builtin.event_splitter filter
+### Fixed
+
+
+## [1.1.7] - 2025-06-23
+### Changed
+### Added
 - Add support for playbook that has set_stats in it
 ### Fixed
 - Fix a bug in the processing of a playbook that has a set_fact in it
--- a/ansible_rulebook/engine.py
+++ b/ansible_rulebook/engine.py
@@ -84,10 +84,23 @@ class FilteredQueue:
         self.queue = queue
 
     async def put(self, data):
+        if not isinstance(data, list):
+            data = [data]
+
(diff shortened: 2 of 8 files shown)


Alex-Izquierdo (changes requested):
I might be missing recent decisions, but as far as I know, there are no more releases of 1.1 scheduled for new features, only defects are planned, so this feature should not be backported and only delivered as for the new 1.2 version.

PR #828, opened 2025-08-13
feat: add lock support for certain actions

mkanoor:
When running the rulebook with the execution strategy set to parallel you can optionally specify a lock field to ensure sequential execution of a specific action.
The lock is currently supported for
* run_playbook
* run_module
* run_job_template
* run_workflow_template

There is no e2e test for this we will be adding an ATF test for this since it would be tested against the controller in the ATF.

https://issues.redhat.com/browse/AAP-51458

### DIFF
--- a/ansible_rulebook/rule_set_runner.py
+++ b/ansible_rulebook/rule_set_runner.py
@@ -16,6 +16,7 @@ import asyncio
 import gc
 import logging
 import uuid
+from collections import defaultdict
 from pprint import pformat
 from types import MappingProxyType
 from typing import Dict, List, Optional, Union, cast
@@ -108,6 +109,7 @@ class RuleSetRunner:
         self.broadcast_method = broadcast_method
         self.event_counter = 0
         self.display = terminal.Display()
+        self.locks = defaultdict(asyncio.Lock)
 
     async def run_ruleset(self):
         tasks = []
@@ -284,6 +286,34 @@ class RuleSetRunner:
             if not self.action_loop_task.done():
                 self.action_loop_task.cancel()
 
+    async def _run_action_with_lock(
+        self,
+        lock_name: str,
+        action: str,
+        metadata: Metadata,
+        control: Control,
+        action_args: dict,
+    ):
+        """
+        Acquire the lock for the specified resource, access the resource,
+        and release the lock.
+        """
+        async with self.locks[lock_name]:
+            logger.debug(
+                f"Acquired lock: {lock_name} for action: {action}, "
+                f"rule: {metadata.rule}"
+            )
+            await ACTION_CLASSES[action](
+                metadata,
+                control,
+                **action_args,
+            )()
+
+        logger.debug(
+            f"Released lock: {lock_name} for action: {action}, "
+            f"rule: {metadata.rule}"
+        )
+
     async def _drain_actionplan_queue(self):
         logger.info("Waiting for actions on events from %s", self.name)
         try:
@@ -316,6 +346,7 @@ class RuleSetRunner:
                     == ExecutionStrategy.SEQUENTIAL
                 ):
                     await task
+
         except asyncio.CancelledError:
             logger.debug(
                 "Action Plan Task Cancelled for ruleset %s", self.name
@@ -451,11 +482,25 @@ class RuleSetRunner:
                     project_data_file=self.project_data_file,
                 )
 
-                await ACTION_CLASSES[action](
-                    metadata,
-                    control,
-                    **action_args,
-                )()
+                lock = action_args.get("lock", None)
+                if (
+                    self.rule_set.execution_strategy
+                    == ExecutionStrategy.PARALLEL
+                    and lock
+                ):
+                    await self._run_action_with_lock(
+                        lock,
+                        action,
+                        metadata,
+                        control,
+                        action_args,
+                    )
+                else:
+                    await ACTION_CLASSES[action](
+                        metadata,
+                        control,
+                        **action_args,
+                    )()
 
             except KeyError as e:
                 logger.error(
--- a/docs/actions.rst
+++ b/docs/actions.rst
@@ -67,6 +67,9 @@ Run an Ansible playbook.
    * - copy_files
      - Boolean, copy the local playbook file to the ansible-runner project directory, this is not needed if you are running a playbook from an ansible collection.
      - No
+   * - lock
+     - An optional string based lock ensures sequential execution of this action when execution strategy is set to parallel. It can also be a string field from the event payload. The locks are per ruleset, if a lock is in place all actions that use the same lock will wait till the earlier action has completed.
+     - No
 
 
 run_module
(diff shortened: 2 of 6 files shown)


ansible_rulebook/rule_set_runner.py
```diff
@@ -284,6 +286,29 @@ def _handle_action_completion(self, task):
...
+        async with self.locks[
+            lock_name
+        ]:  # Acquire lock for the specific resource
+            logger.debug(f"Acquired lock for resource: {lock_name}")
+            await ACTION_CLASSES[action](
+                metadata,
+                control,
+                **action_args,
+            )()
+            logger.debug(f"Released lock for resource: {lock_name}")
```
Alex-Izquierdo: This line should be out of the lock.

docs/actions.rst
```diff
@@ -67,6 +67,9 @@ Run an Ansible playbook.
    * - copy_files
      - Boolean, copy the local playbook file to the ansible-runner project directory, this is not needed if you are running a playbook from an ansible collection.
      - No
+   * - lock
```
Alex-Izquierdo, line 70: Why not enable it also for run_module?

docs/actions.rst
```diff
@@ -178,6 +181,9 @@ Run a job template.
    * - job_args
      - Additional arguments sent to the job template launch API. Any answers to the survey and other extra vars should be set in nested key extra_vars. Event(s) and fact(s) will be automatically included in extra_vars too. Optionally if the job_args includes an attribute called limit it can be used to over write the limit set from the event payload.
      - No
+   * - lock
```
Alex-Izquierdo, line 187: This seems an important feature that many users would use. I feel we should have some more extended documentation with examples in addition to the parameter in the action table.

ansible_rulebook/rule_set_runner.py
```diff
@@ -284,6 +286,29 @@ def _handle_action_completion(self, task):
...
+        metadata: Metadata,
+        control: Control,
+        action_args: dict,
+    ):
+        """
+        Acquire the lock for the specified resource, access the resource,
+        and release the lock.
+        """
+        async with self.locks[
+            lock_name
```
Alex-Izquierdo: The current implementation assumes the lock name is unique per ruleset, if multiple actions define the same name it would have unexpected behaviors.  This is not explained in the docs. 

Do we want to support the same lock across multiple actions? I feel we should not, in that case the application should validate it.

docs/actions.rst
Alex-Izquierdo: I think the documentation should also remark/clarify that this option does not discard subsequent triggers and subsequent actions would be enqueued. I feel users could understand that an event that matches the condition while the action is running is going to be discarded.

ansible_rulebook/rule_set_runner.py
Alex-Izquierdo: When I say multiple actions, I mean multiple rules, because I understand the scope of the lock is per rule (a rule certainly could have multiple actions, in that case, would make sense)

ansible_rulebook/rule_set_runner.py
```diff
@@ -284,6 +286,29 @@ def _handle_action_completion(self, task):
...
+        ]:  # Acquire lock for the specific resource
+            logger.debug(f"Acquired lock for resource: {lock_name}")
```
Alex-Izquierdo: I think here we should also indicate the action and rule, specially if we are going to support the same lock across multiple actions/rules, otherwise I don't see how this message could be useful.

docs/actions.rst
```diff
@@ -516,3 +528,26 @@ none
...
+| to run other jobs in other data centers. All jobs with the same lock will be run sequentially, the user has the
+| choice of what they want to use as a lock, it could be any string attribute from the event payload or it could be
+| a static string, where you want all other actions to run parallel but others to run sequentially. e.g if you have
+| job to run a backup of a database.
+Example:
+    .. code-block:: yaml
+
+        name: sample rule
+        condition: event.level == "error"
+        action:
```
Alex-Izquierdo, line 552: How it works when you use the same lock across multiple actions in the same rule?
How are internally enqueued? If its a LIFO we could get a situation where some actions were never executed in favor of new ones. Probably is worth to test it.

tests/examples/96_job_template_with_lock.yml
```diff
@@ -0,0 +1,32 @@
+---
```
Alex-Izquierdo, line 1: We should have too tests when the lock is reused across multiple rules/actions. WDTY?

tests/test_examples.py
```diff
@@ -2639,3 +2639,132 @@ async def test_93_event_splitter():
...
+            "jt1": "test1",
+            "jt2": "test2",
+            "jt3": "test3",
+        },
+        ["test1", "test2", "test3"],
+    ),
+]
+
+
+@pytest.mark.parametrize("response, my_vars, expected_order", RULEBOOK_DATA)
```
Alex-Izquierdo: This parameterization is hard to read. 
I suggest the following refactor to make it more readable. 
(AI supported)

```
BASE_RESPONSE = {
    "test1": {
        "sleep": 5,
        "result": dict(
            status="successful",
            id=945,
            created="dummy",
            artifacts=dict(a=1),
        ),
    },
    "test2": {
        "sleep": 0,
        "result": dict(
            status="successful",
            id=946,
            created="dummy",
            artifacts=dict(a=1),
        ),
    },
    "test3": {
        "sleep": 0,
        "result": dict(
            status="successful",
            id=947,
            created="dummy",
            artifacts=dict(a=1),
        ),
    },
}

@pytest.mark.parametrize(
    "response, my_vars, expected_order",
    [
        pytest.param(
            BASE_RESPONSE,
            {
                "lock1": "red",
                "lock2": "red",
                "lock3": "green",
                "jt1": "test1",
                "jt2": "test2",
                "jt3": "test3",
            },
            ["test1", "test3", "test2"],
            id="two_red_locks_order_t1_t3_t2"
        ),
        pytest.param(
            BASE_RESPONSE,
            {
                "lock1": "red",
                "lock2": "blue",
                "lock3": "green",
                "jt1": "test1",
                "jt2": "test2",
                "jt3": "test3",
            },
            ["test1", "test2", "test3"],
            id="mixed_locks_order_t1_t2_t3"
        ),
    ]
)
```

docs/actions.rst
mkanoor: @Alex-Izquierdo asyncio.lock is a mutex its not a queue. So which ever task gets the CPU cycle will run. There is no guarantee like a queue. When you have multiple actions for a rule they still run sequentially that behavior has not changed.

Alex-Izquierdo: I can confirm that internally uses  `deque` and behaves as FIFO queue. 
https://github.com/python/cpython/blob/3.12/Lib/asyncio/locks.py#L146

Alex-Izquierdo (approved):

PR #830, opened 2025-08-14
feat: labels can be optionally added to controller actions

mkanoor:
If Labels in Job Template/Workflow Template have Prompt on launch enabled a customer can add labels to the jobs they run.

The following actions support labels

 * run_job_template
 * run_workflow_template

By default ansible-rulebook uses a label called
"Activated by Event-Driven Ansible"

https://issues.redhat.com/browse/AAP-51460

Screenshot 2025-08-14 at 11 15 17 AM

Screenshot 2025-08-14 at 10 53 13 AM

### DIFF
--- a/ansible_rulebook/job_template_runner.py
+++ b/ansible_rulebook/job_template_runner.py
@@ -18,15 +18,18 @@ import logging
 import os
 import ssl
 from functools import cached_property
-from typing import Union
+from http import HTTPStatus
+from typing import Optional, Union
 from urllib.parse import urljoin, urlparse
 
 import aiohttp
 import dpath
 
 from ansible_rulebook import util
+from ansible_rulebook.conf import settings
 from ansible_rulebook.exception import (
     ControllerApiException,
+    ControllerObjectCreateException,
     JobTemplateNotFoundException,
     WorkflowJobTemplateNotFoundException,
 )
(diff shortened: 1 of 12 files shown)


ansible_rulebook/job_template_runner.py
```diff
@@ -203,6 +238,21 @@ async def run_workflow_job_template(
...
+            job_params["labels"] = job_labels
+        else:
+            if labels:
+                logger.warning(
+                    (
+                        "Job template %s does not accept labels, please "
+                        "enable Prompt on launch for Labels"
+                    ),
+                    name,
+                )
```
Alex-Izquierdo: Can't we deduplicate this code?

ansible_rulebook/job_template_runner.py
```diff
@@ -48,19 +57,25 @@ def __init__(
...
         self.password = password
         self.verify_ssl = verify_ssl
-        self.refresh_delay = float(
-            os.environ.get("EDA_JOB_TEMPLATE_REFRESH_DELAY", 10.0)
-        )
+        if isinstance(refresh_delay, float):
+            self.refresh_delay = refresh_delay
+        else:
+            self.refresh_delay = float(
+                os.environ.get("EDA_JOB_TEMPLATE_REFRESH_DELAY", 10.0)
```
Alex-Izquierdo: Here we can raise ValueError if the environment variable has a non casteable value.

My recommendation for robustness, performance and cleaner code is to not use environment variables directly in the client but to use the [Settings object](https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/conf.py#L21) to set a validated, centralized value and then use it from there.

ansible_rulebook/job_template_runner.py
```diff
@@ -234,10 +284,93 @@ async def _launch(self, job_params: dict, url: str) -> dict:
...
+                body = json.loads(await post_response.text())
+                post_response.raise_for_status()
+                return body, False
+        except aiohttp.ClientResponseError as e:
+            # If the object got created by another process, do a retry
+            if e.status == HTTPStatus.BAD_REQUEST and "already exists" in str(
+                body
+            ):
+                return None, True
+            logger.warning(
```
Alex-Izquierdo: Why this is just a warning? If this method wants to be reusable I think it should raise exception and the caller determines if must be ignored or not like you want to do for the specific case of labels.

ansible_rulebook/job_template_runner.py
mkanoor: @Alex-Izquierdo I will remove all of this code, it adds a 10 second delay to tests. We can do this change in a separate PR.

ansible_rulebook/job_template_runner.py
mkanoor: @Alex-Izquierdo I didn't want label creation issues to stall an action so its a warning for now.

Alex-Izquierdo: Yes, I understand that. My point is that you are enforcing it in `_create_obj` method seems to be a more general/reusable method. For consistency of the client, I think this method should raise exception, and  `_get_or_create_label` is who should ignore it and print the warning. WTDY?

hsong-rh (approved):

ansible_rulebook/job_template_runner.py
mkanoor: @Alex-Izquierdo I have added an exception ControllerCreateObjectException to catch it and suppress it for labels.

PR #911, opened 2026-04-17
docs: enhance the any condition examples

mkanoor:
The any condition is a bit confusing since it is treated as a multi condition rule and only 1 of those conditions has to match for the rule to trigger. Multi conditions rule store the matching events under m_{n} prefix which can be overridden with the << assignment operator.

https://redhat.atlassian.net/browse/AAP-68094

## Summary by CodeRabbit

* **Documentation**
  * Clarified singular vs plural usage in conditions and actions (when to use event vs events) and added critical usage notes for single-condition vs multi-condition rules.
  * Overhauled “any” and “all” guidance with detailed behavior, default matched-event naming, and practical examples showing correct vs incorrect patterns.
  * Added reminder that matched-event assignment is unsupported in single-condition rules.
  * Expanded FAQ on undefined attributes and best practices for existence checks.

### DIFF
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -18,11 +18,20 @@ Each of the condition(s) can use information from
  * Variables provided by the user at startup saved as vars
 
 When writing conditions
-  * use the **event** prefix when accessing data from the current event
+  * use the **event** prefix when accessing data from the current event in **single condition rules**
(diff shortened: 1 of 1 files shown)


kaiokmo (dismissed):

mdinwoodie:
LGTM!

ttuffin (approved):

ptoscano (approved):

PR #912, opened 2026-04-17
chore: remove legacy pg_notify method

mkanoor:
This was added in PR #616 when we were planning on using Fanout The plan was that there would be one rulebook that would recieve the events and broadcast it to other rulebooks on a well known channel.

This fanout process was abandoned in favor of EventStreams which now do the [PGNotify ](https://github.com/ansible/eda-server/blob/main/src/aap_eda/services/pg_notify.py) on the server side and we have pg_listener in the ansible-rulebook.

This action is not used anymore and we can safely remove it.

Part of the cleanup happened on the EDA Server side in this PR https://github.com/ansible/eda-server/pull/1008 once fanout was removed. Prior to this PR on the EDA Server there was a canned rulebook that used to call the pg_notify action.

this action was for internal use only and was never documented in the docs/actions.rst

## Release Notes

* **New Features**
  * PostgreSQL NOTIFY action now available to send events directly to PostgreSQL channels. Supports automatic chunking of large message payloads with comprehensive error handling and status reporting.


ttuffin (approved):

ttuffin:
@coderabbitai review

ptoscano (approved):
@mkanoor please rebase it, thanks!

PR #915, opened 2026-05-07
feat: bool filter for parity with ansible core filters

mkanoor:
ansible supports a bool filter which converts a string to a boolean.
This gets used in source plugins where we need to have a boolean value
https://redhat.atlassian.net/browse/AAP-74795

e.g.
```
caching_enabled : "{{ my_caching_var | default("false") | bool }}"
```

## Summary by CodeRabbit

* **New Features**
  * Added a Jinja "bool" filter to convert values to boolean, handling nulls, booleans, trimmed case-insensitive truthy/falsy strings, numeric/collection inputs, and standard coercion.

* **Tests**
  * Added comprehensive tests for many inputs, invalid-string error cases, and template integration that produces true boolean values.

* **Documentation**
  * Changelog updated to document the new release and addition.

### DIFF
--- a/ansible_rulebook/jinja/filters.py
+++ b/ansible_rulebook/jinja/filters.py
@@ -11,9 +11,11 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+from __future__ import annotations
+
 import os
 import re
-from typing import Any, Callable, Dict
+from typing import Any, Callable
 
 
 def regex_replace(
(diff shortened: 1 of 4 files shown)


ansible_rulebook/jinja/filters.py
```diff
@@ -97,9 +97,45 @@ def normpath(
...
+        return False
+
+    if isinstance(value, bool):
+        return value
+
+    if isinstance(value, str):
+        lower_value = value.lower().strip()
+        if lower_value in ("true", "yes", "1", "on", "t", "y"):
+            return True
+        elif lower_value in ("false", "no", "0", "off", "f", "n", ""):
```
kaiokmo, line 124: @mkanoor I wouldn't mind the "non-parity" warning from coderabbit. I think it makes sense to be lenient and accept the extra 't', 'f', 'y', 'n'

what got me thinking was the empty string being Falsy. to me, it's risky to consider that as a false statement, because it's ambiguous. imagine a scenario with an event like:

```
{
  "service": "api",
  "deploy_approved": ""
}
```

I don't know, because of a bug, generating a malformed JSON, problems in serialization... 

on the other hand, let's say that somebody is using the service now source plugin, and there's an event generated with an optional field, and in the event itself, that field is sent as empty because it's optional. so in this case, it makes sense to treat it as falsy.

anyway, I'm not against it, but I think we should be 100% EXTRA clear in the documentation that we have these new matching cases, and even clearer about Falsy-ing and empty string

mkanoor: @kaiokmo This is mostly because Jinja converts everything to a string, since we use Jinja's Native template we can preserve the type of the data, but it has to be converted to the correct type, jinja supports the int, float and string but bool is tricky. Ansible has a filter for a bool which we can use.

for e.g
 feedback : "{{ source_queue_feedback | default("False") | bool }}"

The "" string is triicky since in python its considered to be Falsy. We can remove the "" and let it handle the basic true/false type strings

kaiokmo (approved):

ttuffin (approved):

PR #920, opened 2026-06-02
[AAP-76053] fix: handle retries on transient outages

mkanoor:
## What

Add exponential-backoff retry logic across all external connections
(PostgreSQL, Controller API, WebSocket) so that ansible-rulebook
survives transient infrastructure outages instead of crashing.

## Why

During database failovers, controller restarts, or websocket drops,
ansible-rulebook would crash immediately on the first connection
error. This forces operators to manually restart the process after
every transient outage.

Jira: AAP-76053

## Changes

**pg_listener** — tenacity-based async retries with exponential backoff,
TCP keepalives enabled by default (~40s dead connection detection),
auth failure fast-path (no retry on bad credentials).

**job_template_runner** — aiohttp-retry RetryClient for GET
(429/500/502/503/504) and POST (429/502/503) with
ClientConnectorError, configurable via `--controller-retry-max-timeout`
and `--controller-retry-attempts` CLI args /
`EDA_CONTROLLER_RETRY_MAX_TIMEOUT` and `EDA_CONTROLLER_RETRY_ATTEMPTS`
env vars, availability state tracking to avoid log spam.

**websocket** — retry on all ConnectionClosedError codes including 1011.

**Dependencies** — aiohttp-retry >=2.9,<3 and tenacity >=9,<10.

**Other fixes**
- `close_session()` sets `self._session = None` to avoid reusing closed sessions
- `_launch` and `_create_obj` parse response body before `raise_for_status()` to preserve the "already exists" idempotent-create path
- `conftest.py` patches aiohttp 3.14 `stream_writer` compatibility (replaces forked aioresponses dependency)
- `requirements_test.txt` reverted to stock aioresponses
- `tests/e2e/utils.py` uses explicit `raise RuntimeError` instead of `print(data["bad"])`
- `connection_attempts` type hint fixed to `Optional[list]` (PEP 484)

## Test plan

- [x] pg_listener: retry-then-recover, auth failure fast-path,
      keepalive defaults/overrides, backoff reset between failovers
- [x] Controller: 429/502/503/504 retry/recovery, max-retry exhaustion,
      unavailable/restored logging transitions
- [x] WebSocket: 1011 retry e2e test, unit test updated
- [ ] CI pipeline passes

Assisted-by: Claude Code / Opus 4.6 (Anthropic)

## Summary by CodeRabbit

* **New Features**
  * CLI options to configure controller retry timeout and attempts
  * PostgreSQL listener supports reconnect/backoff and TCP keepalive configuration

* **Improvements**
  * Controller interactions and WebSocket flows use exponential backoff, retries, and clearer restored/unavailable logging
  * Job-controller handling tolerates transient/invalid JSON responses more gracefully
  * WebSocket retry logic refines which close codes are treated as transient

* **Documentation**
  * CLI help documents new retry options

* **Tests**
  * E2E and unit tests expanded for retry/reconnect scenarios

* **Chores**
  * Added runtime dependency for HTTP retry support

### DIFF
--- a/ansible_rulebook/event_source/pg_listener.py
+++ b/ansible_rulebook/event_source/pg_listener.py
@@ -4,7 +4,15 @@ import logging
 from typing import Any
 
 import xxhash
-from psycopg import AsyncConnection, OperationalError
+from psycopg import AsyncConnection, OperationalError, sql
+from psycopg.conninfo import conninfo_to_dict
+from psycopg.errors import InvalidAuthorizationSpecification, InvalidPassword
+from tenacity import (
+    AsyncRetrying,
+    retry_if_exception,
+    stop_after_attempt,
+    wait_exponential,
+)
 
 DOCUMENTATION = r"""
 ---
@@ -35,6 +43,44 @@ options:
     type: list
     elements: str
     required: true
+  retry_max_timeout:
+    description:
+      - Maximum backoff time in seconds when retrying after a
+        transient PostgreSQL connection loss. Default is 60.
+    type: float
+    default: 60
+  retry_attempts:
+    description:
+      - Maximum number of retry attempts when reconnecting after a
+        transient PostgreSQL connection loss. Default is 10.
+    type: int
+    default: 10
+  keepalives:
+    description:
+      - Enable TCP keepalives on the PostgreSQL connection.
+        Default is 1 (enabled). Set to 0 to disable.
+    type: int
+    default: 1
+  keepalives_idle:
+    description:
+      - Seconds of inactivity before sending a TCP keepalive probe.
+        Default is 10. Controls how quickly a dead connection is detected
+        after a database failover.
+    type: int
+    default: 10
+  keepalives_interval:
+    description:
+      - Seconds between TCP keepalive probes once the idle threshold
+        is reached. Default is 10.
+    type: int
+    default: 10
+  keepalives_count:
+    description:
+      - Number of TCP keepalive probes before declaring the connection
+        dead. Default is 3. With the defaults the connection will be
+        detected as dead within ~40 seconds (10 + 10*3).
+    type: int
+    default: 3
 notes:
   - Chunking - this is just informational, a user doesn't have to do anything
     special to enable chunking. The sender, which is the pg_notify
@@ -71,6 +117,14 @@ EXAMPLES = r"""
     channels:
       - my_events
       - my_alerts
+
+- eda.builtin.pg_listener:
+    dsn: "host=localhost port=5432 dbname=mydb"
+    channels:
+      - my_events
+    keepalives_idle: 5
+    keepalives_interval: 5
+    keepalives_count: 2
 """
 
 LOGGER = logging.getLogger(__name__)
@@ -138,35 +192,151 @@ def _validate_args(args: dict[str, Any]) -> None:
         raise ValueError(err_msg)
 
 
+PG_RETRY_MAX_TIMEOUT_DEFAULT = 60
+PG_RETRY_ATTEMPTS_DEFAULT = 10
+
+PG_KEEPALIVE_DEFAULTS = {
+    "keepalives": "1",
+    "keepalives_idle": "10",
+    "keepalives_interval": "10",
+    "keepalives_count": "3",
+}
+
+
 async def main(queue: asyncio.Queue[Any], args: dict[str, Any]) -> None:
     """Listen for events from a channel."""
     _validate_args(args)
 
+    max_timeout = float(
+        args.get("retry_max_timeout", PG_RETRY_MAX_TIMEOUT_DEFAULT)
+    )
+    if max_timeout < 2.0:
+        LOGGER.warning(
+            "retry_max_timeout=%.1f is below minimum backoff (2s), "
+            "retries will have minimal delay. Resetting to 2 seconds.",
+            max_timeout,
+        )
+        max_timeout = 2.0
+
+    max_attempts = int(args.get("retry_attempts", PG_RETRY_ATTEMPTS_DEFAULT))
+    reconnecting = False
+
+    # Each outage gets a fresh AsyncRetrying budget.  The retry
+    # only wraps connection + LISTEN setup (a short-lived operation).
+    # Once connected, _process_notifications() blocks on
+    # conn.notifies() outside the retry scope.  When that connection
+    # drops, the outer loop catches it and starts a new retry
+    # sequence with a fresh budget.
+    while True:
+        conn = None
+        connected = False
+        try:
+            async for attempt in AsyncRetrying(
+                retry=retry_if_exception(_is_retryable_pg_error),
+                wait=wait_exponential(multiplier=1, min=2, max=max_timeout),
+                stop=stop_after_attempt(max_attempts),
+                before_sleep=_log_pg_retry,
+                reraise=True,
+            ):
+                with attempt:
+                    conn = await _connect_and_subscribe(args, reconnecting)
+            connected = True
+            await _process_notifications(conn, queue)
+            return
+        except asyncio.CancelledError:
+            LOGGER.info("pg_listener shutdown requested")
+            raise
+        except OperationalError as e:
+            if not connected or not _is_retryable_pg_error(e):
+                raise
+            LOGGER.warning(
+                "PostgreSQL connection lost (%s); "
+                "any in-progress messages were discarded",
+                e,
+            )
+            reconnecting = True
+        finally:
+            if conn and not conn.closed:
+                await conn.close()
+
+
+def _is_retryable_pg_error(exc: BaseException) -> bool:
+    """Retry on transient OperationalErrors but not on auth failures."""
+    return isinstance(exc, OperationalError) and not isinstance(
+        exc, (InvalidPassword, InvalidAuthorizationSpecification)
+    )
+
+
+def _log_pg_retry(retry_state) -> None:
+    exc = retry_state.outcome.exception()
+    wait = retry_state.next_action.sleep
+    LOGGER.warning(
+        "PostgreSQL connection lost (%s), reconnecting in %.0f seconds",
+        exc,
+        wait,
+    )
+
+
+def _build_connect_params(args: dict[str, Any]) -> dict[str, Any]:
+    """Build connection params, applying keepalive defaults only for
+    keys not already present in the DSN or plugin args.
+    """
+    dsn_keys = conninfo_to_dict(args.get("dsn", ""))
+    keepalive_params = {}
+    for key, default in PG_KEEPALIVE_DEFAULTS.items():
+        if key in args:
+            keepalive_params[key] = str(args[key])
+        elif key not in dsn_keys:
+            keepalive_params[key] = default
+    user_params = args.get("postgres_params", {})
+    return {**keepalive_params, **user_params}
+
+
+async def _connect_and_subscribe(
+    args: dict[str, Any],
+    reconnecting: bool,
+) -> AsyncConnection:
+    """Connect to PostgreSQL and subscribe to channels.
+
+    This is the short-lived operation wrapped by AsyncRetrying.
+    Returns the open connection for the caller to listen on.
+    """
+    conn = await AsyncConnection.connect(
+        conninfo=args.get("dsn", ""),
+        autocommit=True,
+        **_build_connect_params(args),
+    )
+    if reconnecting:
+        LOGGER.warning("PostgreSQL connection re-established")
+    cursor = conn.cursor()
+    for channel in args["channels"]:
+        # Compose the query safely using psycopg v3's sql module
+        query = sql.SQL("LISTEN {};").format(sql.Identifier(channel))
+        await cursor.execute(query)
+        LOGGER.debug("Waiting for notifications on channel %s", channel)
+    return conn
+
+
+async def _process_notifications(
+    conn: AsyncConnection,
+    queue: asyncio.Queue[Any],
+) -> None:
+    """Block on conn.notifies() and forward events to the queue.
+
+    Runs outside the retry scope.  Any OperationalError from a
+    dropped connection propagates to the caller.
+    """
+    chunked_cache: dict[str, Any] = {}
     try:
-        async with await AsyncConnection.connect(
-            conninfo=args.get("dsn", ""),
-            autocommit=True,
-            **args.get("postgres_params", {}),
-        ) as conn:
-            chunked_cache: dict[str, Any] = {}
-            cursor = conn.cursor()
-            for channel in args["channels"]:
-                await cursor.execute(f"LISTEN {channel};")
-                LOGGER.debug(
-                    "Waiting for notifications on channel %s", channel
-                )
-            async for event in conn.notifies():
-                data = json.loads(event.payload)
-                if MESSAGE_CHUNKED_UUID in data:
-                    _validate_chunked_payload(data)
-                    await _handle_chunked_message(data, chunked_cache, queue)
-                else:
-                    await queue.put(data)
+        async for event in conn.notifies():
+            data = json.loads(event.payload)
+            if MESSAGE_CHUNKED_UUID in data:
+                _validate_chunked_payload(data)
+                await _handle_chunked_message(data, chunked_cache, queue)
+            else:
+                await queue.put(data)
     except json.decoder.JSONDecodeError:
         LOGGER.exception("Error decoding data")
-        raise
-    except OperationalError:
-        LOGGER.exception("PG Listen operational error")
         raise
 
 
--- a/ansible_rulebook/job_template_runner.py
+++ b/ansible_rulebook/job_template_runner.py
@@ -24,6 +24,7 @@ from urllib.parse import urljoin, urlparse
 
 import aiohttp
 import dpath
+from aiohttp_retry import ExponentialRetry, RetryClient
 
 from ansible_rulebook import util
 from ansible_rulebook.conf import settings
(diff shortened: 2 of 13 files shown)


requirements_test.txt
```diff
@@ -21,6 +21,7 @@ freezegun
 oauthlib>=3.2.0
 kubernetes
 urllib3 >= 2.6.3
-aioresponses
+# TODO: revert to 'aioresponses' after https://github.com/pnuckowski/aioresponses/issues/289 is released
+aioresponses @ git+https://github.com/agners/aioresponses.git@fix-aiohttp-3.14-stream-writer
```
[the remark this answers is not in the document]
mkanoor: Fixed

tests/test_websocket.py
```diff
@@ -283,9 +286,11 @@ async def test_send_event_log_to_websocket_with_non_recoverable_exception(
...
+    socket_mock.return_value.send.side_effect = [
+        websockets.exceptions.ConnectionClosedError(rcvd=rcvd, sent=None),
+        data_sent.append({"a": 1}),
+        data_sent.append({"b": 2}),
+    ]
 
-    with pytest.raises(websockets.exceptions.ConnectionClosedError):
-        await send_event_log_to_websocket(queue)
+    await send_event_log_to_websocket(queue)
+    assert data_sent == [{"a": 1}, {"b": 2}]
```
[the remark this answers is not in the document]
mkanoor: Fixed

ansible_rulebook/job_template_runner.py
```diff
@@ -120,8 +135,27 @@ async def _get_page(self, href_slug: str, params: dict) -> dict:
...
+            )
         except aiohttp.ClientError as e:
+            if self._controller_available:
+                self._controller_available = False
+                logger.warning(
+                    "Controller unavailable (%s), " "requests will be retried",
+                    e,
+                )
             logger.error(CLIENT_CONNECT_ERROR_STRING, str(e))
             raise ControllerApiException(str(e))
```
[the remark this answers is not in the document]
mkanoor: Fixed

wfealdel:
@coderabbitai resume

wfealdel:
@mkanoor 

### Critical Finding

**`_create_obj` "already exists" check broken by `raise_for_status()` reordering** — `job_template_runner.py`

The PR swapped the order of body parsing and status checking in both `_create_obj` and `_launch`. The original code parsed the response body *before* calling `raise_for_status()`:

```python
# Original (main branch) — _create_obj
body = json.loads(await post_response.text())   # body populated first
post_response.raise_for_status()                 # raises on 400, but body is set
```

The PR reverses this:

```python
# PR — _create_obj
post_response.raise_for_status()                 # raises on 400 immediately
response_text = await post_response.text()       # never reached on error
body = json.loads(response_text)                 # body stays None
```

When the controller returns HTTP 400 with `"already exists"` (concurrent label creation in HA), `raise_for_status()` throws before `body` is assigned. The exception handler evaluates `"already exists" in str(None)` — always `False`. This breaks the idempotent-create retry path in `_get_or_create_label`.

**Same pattern in `_launch`:** the `if body: logger.error("Error: %s", body)` block in the `ClientError` handler is now dead code.

**Suggested fix:** parse the body before `raise_for_status()`, preserving the original order:

```python
async with self._session.post(...) as post_response:
    response_text = await post_response.text()
    if response_text:
        body = json.loads(response_text)
    post_response.raise_for_status()
    if not body:
        raise ControllerApiException(f"Controller returned empty response from {url}")
    return body, False
```

### Observations

- **pg_listener retry budget**: In production, `_listen()` blocks forever on `conn.notifies()`, so after a successful retry, subsequent connection drops consume the *same* `AsyncRetrying` attempt budget rather than resetting. The test `test_backoff_resets_between_failovers` uses a finite mock iterator which masks this — each outage gets a fresh `AsyncRetrying` in the test but not in production. Low risk (10 attempts is generous), but worth noting for very long-running deployments.

- **WebSocket unbounded retries**: `_connect_websocket` retries indefinitely on transient close codes with exponential backoff (capped at 60s). No circuit breaker or retry budget. Acceptable for the current use case but could block the process indefinitely if the server persistently returns transient codes.

- **POST retry on `_launch`**: Retrying POST on 502/503 is a pragmatic trade-off. 502/503 typically mean the request didn't reach the backend, but in edge cases the gateway could return 502 after the controller accepted the launch, risking duplicates. No server-side idempotency is apparent.

### Nits (open from prior reviews)

- `tests/e2e/utils.py:167` — `connection_attempts: list = None` should be `list | None = None` (PEP 484)
- `tests/test_controller.py:816-852` — uses manual `os.environ` try/finally instead of `monkeypatch`; doesn't close the runner session
- `tests/e2e/utils.py:212` — `print(data["bad"])` to force errors is brittle; a `raise RuntimeError(...)` would be more explicit

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

tests/test_controller.py
```diff
@@ -783,3 +776,125 @@ async def test_get_job_url_from_label_invalid_type(new_job_template_runner):
...
+
+@pytest.mark.asyncio
+async def test_controller_unavailable_and_restored_logging(caplog):
+    """Verify unavailable/restored logging."""
+    import os
+
+    from ansible_rulebook.job_template_runner import JobTemplateRunner
+
+    old_val = os.environ.get("EDA_CONTROLLER_RETRY_ATTEMPTS")
+    os.environ["EDA_CONTROLLER_RETRY_ATTEMPTS"] = "1"
```
ptoscano: Simply use `monkeypatch.setenv()` for this, as done in other tests.

requirements_test.txt
```diff
@@ -21,6 +21,7 @@ freezegun
 oauthlib>=3.2.0
 kubernetes
 urllib3 >= 2.6.3
-aioresponses
+# TODO: revert to 'aioresponses' after https://github.com/pnuckowski/aioresponses/issues/289 is released
+aioresponses @ git+https://github.com/agners/aioresponses.git@8f40e0db2eed16621607ecfc857168240c795755
```
ptoscano: This commit doesn't exist in the actual aioresponses.git repository, but in a fork (and thus can disappear any time).

I'd rather copy https://github.com/j7an/dep-rank/pull/123 locally in our `conftest.py`, so we can still use an official version of `aioresponses`.

ansible_rulebook/job_template_runner.py
```diff
@@ -86,16 +89,41 @@ def host(self, value: str):
...
+            self._post_retry_options = ExponentialRetry(
+                attempts=int(os.getenv("EDA_CONTROLLER_RETRY_ATTEMPTS", "5")),
+                start_timeout=2.0,
+                max_timeout=settings.controller_retry_max_timeout,
+                factor=2.0,
+                statuses={429, 502, 503},
+                retry_all_server_errors=False,
+                exceptions={aiohttp.ClientConnectorError},
+                methods={"POST"},
+            )
```
ptoscano, line 130: The only difference between these two `ExponentialRetry` objects is that the one for POST does not handle the states 500 and 504. Doesn't POST need to handle 500 and 504?

Let's use a single object instead, if there is no reason to duplicate them.

ansible_rulebook/job_template_runner.py
```diff
@@ -86,16 +89,41 @@ def host(self, value: str):
...
         if self._session is None:
             limit = int(os.getenv("EDA_CONTROLLER_CONNECTION_LIMIT", "30"))
-            self._session = aiohttp.ClientSession(
+            client_session = aiohttp.ClientSession(
                 connector=aiohttp.TCPConnector(limit=limit),
                 headers=self._auth_headers(),
                 auth=self._basic_auth(),
+            )
+            retry_options = ExponentialRetry(
+                attempts=int(os.getenv("EDA_CONTROLLER_RETRY_ATTEMPTS", "5")),
```
ptoscano: Shouldn't it be better to set this as additional field in `setting`, as done with `controller_retry_max_timeout`?
Does it need to be also a new CLI option?

ansible_rulebook/job_template_runner.py
mkanoor: @ptoscano When we are creating an object in the server and it issues a 500 we dont know what it has created before it crashed. Retrying on 500 with POST is risky it could relaunch multiple jobs but with GET its ok since GET is an idempotent call. Initially I had one Retry but had to change to 2 because of the POST issue.

ansible_rulebook/job_template_runner.py
mkanoor: Added to CLI and env var

tests/test_controller.py
```diff
@@ -783,3 +776,120 @@ async def test_get_job_url_from_label_invalid_type(new_job_template_runner):
...
+    """Verify _launch checks status before parsing JSON on 5xx."""
+    monkeypatch.setattr(settings, "controller_retry_attempts", 1)
+    monkeypatch.setattr(settings, "controller_retry_max_timeout", 0.0)
+    new_job_template_runner._create_session()
+    launch_url = "api/v2/job_templates/1/launch/"
+    with aioresponses() as mocked:
+        url = f"{new_job_template_runner.host}{launch_url}"
+        mocked.post(url, status=502, body="")
+        with pytest.raises(ControllerApiException):
+            await new_job_template_runner._launch({}, url)
```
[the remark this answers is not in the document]
mkanoor: Fixed

docs/usage.rst
```diff
@@ -73,8 +75,12 @@ The `ansible-rulebook` CLI supports the following options:
                             The file containing one ansible vault password, can also be passed via the env var EDA_VAULT_PASSWORD_FILE.
     --vault-id VAULT_ID   label@filename pointing to an ansible vault password file
     --ask-vault-pass      Ask vault password interactively 
-    --persistence-id   PERSISTENCE_ID 
+    --persistence-id   PERSISTENCE_ID
                          The unique id, preferably a UUID to track persistent event data
+    --controller-retry-max-timeout CONTROLLER_RETRY_MAX_TIMEOUT
+                            Maximum backoff time in seconds for controller API retries on transient errors (429/502/503/504). Default is 60. Can also be passed via env var EDA_CONTROLLER_RETRY_MAX_TIMEOUT
+    --controller-retry-attempts CONTROLLER_RETRY_ATTEMPTS
+                            Number of retry attempts for controller API calls on transient errors. Default is 5. Can also be passed via env var EDA_CONTROLLER_RETRY_ATTEMPTS
```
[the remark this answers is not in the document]
mkanoor: Fixed

ansible_rulebook/job_template_runner.py
```diff
@@ -396,12 +455,27 @@ async def _launch(self, job_params: dict, url: str) -> dict:
...
             ) as post_response:
-                body = json.loads(await post_response.text())
+                response_text = await post_response.text()
+                if not response_text:
+                    raise ControllerApiException(
+                        f"Controller returned empty response from {url}"
+                    )
+                body = json.loads(response_text)
                 post_response.raise_for_status()
                 return body
```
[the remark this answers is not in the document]
mkanoor: Fixed

ansible_rulebook/event_source/pg_listener.py
```diff
@@ -138,16 +191,95 @@ def _validate_args(args: dict[str, Any]) -> None:
...
+            LOGGER.warning("PostgreSQL connection lost (%s)", e)
+            async for attempt in AsyncRetrying(
+                retry=retry_if_exception(_is_retryable_pg_error),
+                wait=wait_exponential(multiplier=1, min=2, max=max_timeout),
+                stop=stop_after_attempt(max_attempts),
+                before_sleep=_log_pg_retry,
+            ):
+                with attempt:
+                    await _listen(queue, args, True)
+            reconnecting = False
```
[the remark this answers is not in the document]
mkanoor: Fixed

mkanoor:
@wfealdel Please review updated

ansible_rulebook/event_source/pg_listener.py
```diff
@@ -138,36 +191,137 @@ def _validate_args(args: dict[str, Any]) -> None:
...
+                reraise=True,
+            ):
+                with attempt:
+                    conn = await _connect_and_subscribe(args, reconnecting)
+            await _process_notifications(conn, queue)
+            return
+        except asyncio.CancelledError:
+            LOGGER.info("pg_listener shutdown requested")
+            raise
+        except OperationalError as e:
```
wfealdel, line 249: When a connection drops mid-chunk during `_process_notifications`, the partially accumulated `chunked_cache` is silently discarded (it's created fresh each call). Consider logging a warning here so operators can diagnose missing events after a failover, e.g.:

```suggestion
            LOGGER.warning("PostgreSQL connection lost (%s); "
                           "any in-progress chunked messages were discarded", e)
```

ansible_rulebook/event_source/pg_listener.py
```diff
@@ -138,36 +191,137 @@ def _validate_args(args: dict[str, Any]) -> None:
...
+PG_RETRY_ATTEMPTS_DEFAULT = 10
+
+PG_KEEPALIVES_DEFAULT = 1
+PG_KEEPALIVES_IDLE_DEFAULT = 10
+PG_KEEPALIVES_INTERVAL_DEFAULT = 10
+PG_KEEPALIVES_COUNT_DEFAULT = 3
+
+
 async def main(queue: asyncio.Queue[Any], args: dict[str, Any]) -> None:
     """Listen for events from a channel."""
```
wfealdel, line 207: No validation prevents an operator from setting `retry_max_timeout: 0`, which clamps `wait_exponential(min=2, max=0)` to zero-delay retries — potentially hammering a recovering database. Consider clamping to a minimum or warning:

```suggestion
    max_timeout = float(
        args.get("retry_max_timeout", PG_RETRY_MAX_TIMEOUT_DEFAULT)
    )
    if max_timeout < 2.0:
        LOGGER.warning(
            "retry_max_timeout=%.1f is below minimum backoff (2s), "
            "retries will have minimal delay",
            max_timeout,
        )
```

wfealdel (commented):
@mkanoor 
Overall the retry logic across all three connection paths looks solid. Two minor suggestions on pg_listener resilience — see inline comments.

wfealdel (dismissed):
LGTM!

ptoscano (dismissed):
LGTM now with the changes, thanks!

ansible_rulebook/event_source/pg_listener.py
mkanoor: Updated

ansible_rulebook/event_source/pg_listener.py
```diff
@@ -138,36 +191,149 @@ def _validate_args(args: dict[str, Any]) -> None:
...
+    conn = await AsyncConnection.connect(
+        conninfo=args.get("dsn", ""),
+        autocommit=True,
+        **_build_connect_params(args),
+    )
+    if reconnecting:
+        LOGGER.warning("PostgreSQL connection re-established")
+    cursor = conn.cursor()
+    for channel in args["channels"]:
+        await cursor.execute(f"LISTEN {channel};")
```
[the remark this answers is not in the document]
mkanoor: Fixed

ansible_rulebook/event_source/pg_listener.py
```diff
@@ -138,36 +192,152 @@ def _validate_args(args: dict[str, Any]) -> None:
...
+
+def _is_retryable_pg_error(exc: BaseException) -> bool:
+    """Retry on transient OperationalErrors but not on auth failures."""
+    return isinstance(exc, OperationalError) and not isinstance(
+        exc, (InvalidPassword, InvalidAuthorizationSpecification)
+    )
+
+
+def _log_pg_retry(retry_state) -> None:
+    exc = retry_state.outcome.exception()
```
ttuffin, line 271: Since we're logging the exception here, is there a risk that we may expose some sensitive Postgres DSN info?

ansible_rulebook/job_template_runner.py
```diff
@@ -120,9 +155,33 @@ async def _get_page(self, href_slug: str, params: dict) -> dict:
...
+            )
         except aiohttp.ClientError as e:
-            logger.error(CLIENT_CONNECT_ERROR_STRING, str(e))
+            if self._controller_available:
+                self._controller_available = False
+                logger.warning(
+                    "Controller unavailable (%s), retries exhausted",
+                    e,
+                )
+            logger.exception(CLIENT_CONNECT_ERROR_STRING, str(e))
```
ttuffin: Similar to the question above - since we're changing from logger.error to logger.exception, do we risk exposing sensitive connection secrets in the stack trace? Same concern with the changes below.

wfealdel (dismissed):

wfealdel (approved):

ansible_rulebook/job_template_runner.py
mkanoor: @ttuffin I changed it to 
```
logger.error(CLIENT_CONNECT_ERROR_STRING, str(e))  # NOSONAR
```

ansible_rulebook/event_source/pg_listener.py
mkanoor: @ttuffin I am trying to collect the logs from my testing this one should be ok since its just printing the exception as string and no stack trace

mkanoor: @ttuffin 
```
<run_path> - WARNING - PostgreSQL connection lost (consuming input failed: SSL SYSCALL error: Connection timed out); any in-progress messages were discarded
<run_path> - WARNING - PostgreSQL connection re-established
```

ttuffin (approved):

PR #924, opened 2026-06-15
[AAP-77493] Solve SonarCloud quality gate failures and improve supply-chain security

wfealdel:
Pin GitHub Actions to commit SHAs to resolve SonarCloud security hotspots and improve CI/CD supply-chain security.

  ## Why

  The ansible-rulebook SonarCloud project is failing quality gates with security findings related to GitHub Actions referenced by tags instead of commit SHAs. This change
  addresses:

  - **12 security hotspots (githubactions:S7637)**: GitHub Actions referenced by semantic version tags (@v4, @v6, etc.) instead of immutable commit SHAs
  - **1 vulnerability (text:S8565)**: Missing lock file (resolved via SonarCloud exclusion - this is a library package)

  Pinning GitHub Actions to commit SHAs prevents supply-chain attacks where an attacker could maliciously retag a version to inject malicious code.

  **Note**: Python dependencies and Dockerfile packages remain unpinned to avoid the maintenance burden of stale pinned versions. 
  dependency updates.

  Jira: AAP-77493

  ## Changes

  - Pin all GitHub Actions in workflows and composite actions to commit SHAs with inline version comments
  - Add `sonar-project.properties` to exclude lock file requirement (library packages should not commit lock files)

  ## Test plan

  - [x] All GitHub Actions pinned to commit SHAs from official release tags
  - [x] Inline comments document semantic versions for maintainability
  - [ ] CI pipeline passes (all workflows, tests, linters)
  - [ ] SonarCloud quality gate improves (12 security hotspots resolved)

  Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Chores**
  * Improved CI/CD pipeline security by pinning GitHub Actions and Docker-related actions to specific commit versions across build, test, release, scheduled, and analysis workflows.
  * Updated setup and CI helper actions to use commit-SHA–pinned action references.
  * Added a SonarCloud issue exclusion to ignore a specific issue rule for `pyproject.toml`.

### DIFF
--- a/.github/actions/image-test/action.yml
+++ b/.github/actions/image-test/action.yml
@@ -5,7 +5,7 @@ runs:
   using: composite
   steps:
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.0.0
       with:
         python-version: 3.11
 
@@ -16,10 +16,10 @@ runs:
         echo "SETUPTOOLS_SCM_PRETEND_VERSION=$(python -m setuptools_scm)" >> $GITHUB_ENV
 
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v2
+      uses: docker/setup-buildx-action@885d1462b80bc1c1c7f0b00334ad271f09369c55  # v2.0.0
 
     - name: Build container image
-      uses: docker/build-push-action@v4
+      uses: docker/build-push-action@0a97817b6ade9f46837855d676c4cca3a2471fc9  # v4.0.0
       with:
         context: .
         platforms: linux/amd64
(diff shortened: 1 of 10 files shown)


ptoscano (commented):
I'm fine in pinning the github actions; while each version (including patch releases) will require an update, we have dependabot that helps.

The changes in `Dockerfile` are not needed: #910 will do a better job for them, so please drop them. Also, that `Dockerfile` is for development, and pinning versions there does not bring any advantage (especially when other bits in that `Dockerfile` are not pinned either).

Regarding the other pinned Python dependencies in github workflows and `requirements_lint.txt`: while I understand why it is done, this will soon become a problem: those pinned dependencies will be forgotten, and there will be dependency/installability issues with newer versions of other Python modules (it's a recurring issue in e.g. the `event-driven-ansible` repository).

The changes in `tests/test_examples.py` are completely undocumented and unrelated to the "pin github actions and Python dependencies". If they are needed, please split them in an own PR, so they can be documented properly, and reviewed separately.

ptoscano (approved):

jcraiglo1 (approved):

PR #937, opened 2026-07-07
chore: add fastavro python package for kafka

mkanoor:
Now that we support AVRO we need to add fastavro when building an image.

## Summary by CodeRabbit

* **Chores**
  * Updated the build environment to include an additional Python package.


ptoscano:
While this is a no-brainer, I'd like to have #910 instead if possible (which fixes this and various other things).

mkanoor:
Closed in favor of #910

PR #945, opened 2026-07-09
fix: catch TimeoutError before OSError in websocket connect retry loop

ttuffin:
TimeoutError is a subclass of OSError in Python's exception hierarchy. The OSError handler was catching handshake timeouts first and re-raising them (no [Errno 61] match), making the TimeoutError retry branch unreachable. This caused a single transient handshake timeout to abort an activation instead of retrying with backoff.

Fix by catching asyncio.exceptions.TimeoutError explicitly before OSError. InvalidMessage is also split out of the combined tuple for clarity.

https://redhat.atlassian.net/browse/AAP-82022

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved websocket connection handling so timeouts are treated separately from other connection errors.
  * Retry and backoff behavior now works more reliably when websocket connections take too long to establish.
  * Prevented timeout errors from being misclassified, which helps avoid unexpected connection handling behavior.

### DIFF
--- a/ansible_rulebook/websocket.py
+++ b/ansible_rulebook/websocket.py
@@ -121,6 +121,15 @@ async def _connect_websocket(
             else:
                 logger.warning(f"websocket aborted by InvalidStatus: {e}")
                 raise  # abort
+        except asyncio.exceptions.TimeoutError as e:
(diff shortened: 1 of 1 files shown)


ttuffin:
Wrong source branch

PR #797, opened 2025-05-19
[DON'T MERGE] test ansible-core 2.19

hsong-rh:

PR #950, opened 2026-07-21
[AAP-82418] chore: update ansible-core version t0 2.20.0

AlexSCorey:
Bumps ansible-core version to 2.20.  This is needed because after https://github.com/ansible-automation-platform/de-minimal-container/pull/2791 merges tests will likely start the fail without this change.

## Summary by CodeRabbit

* **Updates**
  * Upgraded the bundled Ansible Core version used in builds to **2.20.7**.
  * Updated the minimal decision environment to use **Ansible Core 2.20.7**.
  * Refined Ansible Core version selection for tests so **2.20.7** is chosen for **Python 3.12+**, with non-overlapping constraints for other Python versions.
* **Tests**
  * Updated an engine test to expect **“Action”** event types in the event log.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ARG SETUPTOOLS_SCM_PRETEND_VERSION
 # git repositories:
 # https://docs.ansible.com/projects/ansible/latest/collections_guide/collections_installing.html
 ARG DEVEL_COLLECTION_REPO=ansible.eda
-ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.16.14}
+ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.20.7}
 
 ENV PIP_BUILD_OPTS="--use-pep517 --disable-pip-version-check --wheel-dir /output/wheels"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ARG SETUPTOOLS_SCM_PRETEND_VERSION
 # git repositories:
 # https://docs.ansible.com/projects/ansible/latest/collections_guide/collections_installing.html
 ARG DEVEL_COLLECTION_REPO=ansible.eda
-ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.20.7}
+ARG ANSIBLE_CORE_VER=2.20.7
 
 ENV PIP_BUILD_OPTS="--use-pep517 --disable-pip-version-check --wheel-dir /output/wheels"
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,11 @@
 import inspect
+import re
+import subprocess
 from unittest.mock import Mock
 
 import aiohttp
 import pytest
+from packaging.version import Version
 
 from ansible_rulebook.condition_types import Condition as Conditions
 from ansible_rulebook.rule_types import (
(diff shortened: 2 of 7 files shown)


tests/test_engine.py
```diff
@@ -542,7 +542,7 @@ async def test_run_assert_facts():
         assert event_log.get_nowait()["type"] == "Action", "0.1"
         assert event_log.get_nowait()["type"] == "Job", "1.0"
         for i in range(47):
-            assert event_log.get_nowait()["type"] == "AnsibleEvent", f"1.{i}"
```
mkanoor, line 545: @AlexSCorey Has AnsibleEvent been changed to Action? There are other AnsibleEvents in this repo would they all have to be changed. Action seems like something that comes from ansible-rulebook. Its possible that there was an extra AnsibleEvent that has been added or removed.

Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG SETUPTOOLS_SCM_PRETEND_VERSION
 # git repositories:
 # https://docs.ansible.com/projects/ansible/latest/collections_guide/collections_installing.html
 ARG DEVEL_COLLECTION_REPO=ansible.eda
-ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.16.14}
+ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.20.0}
```
ptoscano: Use the latest of 2.20.x here, which currently is 2.20.7. This version here is installed as `ansible-core==$ANSIBLE_CORE_VER`, so it is a _precise_ version.

tests/test_engine.py
AlexSCorey: it does seem that ansible-core removed the Ansible event type called AnsibleEvent and replaced it with Action.

ptoscano: heh, #797 did skip this test altogether, and had changes for another test

Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG SETUPTOOLS_SCM_PRETEND_VERSION
 # git repositories:
 # https://docs.ansible.com/projects/ansible/latest/collections_guide/collections_installing.html
 ARG DEVEL_COLLECTION_REPO=ansible.eda
-ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.16.14}
+ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.20.7}
```
mkanoor: @AlexSCorey 
We can just use
```
ARG ANSIBLE_CORE_VER=2.20.7
```
When you call podman you can just override it as --build-arg ANSIBLE_CORE_VER=2.21.0

tests/conftest.py
```diff
@@ -114,3 +117,21 @@ def _ruleset(**kwargs):
         )
 
     return _ruleset
+
+
+def get_ansible_playbook_version():
```
ptoscano: Rather than an helper function saved into a module-level variable, make it a session-wide fixture: this way it is still used easily, and only on-demand in tests that use it.

Nitpick: name it `ansible_core_version`, as IMHO fits better.

tests/conftest.py
```diff
@@ -114,3 +117,21 @@ def _ruleset(**kwargs):
...
+            ["ansible-playbook", "--version"], text=True
+        )
+        match = re.search(
+            r"ansible-playbook\s+(?:\[core\s+)?([0-9.]+)", output
+        )
+        if match:
+            return Version(match.group(1))
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return None
+    return None
```
ptoscano: I'd remove all the error handling here, and assert that the version is extracted from the `--version` text. This way it is easier to notice if something does not work (e.g. the output of `--version` changes in newer releases), and a `None` version would break anyway the tests using it.

tests/playbooks/compare_value_core_220.yml
ptoscano, line 1: This file is empty, whereas the pre-2.20 version has checks. Missing anything here?

AlexSCorey: What!!!!!.  Not sure how that happened.

kaiokmo (approved):
although I was not part of the previous discussions, this looks clean to me. therefore, approving

mkanoor (approved):

PR #961, opened 2026-07-24
fix: add ENV HOME=/app for OCP arbitrary UID

kaiokmo:
When CRI-O injects an arbitrary UID into /etc/passwd with HOME set to /, it causes permission denied because the collections dir fall back to /.ansible which is not writable. We need to explicitly set HOME to /app.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime environment consistency by setting the application home directory explicitly.


ptoscano (approved):
Thanks!

mkanoor (approved):

